PR #30, opened 2022-04-25
Issue #26 - Added ED25519 and SECP256K1 support for browser

thehenrytsai:
1. Added Added browser support for ED25519 and SECP256K1.
1. Added a JWS abstraction so that we can more easily swap out implementation as needed.
1. Miscellaneous updates.

PR #31, opened 2022-05-10
jose changes

mistermoe:
* remove use of `AJV` within JOSE algorithms. validation already happens in a previous step
* move to using a single base64url lib across the codebase instead of two
* resolve some TODOs
* favor `Uint8Array` over `Buffer`. Reasoning here: https://stackoverflow.com/a/54814368/4471535
* add function to generate `secp256k1` keypairs
* fix `secp256k1.sign` by hashing the payload prior to signing
* add JSDoc where missing


src/jose/jws.ts
```diff
@@ -1,9 +1,20 @@
 import * as ed25519 from './algorithms/ed25519';
-import * as jwk from './jwk';
 import * as secp256k1 from './algorithms/secp256k1';
-import base64url from 'base64url';
+
+import { base64url } from 'multiformats/bases/base64';
+
 import type { JwkPrivate, JwkPublic } from './jwk';
 
+const verifiers = {
```
thehenrytsai, line 8:
```suggestion
const verifiers: { [key:string]: (payload: Uint8Array, signature: Uint8Array, publicKeyJwk: JwkPublic) => Promise<boolean> } = {
```
And remove type declaration in line 65.

Consider the same thing for `signers` map

src/message.ts
```diff
@@ -119,32 +119,39 @@ export async function verifyMessageSignature(message: Message, didResolver: DIDR
 
   // use a set to do O(1) "lookups" for the correct id
   const verificationMethodIds = new Set();
-  let publicKey;
+  let verificationMethod: VerificationMethod;
 
-  for (let verificationMethod of verificationMethods) {
-    verificationMethodIds.add(verificationMethod.id);
+  for (const vm of verificationMethods) {
+    verificationMethodIds.add(vm.id);
```
thehenrytsai, line 125: Sanity: is the set of ID necessary? Can't we just compare against `vm.id` directly?

src/jose/jwk.ts
```diff
@@ -1,47 +1,8 @@
...
- * A supported public key in JWK format.
- */
 export type JwkPublic = JwkSecp256k1Public | JwkEd25519Public;
-
-const jwkPublicSchema = {
-  anyOf: [
-    ed25519.jwkPublicJsonSchema,
-    secp256k1.jwkPublicJsonSchema
-  ]
-};
```
thehenrytsai, line 22: I am not seeing code that validates `publicKeyJwk` schema in DID document. Are you saying we should kick the can and remove this for now and worry about this later? 'coz I do see a version of this need to exist somewhere when we actually have DID Document validation.

src/message.ts
```diff
@@ -119,32 +119,39 @@ export async function verifyMessageSignature(message: Message, didResolver: DIDR
...
   }
 
-  const result = await jws.verify(attestation, publicKey);
+  const { publicKeyJwk } = verificationMethod;
+
+  // more info about the `publicKeyJwk` property can be found here:
+  // https://www.w3.org/TR/did-spec-registries/#publickeyjwk
+  if (!publicKeyJwk) {
+    throw new Error(`publicKeyJwk property not found on verification method [${kid}]`);
+  }
```
thehenrytsai, line 149: Need clarification: why these validations are done in place here, but not other validations? Based on our previous discussions, I thought you are in favor of JSON schema based validation, if so why the exception here?

src/jose/jwk.ts
```diff
@@ -1,47 +1,8 @@
...
-import type { JwkEd25519Private, JwkEd25519Public } from './algorithms/ed25519';
 import type { JwkSecp256k1Private, JwkSecp256k1Public } from './algorithms/secp256k1';
+import type { JwkEd25519Private, JwkEd25519Public } from './algorithms/ed25519';
 
-/**
- * A supported private key in JWK format.
- */
-export type JwkPrivate = JwkSecp256k1Private | JwkEd25519Private;
+export { generateKeyPair as generateSecp256k1Jwk } from './algorithms/secp256k1';
+export { generateKeyPair as generateEd25519Jwk } from './algorithms/ed25519';
```
thehenrytsai, line 5: For my understanding, you made a point to import then re-export here as opposed to letting the tests/user importing from respective algorithm files.  Is there a convention that I should be aware of or this is purely preference?

Definitely want to pick my fight, and this one is not worth it, but for context the way I saw `jwk.ts` originally is a container for generic JWK functions, as opposed to algorithm specific ones. This aligns with the intent of `jws.ts`. If one wishes to use algo capabilities, import those directly from those files.

thehenrytsai (changes requested):

tests/jose/jws.spec.ts
```diff
@@ -7,35 +8,47 @@ import jwkSecp256k1Private from './vectors/jwk-secp256k1-private.json';
 chai.use(chaiAsPromised);
 
 describe('Jws', () => {
-  it('should sign and verify SECP256K1 signature using a key vector correctly',  async () => {
-    const publicKeyJwk = Object.assign({ }, jwkSecp256k1Private); // Clone private key.
-    delete publicKeyJwk.d; // Remove the private key portion.
+  it('should sign and verify secp256k1 signature using a key vector correctly',  async () => {
```
thehenrytsai, line 11: This test appears to be broken running browser mode.

src/jose/algorithms/secp256k1.ts
```diff
@@ -44,30 +24,58 @@ export type JwkSecp256k1Private = JwkSecp256k1Public & {
...
-  const yBuffer = base64url.toBuffer(publicKeyJwk.y);
-  const publicKeyBuffer = Buffer.concat([identifierByte, xBuffer, yBuffer]);
-  const result = await secp256k1.verify(signatureBuffer, signatureInputBuffer, publicKeyBuffer);
-  return result;
+export async function verify(content: Uint8Array, signature: Uint8Array, publicKeyJwk: JwkSecp256k1Public): Promise<boolean> {
+  const xBytes = base64url.baseDecode(publicKeyJwk.x);
+  const yBytes = base64url.baseDecode(publicKeyJwk.y);
+
+  const leadingByte = Buffer.from([0x04]);
+  const publicKeyBytes = Buffer.concat([leadingByte, xBytes, yBytes]);
```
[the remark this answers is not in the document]
mistermoe: done!

src/jose/jws.ts
mistermoe: ✅

src/message.ts
mistermoe: @thehenrytsai definitely not necessary. It's potentially an overzealous optimization. Using a set to compare for equality makes each comparison `O(1)` vs. `O(n)`. Since we're looping over an unbounded array, I thought it might be worth leveraging O(1) lookups. As i was looking at this again, we could even remove the use of `add` by refactoring the optimization like so:

```js
// use a set to do O(1) lookups for the correct id
const kidSet = new Set();
let verificationMethod: VerificationMethod;

for (const vm of verificationMethods) {
  if (kidSet.has(vm.id)) {
    verificationMethod = vm;
    break;
  }
}
```

Happy to switch to direct comparison if you think the optimization is unnecessary. Lmk what you think

src/jose/jwk.ts
mistermoe: yep! I think that we should tackle validation right after we merge the Jose PR and the blockstore PR. I didn't want to bloat this PR with validation.

src/message.ts
mistermoe: Just as a placeholder. I don't want to forget that we decided to only support `JsonWebKey2020` and `publicKeyJwk`. I went ahead and added tests for this as well so that we remember to consider this when we tackle validation

src/jose/jwk.ts
mistermoe: Purely preference. I mentioned it in my initial review of #30 here: [here](https://github.com/TBD54566975/dwn-sdk-js/pull/30#pullrequestreview-958697434). Basically just a convenience. I could go either way

tests/jose/jws.spec.ts
mistermoe: Passing now! switching from using `Buffer` -> `Uint8Array` fixed the issue 🙃

src/jose/jwk.ts
thehenrytsai: Sounds good.

src/message.ts
thehenrytsai: I feel this warrants a TODO like you had with an issue number attached, because right now it doesn't look like temporary code.

e.g.
// TODO: #123 - Need to replace these validation logic with schema based validation

mistermoe: added TODOs in the code and created a GH to create issues for each TODO: #32

src/jose/algorithms/types.ts
```diff
@@ -0,0 +1,4 @@
+import type { JwkPrivate, JwkPublic } from '../jwk';
+
+export type Signer = (payload: Uint8Array, privateKeyJwk: JwkPrivate) => Promise<Uint8Array>;
+export type Verifier = (payload: Uint8Array, signature: Uint8Array, publicKeyJwk: JwkPublic) => Promise<boolean>;
```
thehenrytsai, line 4: Not seeing the advantage of declaring and exporting the type here since it is only used internally to `jws.ts`.

Supportive of declaring this explicitly however in `jws.ts`.

mistermoe: Sure. will go ahead and do that now

mistermoe: done

PR #40, opened 2022-06-02
Added code coverage badges

thehenrytsai:


package.json
```diff
@@ -67,6 +70,7 @@
     "esbuild": "0.14.39",
     "eslint": "8.11.0",
     "events": "3.3.0",
+    "istanbul-badges-readme": "1.8.1",
```
mistermoe, line 73: this can be a dev dependency yeah?

mistermoe (changes requested):
can we update`CONTRIBUTING.md` to include the added commands? that plus addressing the 1 comment I left and we're 🥇

package.json
thehenrytsai: Sanity: this is already a dev dependency right?

PR #41, opened 2022-06-06
Added ION DID resolver

thehenrytsai:
Added Basic support for ION DID resolver
This is just the first iteration of resolver related code that suffices as a standalone PR.

More PRs to come include:
1. Renaming a number of classes, variables etc
1. Caching
1. Tests for existing code (e.g. `DIDResolver`)
1. Additional linting rules etc


mistermoe:
@thehenrytsai can you resolve the merge conflicts?

thehenrytsai:
> @thehenrytsai can you resolve the merge conflicts?

@mistermoe, resolved.

src/did/ion-did-resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Resolver for ION DIDs.
+ */
+export class IonDidResolver implements DIDMethodResolver {
+  // cross-platform fetch
+  private fetch = crossFetch;
+
+  /**
+   * @param resolutionEndpoint URL to send DID resolution request to
+   */
+  constructor (private resolutionEndpoint: string) { }
```
mistermoe, line 14: thoughts on adding the `beta.microsoft` resolution endpoint as a default here?

src/did/ion-did-resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+  method(): string {
+    return 'ion';
+  }
+
+  async resolve(did: string): Promise<DIDResolutionResult> {
+    // using `URL` constructor to handle both existence and absence of trailing slash '/' in resolution endpoint
+    // appending './' to DID so 'did' in 'did:ion:abc' doesn't get interpreted as a URL scheme (e.g. like 'http') due to the colon
+    const resolutionUrl = new URL('./' + did, this.resolutionEndpoint).toString();
+    const response = await this.fetch(resolutionUrl);
```
mistermoe, line 23: do we need to wrap this with a `try/catch` because `fetch` can throw exceptions if a network error occurs?

mistermoe (changes requested):

src/did/ion-did-resolver.ts
thehenrytsai: I think it is not necessary for now unless you are thinking to throw our own error?

src/did/ion-did-resolver.ts
thehenrytsai: Was thinking to kick the can as more questions would come up like:
1. Are you okay with a straight-up hardcode?
1. Are we going to change to Block's endpoint in the future.
1. `beta.microsoft` itself will not be permanent.

If you are fine with all of the above, then sure, I'll add it.

thehenrytsai: Done: 2cc5b61

thehenrytsai: Done: 2cc5b61

mistermoe (approved):

PR #54, opened 2022-06-14
Use typescript utility types to define the index search instance

NWylynko:
A better solution would be for `@types/search-index` to be updated and have the types exported. But until that is done, using the inbuilt typescript util types works. A type error is raised on [line 82](https://github.com/TBD54566975/dwn-sdk-js/blob/e8e64eee58ccd741153d405f9a7c7c9fee90d5d2/src/store/message-store-level.ts#L82) because `index.QUERY` returns an object with `RESULT` according to the types but on line 82 `RESULTS` is being pulled out of the object. I am assuming that `RESULTS` is correct and that the types are out of date / wrong so I have not changed it in my commit.


NWylynko:
After looking at the `search-index` source code it looks like the types are correct and that `RESULT` is the field returned in the object for a `QUERY`. Running the node tests I don't get an error either way (besides the type error) so it looks like there isn't tests for this part of the code. For now I will leave it as `RESULTS` as I am assuming that this code has been used by other services and websites and works, but I can easily fix this type error.

thehenrytsai (changes requested):
Broke CI pipeline, please fix.

mistermoe:
Sorry for the delay here! been head's down working my way through a large refactor. You're totally right, there weren't any tests that hit the codepath in question. I ran into an error the moment i wrote a test and ended up fixing it on the refactor branch i've been working in [here](https://github.com/TBD54566975/dwn-sdk-js/blob/refactor/src/store/message-store-level.ts#L101)

mistermoe (changes requested):
@NWylynko i resolved the merge conflicts on your branch hoping to merge your PR into main here but ran into another snag.

```
src/store/message-store-level.ts(161,45): error TS2345: Argument of type '{ tokenSplitRegex: RegExp; }' is not assignable to parameter of type 'PutOptions'.
```

It looks like [`@types/search-index`](https://www.npmjs.com/package/@types/search-index) is outdated. We need to add `tokenSplitRegex` to [this interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/search-index/index.d.ts#L112-L121) and/or find a workaround until the types are updated. thoughts?

thehenrytsai:
Submitted PR to update the type definition: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/62613

Waiting on original author's approval.

thehenrytsai (changes requested):
Hey @NWylynko, my typescript definition update to search-index finally got approved, you can reference 3.2 as seen in https://www.npmjs.com/package/@types/search-index to complete this PR!

PR #55, opened 2022-06-15
calling `searchIndex` twice causes the process to hang, so check to see if the index has already been "opened" before opening it again.

Dgarc359:
This is a PR addressing the above comment that is under the TODOs. I've attached a field to blockstore-level that allows access to the instance db status, and a simple check on that status to determine `open()` logic. Please let me know if adjustments are needed!

### DIFF
--- a/src/store/blockstore-level.ts
+++ b/src/store/blockstore-level.ts
@@ -10,7 +10,7 @@ import type { Blockstore } from 'interface-blockstore';
 //  and Chrome for Android.
 export class BlockstoreLevel implements Blockstore {
   db: Level<string, Uint8Array>;
-
+  status: 'opening' | 'closing' | 'open';
   /**
    * @param location - must be a directory path (relative or absolute) where LevelDB will store its
    * files, or in browsers, the name of
@@ -19,6 +19,7 @@ export class BlockstoreLevel implements Blockstore {
    */
   constructor(location: string) {
     this.db = new Level(location, { keyEncoding: 'utf8', valueEncoding: 'binary' });
+    this.status = this.db.status;
   }
 
   async open(): Promise<void> {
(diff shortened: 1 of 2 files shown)


NWylynko:
From the looks of it, two database's are opened, a `BLOCKSTORE` and an `INDEX` levelDB, looking at the file changes it looks like your checking if the `BLOCKSTORE` database is open/opening to then decide to open the `INDEX` db. Should it not check if the index db is open already to decide whether to open it?

Dgarc359:
> From the looks of it, two database's are opened, a `BLOCKSTORE` and an `INDEX` levelDB, looking at the file changes it looks like your checking if the `BLOCKSTORE` database is open/opening to then decide to open the `INDEX` db. Should it not check if the index db is open already to decide whether to open it?

This was my original approach but I wasn't sure if I was reading the comment correct. I can modify the logic to follow more along what you're saying. If I'm understanding right.

I'll revise the PR with the following changes:
 - Revert changes to BlockStoreLevel class
 - Add status field to MessageStoreLevel
 - Flip status field and add checks on calls to `MessageStoreLevel.open() / close()`

Does that sounds more in line with what you had in mind?

mistermoe:
agreed with the approach that @NWylynko laid out and that's currently how the code on the `main` branch behaves. [`BlockstoreLevel#open`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/blockstore-level.ts#L24-L39) already has logic in place to prevent from calling `open` if the `Blockstore` is already `open`. 

I think the way i wrote the TODOs made it seem as if calling `searchIndex` twice was causing the process to hang as something that needed to be addressed. This wasn't actually a TODO, it's just a comment explaining the conditional. The actual TODO is to parameterize `name`. 

[code snippet](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store-level.ts#L46-L49)
```js
// TODO: look into using the same level we're using for blockstore
// TODO: parameterize `name`
// calling `searchIndex` twice causes the process to hang, so check to see if the index
// has already been "opened" before opening it again.
if (!this.index) {
  this.index = await searchIndex({ name: this.config.indexLocation });
}
```

i forgot to add a period here: `// TODO: parameterize name` which probably made it seem like the comment right below it was a part of the TODO. many apologies 😅

PR #56, opened 2022-06-16
Parameterize name in message store level

### LINKED ISSUE #50: Paramaterize name in message-store-level.ts
Calling `searchIndex` twice causes the process to hang, so check to see if the index has already been "opened" before opening it again.

https://github.com/TBD54566975/dwn-sdk-js/blob/98202beeed579628203b45fdaad45583409bd743/src/store/message-store-level.ts#L43-L53

Dgarc359:
Should resolve #50

I've added name as an optional string that can be passed in to `open()`. If `name` is passed in, use that, otherwise use old logic of accessing `this.config.indexLocation`.

Appreciate the opportunity to collaborate!


thehenrytsai (commented):
@Dgarc359 thanks for looking into this, this is a TODO that should have been removed as line 58 with `{ name: this.config.indexLocation }` is the parameterization referred. Will close this PR!

PR #58, opened 2022-06-24
Refactor

mistermoe:
## Overview
* (foundational) provides the building blocks necessary to implement the interfaces described in the DWN spec. Available in `src/core`
* (permissions interface) Implements `PermissionsRequest` and a part of `PermissionsGrant` using the aforementioned building blocks. Available in `src/interfaces/permissions`
* (foundational) implements use of json-schema for validation. Available in `src/validation`
* (foundational) lays groundwork for developers to plug in their own interfaces
* (message-store-level) chunks message `data` into dag-pb unixfs blocks before storing


src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,133 @@
...
+    };
+
+    const descriptorCid = await generateCid(descriptor);
+
+    const authPayload = { descriptorCid: descriptorCid.toString() };
+    const authPayloadStr = JSON.stringify(authPayload);
+    const authPayloadBytes = new TextEncoder().encode(authPayloadStr);
+
+    const signer = await GeneralJwsSigner
+      .create(authPayloadBytes, [opts.signingMaterial]);
```
thehenrytsai, line 80: Sanity: I guess we will have to figure out how a custom signer gets passed in the future right?

thehenrytsai (commented):
This how far I got to, will resume on Monday, but looks super solid, mostly just naming and nits...

src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const message: JsonPermissionsRequest = { descriptor, authorization: signer.getJws() };
+
+    return new PermissionsRequest(message);
+  }
+
+  /**
+   * @throws {Error} if descriptorCid is missing from Auth payload
```
thehenrytsai, line 88: Feels odd that we list just one error out of many possible ones.

src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,135 @@
...
+   */
+  async verifyAuth(didResolver: DIDResolver): Promise<AuthVerificationResult> {
```
thehenrytsai, line 90: Design discussion: do you think it might be worth while consider singleton for DID resolver so that many methods/interfaces can be less bloated.

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+  signers: string[];
+};
+
+// TODO: add logic to prevent validating duplicate signatures
+export class GeneralJwsVerifier {
+  jws: GeneralJws;
+
+  constructor(jws: GeneralJws) {
+    this.jws = jws;
+  }
```
thehenrytsai, line 25: Style: prefer the other non-class style that you normally use (or class with static methods) and pass in the JWS, it feels strange that I have to instantiate a class to verify a JWS, because there is no real state life-cycle.

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+import { verify as verifyEd25519 } from '../../algorithms/ed25519';
+import { verify as verifySecp256k1 } from '../../algorithms/secp256k1';
+
+const verifiers: { [key:string]: VerifyFn } = {
+  'Ed25519'   : verifyEd25519,
+  'secp256k1' : verifySecp256k1
+};
+
+type VerificationResult = {
+  signers: string[];
```
thehenrytsai, line 16: Not clear what the result contain. Perhaps:
```suggestion
  /**
   * DID of list of signers that passed signature verification.
   */
  verifiedSigners: string[];
```

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (!verificationMethod) {
+      throw new Error('public key needed to verify signature not found in DID Document');
+    }
+
+    // TODO: replace with JSON Schema based validation
+    // more info about the `JsonWebKey2020` type can be found here:
+    // https://www.w3.org/TR/did-spec-registries/#jsonwebkey2020
+    if (verificationMethod.type !== 'JsonWebKey2020') {
+      throw new Error(`verification method [${kid}] must be JsonWebKey2020`);
+    }
```
thehenrytsai, line 77: JFYI. Looks like one more high(er) priority updaet to ION/ION SDK.... this has been a low-pri in the backlog. The SDK still uses `EcdsaSecp256k1VerificationKey2019` for sekp256k1.... should be a simple one line fix?! Just bumped the [issue](https://github.com/decentralized-identity/ion/issues/229) priority.

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+    // more info about the `publicKeyJwk` property can be found here:
+    // https://www.w3.org/TR/did-spec-registries/#publickeyjwk
+    if (!publicKeyJwk) {
+      throw new Error(`publicKeyJwk property not found on verification method [${kid}]`);
+    }
+
+    return publicKeyJwk as JwkPublic;
+  }
+
+  async verifySignature(base64UrlPayload: string, signature: Signature, jwkPublic: JwkPublic): Promise<boolean> {
```
thehenrytsai, line 91:
```suggestion
  static async verifySignature(base64UrlPayload: string, signature: Signature, jwkPublic: JwkPublic): Promise<boolean> {
```

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+    const verifyFn: VerifyFn = verifiers[jwkPublic.crv];
```
thehenrytsai, line 92: Prefer naming style consistency throughout code base.
```suggestion
    const verifyFunction: VerifyFunction = verifiers[jwkPublic.crv];
```

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+    const signatureBytes = base64url.baseDecode(signature.signature);
+
+    return await verifyFn(payload, signatureBytes, jwkPublic);
+  }
+
+  decodePayload(): Uint8Array {
+    return base64url.baseDecode(this.jws.payload);
+  }
+
+  private extractDid(kid: string): string {
```
thehenrytsai, line 108:
```suggestion
  static private extractDid(kid: string): string {
```

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+      const { kid } = JSON.parse(protectedJson);
+      const did = this.extractDid(kid);
+      const publicKeyJwk = await this.getPublicKey(did, kid, didResolver);
+
+      const isVerified = await this.verifySignature(this.jws.payload, signature, publicKeyJwk);
+
+      if (isVerified) {
+        signers.push(did);
+      } else {
+        throw new Error('signature verification failed');
```
thehenrytsai, line 43: Generic question on error messages. Is it your intention that we'll revisit each of these messages in a separate pass? Or we should optimistically improve them? Kinda with the opinion that some error messages need more contexts. For instance here would be nice to know which DID failed validation.

src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,135 @@
...
+      throw new Error('provided descriptorCid does not match expected CID');
+    }
+
+    const { signers } = await verifier.verify(didResolver);
+
+    return { signers };
+  }
+}
+
+export const DEFAULT_CONDITIONS: Conditions = {
```
thehenrytsai, line 129: `Conditions` feels too generic, consider perhaps:
```suggestion
export const DEFAULT_CONDITIONS: GrantConditions = {
```

src/interfaces/permissions/request/method.ts
```diff
@@ -0,0 +1,40 @@
+import type { InterfaceMethod } from '../../types';
+import type { JsonPermissionsRequest } from './types';
+
+import { MessageResult } from '../../../response';
+import { PermissionsRequest } from './message';
+
+export const processPermissionsRequest: InterfaceMethod = async (
```
thehenrytsai, line 7: Nit: if you took the rename `method.ts` -> `handler.ts`, then would be nice to be consistent with terminology:

```suggestion
export const handlePermissionsRequest: InterfaceMethod = async (
```

src/interfaces/permissions/request/method.ts
```diff
@@ -0,0 +1,40 @@
...
+  ctx,
+  message,
+  messageStore,
+  didResolver
+): Promise<MessageResult> => {
+  const request = new PermissionsRequest(message as JsonPermissionsRequest);
+
+  if (ctx.tenant !== request.grantedBy && ctx.tenant !== request.grantedTo) {
```
thehenrytsai, line 15: Sanity: why does `ctx.tenant` have to be thge same as `request.grantedTo`?

Also in delegation scenario, how does Carol make request to Bob to get access to Alice's DWN?

According to [spec](https://identity.foundation/decentralized-web-node/spec/#request)
Bob = grantedBy
Carol = grantedTo
Alice = ???

Am I looking at an outdated spec?

src/interfaces/permissions/request/method.ts
```diff
@@ -0,0 +1,40 @@
...
+    return new MessageResult({
+      status: { code: 400, message: 'grantedBy or grantedTo must be the targeted message recipient' }
+    });
+  }
+
+  // TODO: should we add an explicit check to ensure that there's only 1 signer?
```
thehenrytsai, line 21: Yes, should. Actually, I didn't think we have a scenario for multi-sig right? I thought we switched to general JWS just to be consistent with general JWE.

src/dwn.ts
```diff
@@ -1,15 +1,19 @@
...
+import { MessageStoreLevel } from './store/message-store-level';
+import { Request } from './request';
+import { MessageResult, Response } from './response';
+import { PermissionsInterface } from './interfaces';
+
 export class DWN {
-  static methods = {
-    PermissionsRequest: handlePermissionsRequest
+  static methods: { [key:string]: InterfaceMethod } = {
+    ...PermissionsInterface.methods
```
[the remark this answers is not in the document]
mistermoe: done https://github.com/TBD54566975/dwn-sdk-js/pull/58/commits/e8f66d025bea4a4cb6094714e5ca22478151e8c2

src/interfaces/types.ts
```diff
@@ -0,0 +1,17 @@
...
+import { DIDResolver } from '../did/did-resolver';
+
+export type InterfaceMethod = (
+  ctx: Context,
+  message: JsonMessage,
+  messageStore: MessageStore,
+  didResolver: DIDResolver) => Promise<MessageResult>;
+
+export interface Interface {
+  methods: InterfaceMethod[];
```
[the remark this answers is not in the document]
mistermoe: done https://github.com/TBD54566975/dwn-sdk-js/pull/58/commits/e8f66d025bea4a4cb6094714e5ca22478151e8c2

src/interfaces/permissions/request/types.ts
```diff
@@ -0,0 +1,16 @@
+import type { Authorization, JsonMessage } from '../../../messages/types';
+import type { Conditions, Scope } from '../types';
+
+export type PermissionsRequestDescriptor = {
+  description: string
+  grantedTo: string
+  grantedBy: string
+  method: 'PermissionsRequest'
+  scope: Scope
+  conditions: Conditions
```
thehenrytsai, line 10: Consider: `GrantScope` and `GrantCondition` else seems to generic of a term.

src/response/response.ts
```diff
@@ -0,0 +1,45 @@
...
+  status?: ResponseStatus
+  results?: MessageResult[];
+};
+
+/**
+ * class used to create a response that will be returned for a given request
+ */
+export class Response {
+  // present ONLY if there is a general request-related issue
+  // e.g. malformed request, invalid target. `status` and `replies` are mutually exclusive
```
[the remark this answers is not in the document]
mistermoe: > We changed to "results"

Oops, good catch! it is `replies` in the spec. @csuwildcat @thehenrytsai which do y'all want to use, `results` or `replies`?

image


> Also spec really says mutually exclusive? What's the reason against an explicit 200? bloat?

Yep, that's how i read it:
image

@csuwildcat thoughts? what was your thinking behind making `replies` and `status` mutually exclusive? if we are going to keep them mutually exclusive, and `status` can only show up if there's an error, it may make more sense to just name it `error` instead of `status`?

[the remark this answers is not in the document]
thehenrytsai: 1. I think message<->reply make more sense than message<->result.
1. Agree with you that if it's meant to be mutually exclusive then `error` makes more sense!

src/dwn.ts
```diff
@@ -23,22 +27,46 @@ export class DWN {
     this.messageStore = mergedConfig.messageStore;
   }
 
+  async processRequest(rawRequest: any): Promise<Response> {
+    let request: Request;
+
+    try {
+      request = Request.unmarshal(rawRequest);
```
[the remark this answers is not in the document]
mistermoe: done https://github.com/TBD54566975/dwn-sdk-js/pull/58/commits/11df161ac4818bae98b6055a17c96a7187b99062

src/interfaces/permissions/types.ts
```diff
@@ -0,0 +1,30 @@
+export type Scope = {
+  method: string
+  schema?: string
+  objectId?: string
+};
+
+export type Conditions = {
+  // attestation indicates whether any inbound data should be signed.
+  // defaults to `optional`
```
thehenrytsai, line 9: All the "defaults to <xyz>" are weird here because the code doesn't/can't follow the requirement here. Consider remove.

src/response/response.ts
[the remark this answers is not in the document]
mistermoe: @thehenrytsai `MessageResult` -> `MessageReply` resolved: https://github.com/TBD54566975/dwn-sdk-js/commit/5f47f62123688769db41aacaef7923039f137e1e

src/request.ts
```diff
@@ -0,0 +1,39 @@
+import { JsonMessage } from './messages/types';
+import { validate } from './validation/validator';
+
+type JsonRequest = {
+  messages?: JsonMessage[]
```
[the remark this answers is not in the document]
mistermoe: na let's stick to the convention we figured out together on Friday! i'll go ahead and make that change everywhere

src/jose/algorithms/ed25519.ts
```diff
@@ -1,22 +1,7 @@
...
-};
-
-/**
- * An Ed25519 private key in JWK format.
- */
-export type JwkEd25519Private = JwkEd25519Public & {
-  d: string; // Only used by a private key
-};
-
+import type { JwkEd25519Private, JwkEd25519Public, Signfn, VerifyFn  } from '../types';
```
thehenrytsai, line 4: I guess the intent was to move types into the same place? But feels strange here that `JwkEd25519Private` and `JwkEd25519Public` need to be imported from a place that shouldn't have alg specific knowledge.

src/jose/types.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * A SECP256K1 public key in JWK format.
+ * Values taken from:
+ * https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve
+ * https://datatracker.ietf.org/doc/html/draft-ietf-cose-webauthn-algorithms-06#section-3.1
+ */
+export type JwkEd25519Public = Jwk & {
+  alg: 'EdDSA';
```
thehenrytsai, line 13: Sanity: did three of us agree to not require `alg`? Okay as long as it doesn't blow up when it is not supplied by DID resolution.

src/jose/types.ts
```diff
@@ -0,0 +1,45 @@
...
+ * A SECP256K1 private key in JWK format.
+ */
+export type JwkSecp256k1Private = JwkSecp256k1Public & {
+  d: string; // Only used by a private key.
+};
+
+export type JwkPublic = JwkSecp256k1Public | JwkEd25519Public;
+export type JwkPrivate = JwkSecp256k1Private | JwkEd25519Private;
+
+export type Signfn = (payload: Uint8Array, privateKeyJwk: JwkPrivate) => Promise<Uint8Array>;
```
thehenrytsai, line 44: Note lower casing 'f', but prefer
```suggestion
export type SignFunction = (payload: Uint8Array, privateKeyJwk: JwkPrivate) => Promise<Uint8Array>;
```

src/jose/jws/general/signer.ts
```diff
@@ -0,0 +1,59 @@
...
+  'Ed25519'   : signEd25519,
+  'secp256k1' : signSecp256k1
+};
+
+export class GeneralJwsSigner {
+  private jws: GeneralJws;
+
+  constructor(jws: GeneralJws) {
+    this.jws = jws;
+  }
```
thehenrytsai, line 18: Design: similar sentiment to `GeneralJwsVerifier`, not seeing the point of a class, prefer static methods or your common style of exporting function directly, perhaps `sign(...)`.

src/messages/data.ts
```diff
@@ -0,0 +1,46 @@
+import { base64url } from 'multiformats/bases/base64';
+import { CID } from 'multiformats/cid';
+import { importer } from 'ipfs-unixfs-importer';
+
+export class Data {
+  data: string | number | boolean | Uint8Array | object;
```
thehenrytsai, line 6: Do we really want this much optionality after constructions? The way I see it, unless we are able to deserialize them back to their original form, I think it is fine just to have data being `Unit8Array`, and we just call `toBytes()` at construction.

src/messages/data.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  constructor(data: any) {
+    if (!data) {
+      throw new Error('data cannot be null');
+    }
+
+    this.data = data;
+  }
+
+  toBytes(): Uint8Array {
```
thehenrytsai, line 16: I kinda wanna just stick with convertible raw types (e.g. `Uint8Array`), 'coz if I understand current implementation correctly, the type info is lost the moment we serialize and store the bytes in DB. Else the dev experience is kinda half baked and they'll have to either read this method to understand what to expect, or make assumptions when data is returned to them as bytes upon querying.

src/messages/data.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  async getCid(): Promise<CID> {
+    const dataBytes = this.toBytes();
+    const chunk = importer([{ content: dataBytes }], null, { onlyHash: true, cidVersion: 1 });
+    let root;
+
+    for await (root of chunk);
+
+    return root.cid;
```
thehenrytsai, line 44: chunks? Imma trust the voodoo magic does what it is supposed to do!

src/messages/message.ts
```diff
@@ -0,0 +1,49 @@
...
+  constructor(protected message: JsonMessage) {}
+
+  static getJsonSchema(): object {
+    throw new Error('method not implemented');
+  }
+  static getType(): string {
+    throw new Error('method not implemented');
+  };
+
+  static unmarshal(rawMessage: object): JsonMessage {
```
thehenrytsai, line 17:
```suggestion
  static parse(rawMessage: object): JsonMessage {
```

src/messages/message.ts
```diff
@@ -0,0 +1,49 @@
+import type { JsonMessage } from './types';
+
+import lodash from 'lodash';
+import { validate } from '../validation/validator';
+
+const { cloneDeep, isPlainObject } = lodash;
+export abstract class Message {
```
thehenrytsai, line 7: Reminder that you wanted to remove the `abstract`

src/messages/message.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.message.descriptor.method;
+  }
+
+  toObject(): JsonMessage {
+    return cloneDeep(this.message);
+  }
+
+  toJSON(): string {
+    return JSON.stringify(this.message);
+  }
```
thehenrytsai, line 48: The need for a class here is also so thin: `getMethod()` not used (DWN accesses `JsonMessage`'s `descriptor.method` directly, `toObject()` and `toJson()` can easily pass with static methods or function exports.

I am inclined to favor changing this into just type/interface.

src/request.ts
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/c79974cd893ee4e82305bd1ce96bae89f7d75bdd

src/messages/types.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export interface Attestable {
+  attest(): Promise<void>;
+  verifyAttestation(didResolver: DIDResolver): Promise<string>;
+}
+
+export interface AuthCreateOpts {
+  signingMaterial: SignatureInput
+}
```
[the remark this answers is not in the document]
mistermoe: renamed to `AuthCreateOptions` here: https://github.com/TBD54566975/dwn-sdk-js/commit/c79974cd893ee4e82305bd1ce96bae89f7d75bdd

will add JSDoc here to provide context. You're right, it's for readability sake. [example](https://github.com/TBD54566975/dwn-sdk-js/blob/refactor/src/interfaces/permissions/request/message.ts#L14)

src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  static getType(): string {
+    return 'PermissionsRequest';
+  }
+
+  static getInterface(): string {
+    return 'Permissions';
+  }
+
+  static async create(opts: PermissionsRequestOpts): Promise<PermissionsRequest> {
```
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/c79974cd893ee4e82305bd1ce96bae89f7d75bdd

src/interfaces/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { Context } from '../types';
+import type { JsonMessage } from '../messages/types';
+import type { MessageStore } from '../store/message-store';
+import type { MessageResult } from '../response/message-result';
+
+import { DIDResolver } from '../did/did-resolver';
+
+export type InterfaceMethod = (
```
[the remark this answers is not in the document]
mistermoe: done. went with `MethodHandler`: https://github.com/TBD54566975/dwn-sdk-js/commit/e8f66d025bea4a4cb6094714e5ca22478151e8c2

src/request.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  /**
+   * unmarshals the provided payload into a Request.
+   */
+  static unmarshal(rawRequest: any): Request {
+    if (typeof rawRequest !== 'object') {
+      try {
+        rawRequest = JSON.parse(rawRequest);
+      } catch(e) {
```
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/0dcf95e76ac5b0425f5889f6c6dec8a7703f8977

src/interfaces/permissions/request/message.ts
```diff
@@ -0,0 +1,135 @@
...
+    // signature verification is computationally intensive, so we're going to start
+    // by validating the payload.
+
+    const payloadBytes: Uint8Array = verifier.decodePayload();
+    const payloadStr = new TextDecoder().decode(payloadBytes);
+    let payloadJson;
+
+    try {
+      payloadJson = JSON.parse(payloadStr);
+    } catch(e) {
```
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/0dcf95e76ac5b0425f5889f6c6dec8a7703f8977

src/response/response.ts
```diff
@@ -0,0 +1,43 @@
+import { MessageReply } from './message-reply';
+
+type ResponseStatus = {
+  code: number
+  message: string
+};
+
+type ResponseOpts = {
+  status?: ResponseStatus
+  results?: MessageReply[];
```
thehenrytsai, line 10: Sanity:
```suggestion
  replies?: MessageReply[];
```

src/response/response.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * class used to create a response that will be returned for a given request
+ */
+export class Response {
+  // present ONLY if there is a general request-related issue
+  // e.g. malformed request, invalid target. `status` and `replies` are mutually exclusive
+  status?: ResponseStatus;
+  // responses to individual messages provided within a request
+  results?: MessageReply[];
```
thehenrytsai, line 21: `replies` & same for rest of "results"

src/interfaces/permissions/request/method.ts
```diff
@@ -0,0 +1,40 @@
+import type { MethodHandler } from '../../types';
+import type { JsonPermissionsRequest } from './types';
+
+import { MessageReply } from '../../../response';
+import { PermissionsRequest } from './message';
+
+export const processPermissionsRequest: MethodHandler = async (
+  ctx,
```
thehenrytsai, line 8: Consider:
```suggestion
  context,
```

src/store/message-store-level.ts
```diff
@@ -52,69 +62,95 @@ export class MessageStoreLevel implements MessageStore {
...
+      // we really don't have to access the result of `chunk` because it's just outputting every unix-fs
+      // entry that's getting written to the blockstore. the last entry contains the root cid
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      for await (let _ of chunk);
     }
 
+    const indexDocument = {
+      ...jsonMessage.descriptor,
+      _id    : encodedBlock.cid.toString(),
+      author : ctx.author,
```
thehenrytsai, line 150: Should ignore how `author` and `attester` will get populated in this PR?

src/store/message-store-level.ts
```diff
@@ -52,69 +62,95 @@ export class MessageStoreLevel implements MessageStore {
...
-  async put(message: Message): Promise<void> {
-    const encodedBlock = await block.encode({ value: message, codec: cbor, hasher: sha256 });
-    await this.db.put(encodedBlock.cid, encodedBlock.bytes);
+  async put(message: Message, ctx: Context): Promise<void> {
+    const jsonMessage = message.toObject();
 
-    // index specific properties within message
-    const { descriptor } = message;
-    const { method, objectId } = descriptor;
+    // pre-emptively delete data. If data is presen't we'll be chunking it and storing it as unix-fs dag-pb
```
thehenrytsai, line 128:
```suggestion
    // preemptively delete data. If data is present we'll be chunking it and storing it as unix-fs dag-pb
```

src/store/message-store-level.ts
```diff
@@ -52,69 +62,95 @@ export class MessageStoreLevel implements MessageStore {
...
+    const dataBytes = new Uint8Array(dataDagRoot.size);
+    let offset = 0;
 
-    // parse query into a query that is compatible with the index we're using
-    const indexQueryTerms: string[] = MessageStoreLevel.buildIndexQueryTerms(copy);
-    const { RESULTS: indexResults } = await this.index.QUERY({ AND: indexQueryTerms });
+    for await (const chunk of dataDagRoot.content()) {
+      dataBytes.set(chunk, offset);
+      offset += chunk.length;
+    }
```
thehenrytsai, line 93: Consider hiding this magic in a method for readability, perhaps:

```suggestion
    const dataBytes = wait this.getUnixFsDagPbData(descriptor.dataCid);
```

Similarly for line 137 ~ 144 in `put()`

thehenrytsai, line 93: Consider hiding this magic in a method for readability, perhaps:

```suggestion
    const dataBytes = wait this.getUnixFsDagPbData(descriptor.dataCid);
```

Similarly for line 137 ~ 144 in `put()`

src/types.ts
```diff
@@ -0,0 +1,6 @@
+export interface Context {
```
thehenrytsai, line 1: Can you add a brief description? Interested in:
1. Design consideration on why `AbortSignal` is in context and how you anticipate it is to be used/set.
2. Why the interface didn't implement/extend the Option interface?
3. How `author` and `attesters` will be set.

src/validation/json-schemas/index.ts
```diff
@@ -0,0 +1,13 @@
+import defs from './defs.schema.json';
+import GeneralJws from './general-jws.json';
+import PermissionsDefs from './permissions/defs.json';
+import PermissionsRequest from './permissions/permissions-request.json';
+import Request from './request.json';
+
+export const schemas = {
+  defs,
```
thehenrytsai, line 8: Consider:
```suggestion
  definitions,
```
and `PermissionDefinitions`

src/validation/validator.ts
```diff
@@ -0,0 +1,43 @@
...
+import { schemas } from './json-schemas';
+
+const validator = new Ajv();
+
+for (let schemaName in schemas) {
+  addSchema(schemaName, schemas[schemaName]);
+}
+
+export function addSchema(schemaName: string, schema): void {
+  validator.addSchema(schema, schemaName);
```
thehenrytsai, line 11: Sanity: why the one line function? For ease of unit testing?

src/validation/validator.ts
```diff
@@ -0,0 +1,43 @@
...
+  validateFn(payload);
+
+  if (!validateFn.errors) {
+    return;
+  }
+
+  // AJV is configured by default to stop validating after the 1st error is encountered which means
+  // there will only ever be one error;
+  const [ errorObj ] = validateFn.errors;
+  let { instancePath, message } = errorObj;
```
thehenrytsai, line 36: Prefer const over let when possible.
Let's just add the eslint rule and be done with it:
```    'prefer-const': 'error',```
```suggestion
  const { instancePath, message } = errorObj;
```

thehenrytsai (changes requested):
I think I am done reviewing... the PR is solid! As you can see I have mostly just nits... A few repeating themes for me:

1. Prefer naming style consistency over abbreviation: ctx, fn, opts.
1. A few classes are not adding much value and add more complexity (just a little bit), can consider either turn them into interfaces or just remove.
1. Probably an extension to the point above: I get confused a few times with differences between various "message" types/classes/interfaces what not. As discussed before, I'd still really like to have a naming convention that differentiate between pure data containers/types, 
functional interfaces and classes, it helps me to have the right implicit context when reading code. I see `Json` is prefixed to many data containers, I'd be onboard with that, but I also don't see the pattern applied consistently though (as an alternative as mentioned before, I normally suffix with `Model`). How/whether to differentiate classes and interfaces is a trickier discussion so I don't necessarily want to die in the battle for this PR.

src/messages/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  descriptor: {
+    method: string;
+    dataCid: string;
+    [key: string]: unknown;
+  };
+
+  data: string;
+
+  [key:string]: any;
+}
```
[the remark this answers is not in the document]
mistermoe: refactored: https://github.com/TBD54566975/dwn-sdk-js/commit/7d86edbe7891f4f0fb582d44c721acc289660d0b

src/response/response.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/7d86edbe7891f4f0fb582d44c721acc289660d0b

src/response/response.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/7d86edbe7891f4f0fb582d44c721acc289660d0b

src/store/message-store-level.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/7d86edbe7891f4f0fb582d44c721acc289660d0b.

good catch!

mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/7d86edbe7891f4f0fb582d44c721acc289660d0b

src/request.ts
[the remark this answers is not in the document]
mistermoe: where do you think that static method should live?

mistermoe:
@thehenrytsai changed the directory structure a bit per our discussion last Friday. It now looks like this:

```
src
├── core
│   ├── index.ts
│   ├── message-reply.ts
│   ├── message.ts
│   ├── request.ts
│   ├── response.ts
│   └── types.ts
├── did
│   └── did-resolver.ts
├── dwn.ts
├── index.ts
├── interfaces
│   ├── index.ts
│   ├── permissions
│   │   ├── handlers
│   │   │   └── permissions-request.ts
│   │   ├── index.ts
│   │   ├── messages
│   │   │   └── permissions-request.ts
│   │   └── types.ts
│   └── types.ts
├── jose
│   ├── algorithms
│   │   ├── ed25519.ts
│   │   └── secp256k1.ts
│   ├── jwk.ts
│   ├── jws
│   │   └── general
│   │       ├── index.ts
│   │       ├── signer.ts
│   │       ├── types.ts
│   │       └── verifier.ts
│   └── types.ts
├── store
│   ├── blockstore-level.ts
│   ├── message-store-level.ts
│   └── message-store.ts
├── types.ts
├── utils
│   ├── cid.ts
│   └── data.ts
└── validation
    ├── json-schemas
    │   ├── defs.schema.json
    │   ├── general-jws.json
    │   ├── index.ts
    │   ├── permissions
    │   │   ├── defs.json
    │   │   └── permissions-request.json
    │   └── request.json
    └── validator.ts
```

src/interfaces/permissions/request/index.ts
```diff
@@ -0,0 +1,2 @@
+export { PermissionsRequest, DEFAULT_CONDITIONS } from './message';
+export { processPermissionsRequest } from './method';
```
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/pull/58#issuecomment-1169487287

src/jose/jws/general/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  protected: string
+
+  /**
+   * The "signature" member MUST be present and contain the value
+   * BASE64URL(JWS Signature).
+   */
+  signature: string
+}
+
+export interface JwsHeaderParameters {
```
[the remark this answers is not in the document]
mistermoe: all types now. hadn't settled on whether to use types or interfaces across the board yet.

src/interfaces/permissions/request/message.ts
mistermoe: definitely down for that. I think it should be tackled in a separate PR that addresses `DIDResolver`

src/jose/jws/general/verifier.ts
mistermoe: The motivation behind a class driven design is to support the ability to pass the instance around and easily add signatures. e.g.

```js
// instantiated somewhere
const jwsSigner = GeneralJwsSigner.create(payload);

// signature added elsewhere
await jwsSigner.addSignature({ privateJwk, signatureInput });

// signature added elsewhere
await jwsSigner.addSignature(({ privateJwk, signatureInput }));

const jws = jwsSigner.getJws();
```

```js
// example of adding a signature to an existing general Jws
const jwsSigner = new GeneralJwsSigner(jws);
await jwsSigner.addSignature({ privateJwk, signatureInput });
```

```js
// example for if you have all the signature inputs upfront
const jws = GeneralJwsSigner.create(payload)
  .addSignature({ jwkPrivate, signatureInput })
  .addSignature({ jwkPrivate2, signatureInput2 })
  .getJws();
```

lmk what you think. If you're on board i can add these examples and JSDoc to `signer`

src/jose/jws/general/verifier.ts
mistermoe: seems like we'll still want to support `EcdsaSecp256k1VerificationKey2019` though yeah? in order to support DIDs created prior to that issue getting resolved?

mistermoe: @csuwildcat

src/interfaces/permissions/request/method.ts
mistermoe: done

src/messages/message.ts
mistermoe: done

src/interfaces/permissions/request/message.ts
mistermoe: correct

src/jose/jws/general/verifier.ts
thehenrytsai: Got it, as opposed to passing the `jws` around instead? I guess method chaining is making it look nicer. So, the `verifier` here follows the same pattern as the `signer` mainly for consistency? I hope there is a real need for adding signatures in multiple places, I am not seeing it, bit you can make the call since you obviously thought it about it.

src/jose/jws/general/signer.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { Signfn } from '../../types';
+
+import { base64url } from 'multiformats/bases/base64';
+import { sign as signEd25519 } from '../../algorithms/ed25519';
+import { sign as signSecp256k1 } from '../../algorithms/secp256k1';
+
+const signers: { [key:string]: Signfn } = {
+  'Ed25519'   : signEd25519,
+  'secp256k1' : signSecp256k1
+};
```
thehenrytsai, line 11: Sanity: As we discussed earlier, we would need to actually pass in custom signers, we are not designing for that in this PR correct?

mistermoe: correct

src/messages/data.ts
mistermoe: lol

src/jose/types.ts
mistermoe: thoughts on `Signer`?

src/jose/types.ts
mistermoe: we did. can make it optional here as well. Didn't think it was going to make much of a difference if we made it required here since it's confined to this codebase and doesn't imply any runtime checks. Not a fan of explosions so i went ahead and made it optional: https://github.com/TBD54566975/dwn-sdk-js/commit/f9ab6086604c1499eb8cc699d9d492ba0a705113

src/jose/jws/general/verifier.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/f9ab6086604c1499eb8cc699d9d492ba0a705113

src/jose/jws/general/verifier.ts
mistermoe: It was @csuwildcat actually

src/jose/algorithms/ed25519.ts
mistermoe: true. i feel the same way. think it makes more sense to put these in `algorithms/types.ts`?

src/jose/jws/general/verifier.ts
mistermoe: 100% that some error messages need more context. I think we can incrementally improve them as we run into them. went ahead and took care of this one: https://github.com/TBD54566975/dwn-sdk-js/commit/ce65f493f63bf276b01337fed8b78fa1187ba8ba

src/jose/jws/general/verifier.ts
mistermoe: Thoughts on `Verifier` instead? I actually have formed the opinion that signing should imply verification. This should be combined into an interface type called `Signer` that algorithms implement. Thoughts?

src/jose/types.ts
mistermoe: actually: https://github.com/TBD54566975/dwn-sdk-js/pull/58#discussion_r911577125

src/messages/message.ts
mistermoe: how do we provide default implementations of static methods using an type/interface?

src/types.ts
mistermoe: Generally speaking, `Context` is a design pattern that is used to pass request-scoped values and cancellation signals across API boundaries. Context is common in node APIs and provided natively in [go](https://go.dev/blog/context)

1. `AbortSignal` is in context because it provides each caller up the stack to cancel any async operations. Here's a good [resource](https://blog.logrocket.com/complete-guide-abortcontroller-node-js/)
2. Which `Option` interface? The one provided via `interface-blockstore`?
3. My general thought was that both will be set through verifying auth and/or attestation

src/validation/validator.ts
mistermoe: `let` is used here because `instancePath` is re-assigned in the conditional below.

src/messages/message.ts
mistermoe: currently leaning towards keeping it abstract because we don't to expose the ability to create instances of this class. thoughts?

src/interfaces/permissions/types.ts
mistermoe: This `Conditions` typed has to get canned anyway because `Scope` and `Conditions` are now specific to interface methods ☹️

src/jose/jws/general/verifier.ts
mistermoe: `verifiedSigners` feels a bit redundant given that the method throws an exception if any signature verification fails.

src/interfaces/permissions/request/message.ts
mistermoe: so true lol. adding a ton of JSDoc in a PR right now

src/request.ts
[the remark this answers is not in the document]
thehenrytsai: can live in the same file, the comment is more on the constructor

src/core/types.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * concrete Message classes should implement this interface if the Message contains authorization
+ */
+export interface Attestable {
+  attest(): Promise<void>;
+  verifyAttestation(didResolver: DIDResolver): Promise<string>;
+}
+
+export interface AuthCreateOptions {
```
thehenrytsai, line 76: Sanity:
```suggestion
export type AuthCreateOptions {
```

src/interfaces/permissions/request/message.ts
thehenrytsai: Cool, created an issue to track: https://github.com/TBD54566975/dwn-sdk-js/issues/61

src/core/response.ts
```diff
@@ -0,0 +1,43 @@
+import { MessageReply } from './message-reply';
+
+type ResponseStatus = {
+  code: number
+  message: string
+};
+
+type ResponseOpts = {
```
thehenrytsai, line 8: For consistency with the other comments taken.
```suggestion
type ResponseOptions = {
```

src/core/message-reply.ts
```diff
@@ -0,0 +1,25 @@
+import { BaseMessageSchema } from './types';
+
+type Status = {
+  code: number
+  message: string
+};
+
+type MessageResultOpts = {
```
thehenrytsai, line 8: For consistency
```suggestion
type MessageResultOptions = {
```

src/validation/validator.ts
thehenrytsai: Ah, I see, still trying understand the design here. So why is it that we have two place where schemas are added? Is it for customization/extensibility?

src/jose/jws/general/verifier.ts
thehenrytsai: Oh... I must've read the code wrong, I thought the code would continue if any signature verification failed. Will read the code again!

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+// TODO: add logic to prevent validating duplicate signatures
+export class GeneralJwsVerifier {
+  jws: GeneralJws;
+
+  constructor(jws: GeneralJws) {
+    this.jws = jws;
+  }
+
+  async verify(didResolver: DIDResolver): Promise<VerificationResult> {
+    const signers: string[] = [];
```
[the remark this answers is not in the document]
thehenrytsai: Read the code wrong.

src/jose/jws/general/verifier.ts
thehenrytsai: Make sense, I thought the reason you didn't want to have algorithms implement an interface so that we don't need to instantiate instances to use those functions?

src/jose/algorithms/ed25519.ts
thehenrytsai: Same as the other question on `types.ts` why can't `ed25519.ts`  directly export ed25519 related types?

src/messages/message.ts
thehenrytsai: I was actually thinking to get rid of this abstract class entirely and just pass around `MessageSchema` directly.

src/messages/message.ts
thehenrytsai: Can't define interface for static methods. But here I see them as just static method with no requirements to override. Essentially:

```typescript
static getMethod(message: MessageSchema): string {
...
}

static toObject(message: MessageSchema): MessageJson {
...
}


static toJSON(message: MessageSchema): string {
...
}
```

src/types.ts
thehenrytsai: Was more referring to the lack of documentation here for `Context` and I thought you were trying to be compatible with methods defined in `interface Blockstore extends Store<CID, Uint8Array>`, so thought was worthwhile adding some here.

src/validation/validator.ts
thehenrytsai: Do you see value in enabling linting rule 'prefer-const': 'error'? Or it'd probably annoy you because it'll complain about this exact line due to `message`...

src/validation/validator.ts
mistermoe: yep!

src/validation/validator.ts
mistermoe: definitely see the value in adding it. There's an option that allows re-assignment of destructured properties. added the eslint rule. good call! https://github.com/TBD54566975/dwn-sdk-js/commit/17a1689fc67ff4d02c4dfaed1eea51a34e0a6f85

src/core/types.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/79558b120f081869b43b92826b673d788cec72df

src/core/response.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/79558b120f081869b43b92826b673d788cec72df

src/core/message-reply.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/79558b120f081869b43b92826b673d788cec72df

src/interfaces/permissions/index.ts
```diff
@@ -1,4 +1,6 @@
-export * from './permission';
-export * from './permissions-request';
+import { PermissionsRequest, processPermissionsRequest } from './request';
 
-export type PermissionsMethod = 'PermissionsRequest' | 'PermissionsGrant' | 'PermissionsRevoke';
\ No newline at end of file
+export const PermissionsInterface = {
+  methods  : { 'PermissionsRequest': processPermissionsRequest },
+  messages : [ PermissionsRequest ]
```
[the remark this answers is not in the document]
mistermoe: which line? 5?

src/jose/jws/general/verifier.ts
mistermoe: we can open an issue for this if we need to add support for it

src/interfaces/permissions/index.ts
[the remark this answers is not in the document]
thehenrytsai: Yeah, line 5, doesn't look like there is any reference to it?

src/store/message-store-level.ts
thehenrytsai: I think is address in another comment earlier.

src/jose/jws/general/verifier.ts
mistermoe: went a slightly different route but it did:
* address the abbreviation
* implement an interface without the need to instantiate
* reduce the number of files that need to be touched to add a new algo down to 2 (though i'm not sure how often we'll be adding keys)
* prepared the jose dir for encryption-related algos


you may like it less but i think it should work for now 🤣 https://github.com/TBD54566975/dwn-sdk-js/commit/2456261f3856e561e5af7b67190dd02a598faa33

src/jose/algorithms/ed25519.ts
mistermoe: went a slightly different route but it did:
* address the weirdness
* reduce the number of files that need to be touched to add a new algo down to 2 (though i'm not sure how often we'll be adding keys)
* prepared the jose dir for encryption-related algos


you may like it less but i think it should work for now 🤣 https://github.com/TBD54566975/dwn-sdk-js/commit/2456261f3856e561e5af7b67190dd02a598faa33

src/jose/types.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/2456261f3856e561e5af7b67190dd02a598faa33

src/messages/data.ts
mistermoe: nixed the `Data` class. it's now within `utils`.

src/messages/data.ts
mistermoe: Nixed the data class. relevant functions now live within the `utils` directory. We can accept a wide range of types for `data` just for the static `create` methods to make it easier for end-users. that's the hope at least

src/messages/message.ts
mistermoe: hmm. will we still have classes with `create` methods to simplify creation of specific `MessageSchema` types?

src/types.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/f474d19da9be9e9a1b11ca06b209705bba7a0148

src/interfaces/permissions/request/types.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/f99c98c46a5aa494a8041bea0cba32d754c3030a

src/interfaces/permissions/request/message.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/f99c98c46a5aa494a8041bea0cba32d754c3030a

src/request.ts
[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/18fef4444fad3ac3e6990e83454977682a54511e

src/interfaces/permissions/request/method.ts
mistermoe: > Sanity: why does `ctx.tenant` have to be the same as `request.grantedTo`?

the tenant should be either the grantee or grantor, otherwise it seems a bit strange for the person to be storing the permission.

> Also in delegation scenario, how does Carol make request to Bob to get access to Alice's DWN?

Carol would actually make a request to Alice's DWN directly. I don't think we had ever considered requesting delegated permissions. Interesting thought. We should discuss later on and incorporate if needed

src/messages/message.ts
mistermoe: I think i'm going to leave these as is for now. feels a bit easier to just call `toObject` on an "instance" of `Message` and `toJSON` overrides default `JSON.stringify` behavior so i don't think we can make it static

src/validation/json-schemas/index.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/b48197606af323c31f6f31f78e941b4ffc03f61e

src/jose/jws/general/verifier.ts
```diff
@@ -0,0 +1,112 @@
...
+        signers.push(did);
+      } else {
+        throw new Error('signature verification failed');
+      }
+    }
+
+    return { signers };
+  }
+
+  async getPublicKey(did: string, kid: string, didResolver: DIDResolver): Promise<JwkPublic> {
```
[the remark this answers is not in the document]
mistermoe: going to leave this as is for now because the method really should be outright deleted in favor of DID resolving handling this anyway

src/jose/jws/general/verifier.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/cf230926fef5393d78dcfcf8460e3d04ebf32430

src/jose/jws/general/verifier.ts
mistermoe: https://github.com/TBD54566975/dwn-sdk-js/commit/cf230926fef5393d78dcfcf8460e3d04ebf32430

src/interfaces/permissions/request/method.ts
mistermoe: such a strong habit because: https://koajs.com/#context

src/jose/jws/general/verifier.ts
thehenrytsai: This will do! This code has under gone 900 iteration of abstractions. = ]

By the way, REALLY love how you address each comment as individual commits, makes reviewer (my) job so much easier!

tests/jose/jws/general.spec.ts
```diff
@@ -0,0 +1,151 @@
+import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import { DIDResolver } from '../../../src/did/did-resolver';
+import { GeneralJwsSigner } from '../../../src/jose/jws/general/signer';
+import { GeneralJwsVerifier } from '../../../src/jose/jws/general/verifier';
+import { signers } from '../../../src/jose/algorithms';
+import sinon from 'sinon';
+
+const { Ed25519, secp256k1 } = signers;
```
thehenrytsai, line 9: Super nit: casing in consistency.

src/jose/algorithms/signing/ed25519.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import type { PrivateJwk, PublicJwk, Signer  } from '../../types';
+
+function validateKey(jwk: PrivateJwk | PublicJwk): void {
+  if (jwk.kty !== 'OKP' || jwk.crv !== 'Ed25519') {
+    throw new Error('invalid jwk. kty MUST be OKP. crv MUST be Ed25519');
+  }
+}
+
+export const Ed25519: Signer = {
```
thehenrytsai, line 12: Can we export with lower case since it's an object? Same for `Secp256k1`:

`Ed25519` -> `ed25519`
`Secp256k1` -> `secp256k1`

src/messages/data.ts
thehenrytsai: Going to de-dupe with the next comment below.

src/messages/message.ts
thehenrytsai: You are right that classes will still have specific `create` methods that is not part of the abstract class. So in my view no value for this abstraction except calling `super(...)`, but that's just trading one line for another line of code at the moment.

src/messages/data.ts
thehenrytsai: Got it, just making sure there is no disconnect here: we discussed that the job is only half way and unless we also deserialize back to the original form right? My understanding is that we will do this by setting the optional IPFS mime type property thingy, which I am okay for us to now further bloat the PR with, perhaps we can create an issue and reference it here with a `TODO`?

tests/jose/jws/general.spec.ts
mistermoe: done

src/jose/algorithms/signing/ed25519.ts
mistermoe: done

src/messages/message.ts
thehenrytsai: Don't understand the comment on "default `JSON.stringify` behavior" so not 100% convinced static methods can't be used. At the same time there are better things to argue over for sure, so not losing sleep on this one.

Will leave this open one last time and let it simmer a bit more since there is an earlier comment on `create()` methods related to this class, will resolve both if no additional action comes out of that.

src/interfaces/permissions/request/method.ts
thehenrytsai: Just making sure: so you prefer to stay with `ctx` 'coz it's such a special common abbreviation that deserves special treatment? = ]
Wanna be 100% sure, 'coz I don't see it being that special.

tests/jose/jws/general.spec.ts
thehenrytsai: Ignore me, I am dumb, closing.

src/interfaces/index.ts
```diff
@@ -0,0 +1 @@
+export * from './permissions';
```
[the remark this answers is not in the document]
mistermoe: I'm a fan of barrel exporting for the following reasons:
- less import code at the top of files and easier to read the location of those imports. 
- easier mocking/dependency injection
- allows other modules to leverage the barreled imports as a dynamic map

src/interfaces/permissions/request/method.ts
mistermoe: done: https://github.com/TBD54566975/dwn-sdk-js/commit/049a4ec93daa55bacf64cf3e3adb6610710740f2

src/messages/data.ts
mistermoe: sure i can do that once we get this PR merged. adding a link to this comment into my notes

src/messages/message.ts
mistermoe: are you cool with us revisiting after we implement a few more interfaces? wondering if there may be some method we end up realizing needs to be implemented by each concrete message class

src/messages/message.ts
mistermoe: * i like being able to call `toObject` like [this](https://github.com/TBD54566975/dwn-sdk-js/blob/refactor/src/store/message-store-level.ts#L129)
* [overriding default `JSON.stringify` behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior)

src/interfaces/permissions/request/method.ts
mistermoe: 🤣 moreso saying that this won't be the last time you'll be reminding me because `ctx` is etched into my brain. `ctx` is also allll over the codebase 😭

src/store/message-store-level.ts
mistermoe: i don't know if i'm being consistent in saying this relative to the rest of the codebase but i'll say it anyway! 

it's only a handful of lines and it isn't being used anywhere else, so i think it might be ok here for the time being. I also can't think of a good method name lol

src/interfaces/permissions/index.ts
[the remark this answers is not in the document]
mistermoe: will flag this in the next PR that includes `PermissionsGrant`, `PermissionsRevoke`, and `PermissionsQuery`. it'll make more sense in that context

src/messages/message.ts
thehenrytsai: Cool with that.

src/store/message-store-level.ts
thehenrytsai: Sure, this is just a suggestion. Looks like the code is different now to when I first reviewed it anyway. I'll put it into a method if I come across it in next time and having to spend cycles re-learn the block.

src/interfaces/permissions/request/method.ts
thehenrytsai: 1. This is "just" a consistency nit. and
1. sounds like you can't be bothered, and
1. it's "just" a variable name (not a class/type name)

So not a big deal really. Will resolve.

src/interfaces/index.ts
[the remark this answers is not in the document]
thehenrytsai: Barrel export for external consumption makes total sense to me. For internal consumption, not as much, because:

1. currently I am not seeing a clear pattern when `index.ts` should or should not be created
2. there are more than "just" barrel exports in some `index.ts` (exporting a map), which is seems to be bending the intent of barrel exports to my understanding
3. i generally do not look at import statements, so value-add of less lines is "meh"

Don't understand point 2 & 3, but again you clearly are making deliberate points that I am not yet grasping so won't block the PR on this. Will leave open for discussion.

thehenrytsai (approved):
Approved!

Left two comments open on:

1. The need of internal barrel exports
1. The need of `Message` class

Neither is a major concern at all so we can sync separately and iterate in next PR if needed.

src/interfaces/permissions/handlers/permissions-request.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const handlePermissionsRequest: MethodHandler = async (
+  ctx,
+  message,
+  messageStore,
+  didResolver
+): Promise<MessageReply> => {
+  const request = new PermissionsRequest(message as PermissionsRequestSchema);
+
+  if (ctx.tenant !== request.grantedBy && ctx.tenant !== request.grantedTo) {
```
shobitb, line 15: Shouldn't this be `||` instead of `&&`?

src/interfaces/permissions/messages/permissions-grant.ts
```diff
@@ -0,0 +1,140 @@
...
+    const providedConditions = conditions ? conditions : {};
+    const mergedConditions = { ...DEFAULT_CONDITIONS, ...providedConditions  };
+
+    const descriptor: PermissionsGrantDescriptor = {
+      conditions  : mergedConditions,
+      description : options.description,
+      grantedTo   : options.grantedTo,
+      grantedBy   : options.grantedBy,
+      method      : 'PermissionsGrant',
+      objectId    : options.objectId ? options.objectId : uuidv4(),
```
shobitb, line 42: I think the spec calls this `permissionGrantId`?

PR #59, opened 2022-06-24
Added memory cache

thehenrytsai:
Added memory cache. Will hook up to DID resolver in subsequent PR.


package.json
```diff
@@ -52,6 +52,7 @@
     "lodash": "4.17.21",
     "ms": "2.1.3",
     "multiformats": "9.6.4",
+    "node-cache": "5.1.2",
```
mistermoe, line 55: `node-cache` may not be supported in the browser. Though when I went and looked at it's dependencies, i can't see why it wouldn't work in the browser? https://github.com/isaacs/node-lru-cache seems to have explicit browser support.

More generally, how'd you go about deciding on a TTL cache vs. other types of caches (e.g. LRU)

src/did/cache.ts
```diff
@@ -0,0 +1,43 @@
...
+   * Gets the value corresponding to the given key.
+   * @returns value stored corresponding to the given key; `undefined` if key is not found or expired
+   */
+  get(key: string): Promise<any | undefined>;
+}
+
+/**
+ * A cache using local memory.
+ */
+export class MemoryCache implements Cache {
```
mistermoe, line 23: do you think it makes sense to put this into `src/utils` instead? it's so general purpose that we could definitely use it whenever an in-mem cache is needed

mistermoe (changes requested):
Added 2 comments. Aside from that, just need to resolve the merge conflicts!

src/did/cache.ts
thehenrytsai: Sure, will move over to `utils`.

package.json
thehenrytsai: In context of DID resolution, there must be an element of TTL as documents will be changed overtime. But will switched to the lib you referenced because:

1. It has 10x the usage
1. Has support for TTL
1. Seems to have explicit support. (Though I believe `node-cache` made a fix to make it browser compatible and tests pass in `test:browser` mode).

Thanks for the great find!

src/utils/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+  set(key: string, value: any): Promise<boolean>;
+
+  /**
+   * Gets the value corresponding to the given key.
+   * @returns value stored corresponding to the given key; `undefined` if key is not found or expired
+   */
+  get(key: string): Promise<any | undefined>;
+}
+
```
mistermoe, line 19: 2 Questions:
1. What's the primary motivation behind a generalized `Cache` interface? is it so different caches can be implemented with different underlying stores (e.g. redis) while maintaining the same interface?
2. to remain consistent with where exported types are put in the rest of the project, should we put the `Cache` interface into `utils/types.ts`?

src/utils/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+  public constructor (private timeToLiveInSeconds: number) {
+    this.cache = new LruCache({
+      max : 100_000,
+      ttl : timeToLiveInSeconds * 1000
+    });
+  }
+
+  async set(key: string, value: any): Promise<boolean> {
+    this.cache.set(key, value);
+    return true;
```
mistermoe, line 38: what's the purpose of returning a boolean if it always and only returns `true`?

mistermoe (changes requested):

src/utils/cache.ts
thehenrytsai: 1. Yes, this already happened in this PR, and preps for redis.
1. Make sense: 82fcfda

src/utils/cache.ts
thehenrytsai: Fair, half-baked implementation at best.

I flipped back and forth between various thoughts and arrived at changing the interface: 025377d

Let me know if you want to discuss further!

mistermoe (approved):

PR #74, opened 2022-07-26
Fix ESM package

serverwentdown:


thehenrytsai:
Thanks for the PR. The `types` property has been updated since.

PR #77, opened 2022-08-04
CollectionsQuery

thehenrytsai:


mistermoe (changes requested):
awesome start to `CollectionsQuery`. aside from adding the `filter` property to `descriptor` everything looks solid

mistermoe (approved):

PR #79, opened 2022-08-08
feat/78-message-create-input-property-validation closes #78

### LINKED ISSUE #78: Add input property validation for all message `create()` methods
Add input validation for all message `create()` methods by reusing the JSON schema already being used during message processing: https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/validation/validator.ts#L20

Example create method needing update:
https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/collections/messages/collections-write.ts#L26

virajjiwane:
Add input property validation for all message create() methods closes #78


thehenrytsai (changes requested):
@virajjiwane, thanks for working on this and submitting this PR!

The issue is specifically hoping that we can reuse the JSON schema [validation code](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/validation/validator.ts#L20) that we already have so we minimize extra code.

Also `CollectionsWrite` is newly checked in so also needs the same validation.

virajjiwane:
Hi @thehenrytsai, thanks for the reply. I understood the requirement, will work on the changes.

virajjiwane:
@thehenrytsai changes done. please review

src/interfaces/collections/messages/collections-write.ts
```diff
@@ -45,6 +46,9 @@ export class CollectionsWrite extends Message implements Authorizable {
     const authorization = await sign({ descriptor }, options.signatureInput);
     const message = { descriptor, authorization };
 
+    const messageType = descriptor['method'];
```
thehenrytsai, line 49:
```suggestion
    const messageType = descriptor.method;
```

src/interfaces/collections/messages/collections-write.ts
```diff
@@ -45,6 +46,9 @@ export class CollectionsWrite extends Message implements Authorizable {
...
+    validate(messageType, message);
```
thehenrytsai, line 50: validate before `await sign(...)` since signing it costly.

thehenrytsai (changes requested):
Thanks for adding the change, almost there, a couple of comments.

virajjiwane:
@thehenrytsai did changes in all Message create methods and moved validation before any auth functions I could find. please review again.

Issue #80, opened 2022-08-09
Search index appear to be broken for some special characters


thehenrytsai:
After a lot of confusion testing CollectionsWrite, it appears that search index appear to be broken for some special characters. For instance:

`dataFormat` with forward slash: `application/json`
`recordId` with hyphens: `f227a18b-0d2f-4b83-b7c6-1b09e04d98b9`

Running tests against the `henrytsai/collections-write-2` branch: `npm run test:node` will demonstrate the problem.

mistermoe:
@thehenrytsai [here's](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store-level.ts#L165-L167) where you can add additional characters to exclude during tokenization

PR #82, opened 2022-08-10
issue #80 - Search index broken for non letter/digit characters

thehenrytsai:
1. Allowed all characters for indexing
2. Added regression test + test code refactoring


mistermoe (approved):
woohoo

PR #88, opened 2022-09-08
Made changed required by downstream nodejs projects

thehenrytsai:
Made changes required to enable this SDK to be imported to spin up a nodejs DWN. Outlining journey that was taken in this PR:

1. Downstream TS nodejs project is not able to import this SDK:
    Fixed incorrect `types` value in `package.json`, missing `esm` in path; but:
1. Crash on start because JSON-schema file import need import assertion:
    Fixed by adding import assertions; but:
1. Build failure upon adding import assertions:
    Fixed by update `module` value in `tsconfig` to enable import assertion syntax in TS. but:
1. ts-mocha throws TypeError: Unknown file extension ".ts" error when attempting to run tests using `npm run test:ndoe`:
    After trying many things with no success, decided to use this opportunity to ditch `ts-mocha` and `ts-node` entirely by using building the tests and run tests using `mocha` directly, this will give us true code coverage also; but
1. `mocha` fails importing files because esm requires explicit `.js` file imports but we don't do that:
    Added `--es-module-specifier-resolution=node` flag to make node import files correctly. but:
1. `nyc` code coverage tool also mysteriously stopped reporting code coverage at all:
    After trying many things with no success again, decided to replace `nyc` with `c8`, which is now a far more active code coverage tool than the stalled `nyc` anyways; but
1. Spinning up DWN require explicit construction of DID resolver which is a pain:
   Made it an optional parameter with ION being the default resolvable method.
1. Now the external DWN can finally officially run on nodejs for the first time, but immediately throws error for any request because the request parser does not JSON parse if input is an `object`, but raw input is actually ingested as an array of bytes, which is of `object` type:
    Refactored the code to parse the raw bytes before passing to request parser.

### DIFF
--- a/.c8rc
+++ b/.c8rc
@@ -1,0 +1,16 @@
+{
+  "all": true,
+  "cache": false,
+  "extension": [
+    ".js"
+  ],
+  "include": [
+    "dist/esm/src/**"
+  ],
+  "reporter": [
+    "text",
+    "cobertura",
+    "html",
+    "json-summary"
+  ]
+}
--- a/.github/workflows/integrity-check.yml
+++ b/.github/workflows/integrity-check.yml
@@ -26,7 +26,7 @@ jobs:
     # read more about matrix strategies here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix
     strategy:
       matrix:
-        node-version: [16.x]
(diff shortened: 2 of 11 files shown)


mistermoe (approved):
awesome stuff @thehenrytsai! dealing with JS bundle woes is definitely the worst part of TS/JS development IMO. props for removing `ts-node` and `ts-mocha`

PR #89, opened 2022-09-16
Protocol-Based Authorization for CollectionsWrite

thehenrytsai:
Please set aside questions you might have regarding CollectionsQuery as that part is still being ironed out.

Tests added to cover basic mainline scenarios, but code coverage is lacking, will boost it in subsequent PRs on protocol interface itself.


mistermoe (approved):
🎖️

PR #115, opened 2022-10-17
refactored entire codebase to use encoder.ts

Cxxshyy:
This change was to implement a refactor of the codebase which would modify the files to use the encoder.ts file instead of directly referencing underlying dependencies/libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Code Coverage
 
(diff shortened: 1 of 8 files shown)


thehenrytsai (approved):
🍾 Thanks for the contribution!! There is still the replacement for `new TextEncoder().encode(...)` occurrences, but not going to hold this PR for it!

Cxxshyy:
I can take a look at this tomorrow and implement it with a new PR

PR #116, opened 2022-10-18
Adding a decoder.ts file

Cxxshyy:
This pull requests aims to do 2 things first get rid of direct calling for "new TextEncoder().encode(...)"  and "new TextDecoder().decode(...)" occurrences and implement a new file for the decoding aspect of the codebase rather than have both decoding and encoding in one file, makes it more clean and efficient IMO.


thehenrytsai (changes requested):
@Cxxshyy, great observation and call on refactoring `new TextDecoder().decode(...)`!!

Note though t is the intent of the `Encoder` to:

1.  have all the data "transform" methods in one place; and
2. the method name being explicit so there is no need to interpret what decode/encode means

e.g. `bytesToString()` can be interpreted as "encoding bytes to string" or "decoding bytes to string" depending on perspective, so creating a separate class  is not useful from practical use standpoint; in fact it had occurred multiple times that we used `TextDecoder` when we meant to use TestEncoder` and vise versa, which is one of (not the only) motivation behind adding the `Encoder` abstraction in the first place.

All is to say: please consider adding a `bytesToString()` method in Encoder and remove `Decoder`.

Cxxshyy:
@thehenrytsai now made the changes if you want to review again on the recent commit

thehenrytsai (approved):
🚀 🍾

Issue #67, opened 2022-07-22
Replace verificationMethod checks with JSON Schema based validation (JsonWebKey2020)


bnonni:
Replace with JSON Schema based validation for `JsonWebKey2020`. More info about the `JsonWebKey2020` type can be found here: https://www.w3.org/TR/did-spec-registries/#jsonwebkey2020

https://github.com/TBD54566975/dwn-sdk-js/blob/fcea84929e12dd3e3ef0e8866b0e592a025786fa/src/jose/jws/general/verifier.ts#L66-L73

flothjl:
[WIP PR](https://github.com/TBD54566975/dwn-sdk-js/pull/121)

flothjl:
Hey there! 👋 
Big apologies - I goofed for accidentally marking my PR as `requesting review` first - I have changed it to a draft!

flothjl:
In looking at verification method [spec](url) it looks like `controller` is required, so I went ahead and updated current jws tests where verification methods are defined to include `controller` property. Any thoughts on this?

PR #121, opened 2022-10-25
#67 - adding JSON schema validation to verificationMethod

flothjl:


tests/jose/jws/verifier.spec.ts
```diff
@@ -1,14 +1,133 @@
...
+    // more info here: https://sinonjs.org/releases/v13/general-setup/
+    sinon.restore();
+  });
   describe('getPublicKey', () => {
-    xit('throws an exception if publicKeyJwk isn\'t present in verificationMethod', () => {});
-    xit('throws an exception if DID could not be resolved', () => {});
-    xit('throws an exception if appropriate key isnt present in DID Doc', () => {});
-    xit('throws an exception if verificationMethod type isn\'t JsonWebKey2020', () => {});
-    xit('returns public key', () => {});
+    const buildDidResolutionResult = (
```
thehenrytsai, line 16: since the change is isolated to the usage of the schema validator, we can test directly against the validator by adding a file under `https://github.com/TBD54566975/dwn-sdk-js/tree/main/tests/validation/json-schemas` called `jwk-verification-method.spec.ts`, so the test code is not bloated with all the code to stub the resolver.

src/validation/json-schemas/index.ts
```diff
@@ -2,6 +2,7 @@ import CollectionsQuery from './collections/collections-query.json' assert { typ
 import CollectionsWrite from './collections/collections-write.json' assert { type: 'json' };
 import Definitions from './definitions.json' assert { type: 'json' };
 import GeneralJws from './general-jws.json' assert { type: 'json' };
+import GeneralJsonWebToken2020 from './general-jwk-2020.json' assert {type: 'json'};
```
thehenrytsai, line 5:
```suggestion
import JwkVerificationMethod from './jwk-verification-method.json' assert {type: 'json'};
```

flothjl: @thehenrytsai should be addressed in latest commit. Thanks!

tests/jose/jws/verifier.spec.ts
flothjl: great suggestion. Moved tests to `verification-method.spec.ts`

thehenrytsai (approved):
Thanks @flothjl for your valuable contribution! 🥇 🎉 🚀 🎊

Issue #62, opened 2022-07-22
Add cache support to DidResolver


bnonni:
Add DIDCache to constructor method signature.

https://github.com/TBD54566975/dwn-sdk-js/blob/fcea84929e12dd3e3ef0e8866b0e592a025786fa/src/did/did-resolver.ts#L9-L17

Cxxshyy:
i'd like to try this out if someone could guide me

Cxxshyy:
yes i think so. 1st i add the resolvers to the local cache using the set method then in the resolve method i add the get function from cache to check whether the resolver is currently residing in cache at that point in time? elt me know if this is correct.

thehenrytsai:
No no, it's not the list of `DidMethodResolver` themselves that need caching, it's the result returned by  the `resolve(...)` call, specfically:

```typescript
const resolutionResult = await didResolver.resolve(did);
```

`resolve(...)` is possibly expensive, so adding a cache in front of it would help with perf. Let me know if this makes more sense now.

Cxxshyy:
oh yes that's what i meant by my explanation but i guess i didn't  convey it great sorry.

Cxxshyy:
#124 aims to tackle this issue

Cxxshyy:
![image](https://user-images.githubusercontent.com/74113508/198318325-a5f30e18-552d-4b33-8576-c92581859641.png)
im now getting this error

Cxxshyy:
although this file hasnt been modified

thehenrytsai:
I think you pulled recently, but you'll also need to do an `npm install` 'coz dependencies have changed.

thehenrytsai:
Addressed by #155.

PR #124, opened 2022-10-26
added cache for DidResolver

Cxxshyy:
Added in a cache that would check if the did being looked at has been cached and already has a resolution result value prepared and if not to use the resolve function to get the resolution result, then place the value in cache with the did as the key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Code Coverage
 
-![Statements](https://img.shields.io/badge/statements-79.59%25-red.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-87.05%25-yellow.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-81.5%25-yellow.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-79.59%25-red.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-79.61%25-red.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-87.33%25-yellow.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-81.63%25-yellow.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-79.61%25-red.svg?style=flat)
 
 ## Introduction
 
--- a/README.md
+++ b/README.md
@@ -4,11 +4,7 @@
 
 Code Coverage
 
-<<<<<<< HEAD
-![Statements](https://img.shields.io/badge/statements-79.61%25-red.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-87.33%25-yellow.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-81.63%25-yellow.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-79.61%25-red.svg?style=flat)
-=======
 ![Statements](https://img.shields.io/badge/statements-79.86%25-red.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-86.79%25-yellow.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-81.87%25-yellow.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-79.86%25-red.svg?style=flat)
->>>>>>> cdbff6c (added cache for didresolver)
 
 ## Introduction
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -35,34 +38,44 @@ npm link ../path/to/where/you/cloned/dwn-sdk-js
...
+  ></script>
   ```
 
 - **ESM Bundle**
   ```html
-  <script type="text/javascript" src="node_modules/dwn-sdk/dist/bundles/bundle.esm.js"></script>
+  <script
+    type="text/javascript"
+    src="node_modules/dwn-sdk/dist/bundles/bundle.esm.js"
+  ></script>
```
thehenrytsai, line 72: Why the change to double quotes and formatting? Consider undo.

Cxxshyy: not to sure, i ran n lint:fix and this was the output

src/did/did-resolver.ts
```diff
@@ -1,12 +1,14 @@
...
 
 /**
  * TODO: add docs, Issue #72 https://github.com/TBD54566975/dwn-sdk-js/issues/72
  */
 export class DidResolver {
   didResolvers: Map<string, DidMethodResolver>;
 
+  memoryCache = new MemoryCache(86400);//Time to live in seconds is 24 hours
   // TODO: add DIDCache to constructor method signature, Issue #62 https://github.com/TBD54566975/dwn-sdk-js/issues/62
   constructor(resolvers: DidMethodResolver[]) {
```
thehenrytsai, line 13: Pass in the cache through constructor as an optional parameter such that it can be changed and mocked. e.g.

```suggestion
  constructor(resolvers: DidMethodResolver[], cache?: Cache) {
```

Also the TODO should be removed since this PR addresses the issue.

src/did/did-resolver.ts
```diff
@@ -36,7 +38,14 @@ export class DidResolver {
       throw new Error(`${didMethod} DID method not supported`);
     }
 
-    const resolutionResult = await didResolver.resolve(did);
+    const resolutionResult = await this.memoryCache.has(did) ? await this.memoryCache.get(did): await didResolver.resolve(did);
+    const cache = await this.memoryCache.has(did);
+    if (cache){
+      await this.memoryCache.set(did,resolutionResult);
```
thehenrytsai, line 44: Looks like a bug? If result is in cache, no need to store it in cache again, you meant to do `!cache`?

src/did/did-resolver.ts
```diff
@@ -36,7 +38,14 @@ export class DidResolver {
       throw new Error(`${didMethod} DID method not supported`);
     }
 
-    const resolutionResult = await didResolver.resolve(did);
+    const resolutionResult = await this.memoryCache.has(did) ? await this.memoryCache.get(did): await didResolver.resolve(did);
+    const cache = await this.memoryCache.has(did);
```
thehenrytsai, line 42: rename `cache` to `isCached`

src/did/did-resolver.ts
```diff
@@ -36,7 +38,14 @@ export class DidResolver {
       throw new Error(`${didMethod} DID method not supported`);
     }
 
-    const resolutionResult = await didResolver.resolve(did);
+    const resolutionResult = await this.memoryCache.has(did) ? await this.memoryCache.get(did): await didResolver.resolve(did);
```
thehenrytsai, line 41: There is no need to introduce `.has()`, `.get()` will return `undefined` if not in cache, you can use that as your condition.

Cxxshyy: i did use that first but thought the .has would be more clean

src/did/did-resolver.ts
```diff
@@ -1,12 +1,14 @@
...
 import { Did } from './did';
+import { MemoryCache } from '../utils/memory-cache';
 
 /**
  * TODO: add docs, Issue #72 https://github.com/TBD54566975/dwn-sdk-js/issues/72
  */
 export class DidResolver {
   didResolvers: Map<string, DidMethodResolver>;
 
+  memoryCache = new MemoryCache(86400);//Time to live in seconds is 24 hours
```
thehenrytsai, line 11: 24 hours is rather aggressive as default, change to say 10 minutes.

src/did/did-resolver.ts
```diff
@@ -36,7 +38,14 @@ export class DidResolver {
       throw new Error(`${didMethod} DID method not supported`);
     }
 
-    const resolutionResult = await didResolver.resolve(did);
+    const resolutionResult = await this.memoryCache.has(did) ? await this.memoryCache.get(did): await didResolver.resolve(did);
+    const cache = await this.memoryCache.has(did);
+    if (cache){
```
thehenrytsai, line 43: Will need to add tests to cover the code added.

thehenrytsai (changes requested):
@Cxxshyy thanks again for your 3rd PR! 🙏 🏆 🎸 
This one requires more changes though!

src/did/did-resolver.ts
thehenrytsai: Probably logically more clean, but "dirtier" in practice as it requires two round trips to the cache.

Cxxshyy:
@thehenrytsai yeah thanks i'm still learning typescript/javascript and appreciate the guidance🤌

src/did/did-resolver.ts
Cxxshyy: sorry dont understand this, though i did add the cache as an optional parameter however what do you mean by dependency on interface

thehenrytsai: > what do you mean by dependency on interface

The DidResolver should have no reference/knowledge of "memory cache" except for the construction in constructor.

> also what operations are to be done on the cache passed in?

Not sure what you mean.

Cxxshyy: ok thanks

Cxxshyy: the thing still confusing me is the fact that we have the memoryCache being instantiated but then if we are passing in a cache as an optional parameter to allow for ease of testing what operations does the cache passed in do? For example it cant instantiate the memoryCache class as it requires a parameter of TTL(time to live) meaning in cant modify that class unless im missing something, so essentially what do i do with the cache parameter passed in.

thehenrytsai: You will only instantiate the memory cache if no cache is passed in assuming cache is an optional parameter, in which case you should simply instantiate the cache with a default timeout as suggested. The design of cache as an "optional" parameter is simply a convenience, if the developer doesn't like to use memory cache or the default timeout set for memory cache, fine, pass in their own.

One could argue for making the cache always required to be passed in, I am okay with that, not too opinionated on that, but dependency injection of the cache is a must.

Cxxshyy: ok thanks that makes sense but cba you also check the issues comments as there is a new problem affecting testing of the code

Cxxshyy: can

thehenrytsai: Answered in issue #62

Cxxshyy: i have another question so right now i have modified my code to have a boolean arg which will allow for injection of a new cache and if its false will use memorycache instantiating and youre able to pass in a timeout else default value is 600.

Now the question is if the type of cache can be modified then how will that affect the functions currently being used for example .get() cannot be used if the type of cache is Cache but can be used if its MemoryCache?

Cxxshyy: unless im being confused and cache:Cache is only meant to be used for testing?

thehenrytsai:
@Cxxshyy, checkout PR https://github.com/TBD54566975/dwn-sdk-js/pull/133, you can follow the pattern for adding the cache!

src/did/did-resolver.ts
thehenrytsai: Cache is the interface. So as long as object of type Cache is passed in, the resolver can use it the same way regardless of implementation.

Cxxshyy:
hi @thehenrytsai im having trouble identifying the error 
image

Cxxshyy:
image

thehenrytsai:
`resolvers` is being passed in as `undefined` in  `constructor(resolvers: DidMethodResolver[], cache? :Cache)`.

It appears you got rid of a block of code when you pulled and merged main into your branch:

https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/did/did-resolver.ts#L13

Cxxshyy:
thanks another issue 
image
for some reason im getting an error related to a file i never modified

Cxxshyy:
i also tried to write a test if you can also check it thanks its in did-resolver.spec.ts

thehenrytsai:
I'd suggest that given there is not a lot of changes, and I see a some incorrect/unintended changes probably due branch merges.

1. stash/save your changes
2. pull and revert back to latest clean state
3. run tests to make sure everything passes
4. apply/add your changes on top, if tests starts to break, you know is your changes for sure.

Cxxshyy:
Ok all the code ran fine but now when i try using (this.cache = cache || new MemoryCache(600);) it comes up saying (Type 'Cache | MemoryCache' is not assignable to type 'Cache'). Even though it should be allowed

Cxxshyy:
nevermind got it

Cxxshyy:
if possible can close this PR and use the new one i created @thehenrytsai

thehenrytsai:
Happy to! Sorry for the slow response, have been out of town last week but back now.

Issue #68, opened 2022-07-22
Replace verificationMethod checks with JSON Schema based validation (publicJwk)


bnonni:
Replace with JSON Schema based validation for `publicJwk`. More info about the `publicJwk` property can be found here: https://www.w3.org/TR/did-spec-registries/#publicJwk 

https://github.com/TBD54566975/dwn-sdk-js/blob/fcea84929e12dd3e3ef0e8866b0e592a025786fa/src/jose/jws/general/verifier.ts#L74-L84

flothjl:
@thehenrytsai interested in your thoughts on depth of validation in my first pass at this. Initially started with basic checks for required fields, but then decided might be valuable to validate there is no private properties and that the curve is valid for a keytype. 

Thanks!

thehenrytsai:
@flothjl, thanks for tackling this, I've merged in your PR! 🎆 🍾 🎊

PR #129, opened 2022-10-28
#68 publick jwk validation

flothjl:
Initial implementation of JSON schema validation for publicJwk. Supports EC/secp256k1 and OKP/Ed25519 since that is what's currently supported for [verifying signatures](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/jose/algorithms/signing/index.ts)


src/validation/json-schemas/public-jwk.json
```diff
@@ -0,0 +1,68 @@
...
+      {
+        "properties": {
+          "kty": {
+            "const": "OKP"
+          }
+        },
+        "anyOf": [{ "required": ["d"] }]
+      }
+    ]
+  }
```
thehenrytsai, line 67: Would the below be sufficient?
```suggestion
"additionalProperties": false
```

flothjl: @thehenrytsai  the reason I did the above is 3-fold, but could be offbase so let me know your thoughts. 
1. If and  when new curves/keytypes are supported that don't have "d" as a private property
2. There are a lot of different optional properties that can exist in the publicJwk so didn't want to just restrict everything
3. This might allow us to abstract away some of this validation into a common jwk schema that includes private and public properties and then use the "not" above to validate that it's public.

thehenrytsai: You definitely have given it way more thought that I do! Sounds logical!

src/validation/json-schemas/jwk/general-jwk.json
```diff
@@ -0,0 +1,96 @@
...
+    },
+    "x5t#S256": {
+      "type": "string"
+    }
+  },
+  "oneOf": [
+    {
+      "properties": {
+        "kty": {
+          "const": "EC"
```
thehenrytsai, line 42: For my understanding: why is it that `d` is not included in the "generic" JWK schema?

thehenrytsai (approved):

Issue #66, opened 2022-07-22
Add cache support to avoid costly validation of the same signature


bnonni:
Add logic to prevent validating duplicate signatures 

https://github.com/TBD54566975/dwn-sdk-js/blob/fcea84929e12dd3e3ef0e8866b0e592a025786fa/src/jose/jws/general/verifier.ts#L14-L44

thehenrytsai:
We've implemented an [in-memory cache](https://github.com/TBD54566975/dwn-sdk-js/blob/fcea84929e12dd3e3ef0e8866b0e592a025786fa/src/utils/memory-cache.ts) that can be used.

thehenrytsai:
Thanks again @flothjl!

PR #133, opened 2022-11-03
#66 - add cache support to jws verification

flothjl:


src/jose/jws/general/verifier.ts
```diff
@@ -11,22 +13,33 @@ type VerificationResult = {
...
       const publicJwk = await GeneralJwsVerifier.getPublicKey(kid, didResolver);
 
-      const isVerified = await GeneralJwsVerifier.verifySignature(this.jws.payload, signatureEntry, publicJwk);
+      const cachedValue = await this.cache.get(cacheKey);
+
+      if (cachedValue === undefined) {
+        isVerified = await GeneralJwsVerifier.verifySignature(this.jws.payload, signatureEntry, publicJwk);
+        await this.cache.set(cacheKey, isVerified.toString());
+      } else {
+        isVerified = cachedValue === 'true' ? true : false;
```
thehenrytsai, line 40:
```suggestion
        isVerified = cachedValue;
```

src/jose/jws/general/verifier.ts
```diff
@@ -11,22 +13,33 @@ type VerificationResult = {
...
+      const cacheKey = `${signatureEntry.protected}.${this.jws.payload}.${signatureEntry.signature}`;
       const kid = GeneralJwsVerifier.getKid(signatureEntry);
       const publicJwk = await GeneralJwsVerifier.getPublicKey(kid, didResolver);
 
-      const isVerified = await GeneralJwsVerifier.verifySignature(this.jws.payload, signatureEntry, publicJwk);
+      const cachedValue = await this.cache.get(cacheKey);
+
+      if (cachedValue === undefined) {
+        isVerified = await GeneralJwsVerifier.verifySignature(this.jws.payload, signatureEntry, publicJwk);
+        await this.cache.set(cacheKey, isVerified.toString());
```
thehenrytsai, line 38:
```suggestion
        await this.cache.set(cacheKey, isVerified);
```

flothjl: If the Cache provider returns serialized values, then even if we cache `false`, it would return from cache as `'false'` which would be truthy, no? so we wouldn't reverify the signature, but `if(isVerified)` would pass and the signature would get added to the list of signers?

thehenrytsai: The cache shouldn't  morph the data cached, but valid concern, I'd say we can put the `===` on the `if(isVerified)`:

`if (isVerified === true) {`

And add comments above why strict equality is there as it is not immediately obvious. What do you think?

tests/jose/jws/general.spec.ts
```diff
@@ -152,4 +152,57 @@ describe('General JWS Sign/Verify', () => {
...
+      // @ts-ignore
+      resolve: sinon.stub().withArgs('did:jank:alice').resolves(mockResolutionResult)
+    });
+
+    const verifier = new GeneralJwsVerifier(jws);
+
+    const verifySignatureSpy = sinon.spy(GeneralJwsVerifier, 'verifySignature');
+    const cacheSpy = sinon.spy(verifier.cache, 'set');
+
+    const verificationResult = await verifier.verify(resolverStub);
```
thehenrytsai, line 201: Can we change the test slight to call `verify()` twice instead? Would be more obviou/readable.

thehenrytsai (approved):
Thanks for the contribution again @flothjl! 🥇

Issue #113, opened 2022-10-15
Change `dateCreated` descriptor property from epoch time to higher precision format


thehenrytsai:
Requirements:
1. Adopt ISO 8601 UTC format with microsecond precision enforcement, disallowing timezone selector (even 'Z'):
   `2022-10-14T10:20:30.405060` would be an accepted string.
   `2022-10-14T10:20:30.405060Z` would NOT be an accepted string.
   `2022-10-14T10:20:30.405` would NOT be an accepted string.

This allows us to compare time using a direct string compare.

csuwildcat:
Can we even get microsecond on all platforms/envs? I guess you'd just write in 0s if not, right?

thehenrytsai:
Pretty sure all modern machine can get microsecond precision, if not implementor can either fill with 0s, or random generate 3 digits.

rajakash-dev:
Have a doubt , the current format is number and all comparisons happen with the same but to support format like `2022-10-14T10:20:30.405060` we will have to convert the usage of `dateCreated ` everywhere from number to string .Is this expected as a part of this feature?

thehenrytsai:
@rajakash-dev, yes you are correct, that's the change

rajakash-dev:
https://github.com/TBD54566975/dwn-sdk-js/pull/142

thehenrytsai:
@csuwildcat, please update DWN spec accordingly.

csuwildcat:
Will do

Issue #135, opened 2022-11-04
Better schema validation for the coupling between `published` and `datePublished` properties in `CollectionsWrite`


thehenrytsai:
Update `collections-write.json` schema such that:
1. `datePublished` must exist when `published` is `true`
2. `datePublished` must not exist when `published` is `false`

Using either `anyOf` or `oneOf` would probably do the trick, where I suspect `anyOf` is a better perf choice.

A quick search leads to some hint to how this can be done:
https://stackoverflow.com/questions/9029524/json-schema-specify-field-is-required-based-on-value-of-another-field

PR #142, opened 2022-11-06
Changes to date format

rajakash-dev:
Fixes https://github.com/TBD54566975/dwn-sdk-js/issues/113

- Changes `dateCreated` to string
- Converted ISO format to appropriate precision as required 
- Edited the test cases

### DIFF
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -6,3 +6,7 @@
 export function sleep(durationInMillisecond): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, durationInMillisecond));
 }
+
+export function getCurrentDateInHighPrecision(): string {
+  return (new Date().toISOString().replace('Z',Math.floor(Math.random()*(999-100+1)+100).toString())).toString();
+}
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -8,5 +8,5 @@ export function sleep(durationInMillisecond): Promise<void> {
 }
 
 export function getCurrentDateInHighPrecision(): string {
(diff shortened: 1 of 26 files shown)


src/utils/time.ts
```diff
@@ -5,4 +5,8 @@
  */
 export function sleep(durationInMillisecond): Promise<void> {
   return new Promise(resolve => setTimeout(resolve, durationInMillisecond));
+}
+
+export function getCurrentDateInHighPrecision(): string {
+  return (new Date().toISOString().replace('Z',Math.floor(Math.random()*(999-100+1)+100).toString())).toString();
```
thehenrytsai, line 11: Instead of randomly generating microseconds, I think the following would suffice for this PR:
```suggestion
  return new Date().toISOString().replace('Z','000');
```

If you want to go the extra miles, you can either:
1. look into using npm package that are BOTH browser and nodejs compatible; or
2. look into `performance.timeOrigin` and `performance.now()`

But that's optional as you've done the painful work of fixing all references already!

thehenrytsai (changes requested):
Thanks for the contribution @rajakash-dev, I have made a simple suggestion, but rest looks good!

Technically `datePublished` needs to receive the same treatment, but since it's not mentioned in the original issue AND there is a competing PR on `datePublished` now, I will file a separate issue to address it!

rajakash-dev:
I can fix https://github.com/TBD54566975/dwn-sdk-js/issues/135 as well in minimum time , but I see there is a competing PR for the same . Should i wait on the same to fix @thehenrytsai ??

thehenrytsai (approved):
This is good to go, I think we should to add a regex check for the date time in JSON schema, but I will file a separate issue for it.

Thanks again for the contribution! 🥇 🎖️ 🏅

PR #146, opened 2022-11-08
replaced interface method strings

nearlyjuly:
Changed all method strings as per issue ( including PermissionsRequest and PermissionsGrant )


thehenrytsai (changes requested):
CICD checks failed.

PR #148, opened 2022-11-10
Pre-compile json schemas into AJV validators

mistermoe:
Ajv supports generating standalone validation functions from JSON Schemas at compile/build time. These functions can then be used during runtime to do validation without initializing Ajv. It is useful for several reasons:

* to avoid dynamic code evaluation with Function constructor (used for schema compilation) - when it is prohibited by the browser page [Content Security Policy](https://ajv.js.org/security.html#content-security-policy).
* to reduce the browser bundle size - Ajv is not included in the bundle 
* to reduce the start-up time - the validation and compilation of schemas will happen during build time.

the primary motivation behind this PR is the first point noted above. Merging this PR will allow consumers to use DWNs within browser extensions (e.g. janky wallet)


thehenrytsai (approved):
Looks good @mistermoe! Approved! A minor nit/question, do you think it is possible/good idea that we don't check in the generated validator code `generated/precompiled-validators.js` at all?

thehenrytsai (approved):

PR #149, opened 2022-11-10
moved `contextId` to be a sibling property with `descriptor`

thehenrytsai:


thehenrytsai:
@csuwildcat this is in.

PR #156, opened 2022-11-17
update getCurrentDateInHighPrecision to use Temporal API

alexlemons:
This PR updates `getCurrentDateInHighPrecision` to use the upcoming [Temporal API](https://github.com/tc39/proposal-temporal). The Temporal API proposal is in Stage 3 and is not expected to change significantly. I think it is a better alternative to using the Performance API, which is intended for different purposes.

For now it is implemented using the recommended [polyfill](https://github.com/js-temporal/temporal-polyfill).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@ipld/dag-cbor": "7.0.1",
+        "@js-temporal/polyfill": "^0.4.3",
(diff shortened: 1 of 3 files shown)


csuwildcat:
This looks like a good way to go, thanks for filing this Issue!

thehenrytsai (approved):
@alexlemons this is fantastic, thanks a lot! 🥇 💯 🍾

PR #157, opened 2022-11-21
moved `recordId` to be a sibling property with `descriptor`

thehenrytsai:
1. moved `recordId` to be a sibling property with `descriptor`
2. added `curly`  linter rule
3. renamed `validateSchema()`


thehenrytsai:
@csuwildcat this is now in.

PR #162, opened 2022-12-07
#140 - moved `target` out of `descriptor` into `authorization`

thehenrytsai:


csuwildcat:
@thehenrytsai is the target needed to be able to address a DWN with any message? If so, what happens when we allow unauthenticated queries?

thehenrytsai:
@csuwildcat, yes, `target` is always needed by DWN to route message correctly. Let's sync on unauthenticated queries and move `target` (again) if needed.

Issue #107, opened 2022-10-10
Remove the need for `-es-module-specifier-resolution=node` flag in node.js projects using the SDK


thehenrytsai:
Currently our source code import files without adding `.js` extension. As a side effect, all node.js projects using the SDK need to use the `-es-module-specifier-resolution=node` flag so that imports will resolve correctly.

e.g. https://github.com/TBD54566975/dwn-js/blob/main/package.json#L36

Investigate to see if there is an elegant way so such that this flag is not necessary. The ugly fallback fix is to add `.js` to all import statements referencing internal files.

PR #163, opened 2022-12-09
#107 - removed the need for es-module-specifier-resolution=node flag in node.js projects

thehenrytsai:
1. Removed the need for es-module-specifier-resolution=node flag in node.js projects
2. Updated `tsconfig` so that error will be thrown at compile time if .js extension is not added for imports with relative paths


mistermoe:
huuuge! (pronounced yooooooj). @thehenrytsai  should we test this by pushing up an alpha and using it in `dwn-js` or `dwn-relay` to make sure it works as expected or have you already done that

thehenrytsai:
> huuuge! (pronounced yooooooj). @thehenrytsai should we test this by pushing up an alpha and using it in `dwn-js` or `dwn-relay` to make sure it works as expected or have you already done that

It was really honest work. Build `0.0.13-unstable-20221208-1` was it, it appeared to work in both node and browser. Will cut out a "real" v0.0.13 once this goes in.

Issue #145, opened 2022-11-08
Change `datePublished` descriptor property from epoch time to higher precision format


thehenrytsai:
Requirements:
1. Adopt ISO 8601 UTC format with microsecond precision enforcement, disallowing timezone selector (even 'Z'):
   `2022-10-14T10:20:30.405060` would be an accepted string.
   `2022-10-14T10:20:30.405060Z` would NOT be an accepted string.
   `2022-10-14T10:20:30.405` would NOT be an accepted string.

This allows us to compare time using a direct string compare.

See issue #113 and PR #142 for the same change on `dateCreated`.

NikolaGundzikova:
Hi, can be this issue assigned to me?

thehenrytsai:
Happy to!

alexlemons:
Using the Performance API a precision string could potentially be generated using something like:

```js
function getCurrentDateInHighPrecision() {
	const now = (performance.timeOrigin + performance.now()).toFixed(3).split('.');
	return new Date(Number(now[0])).toISOString().replace('Z', now[1]);
}
```

However, Im not convinced that these values will be consistent and comparable when they're being generated in different environments/browsers, as there seems to be some variation in how the performance API is implemented.
Chrome currently has a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1365808) related to `performance.timeOrigin` which results in the date created being incorrect.

alexlemons:
When ready the new Temporal API will include [Temporal.Instant](https://tc39.es/proposal-temporal/docs/instant.html) with nanosecond precision.

thehenrytsai:
Change merged.

Issue #167, opened 2022-12-14
Implement `dateSort` for CollectionsQuery


thehenrytsai:
Currently providing `dateSort` in CollectionsQuery will cause error to be thrown: https://github.com/TBD54566975/dwn-sdk-js/blob/2dbf7e723b71e0630cda85a2d7a761bb96689970/src/interfaces/collections/messages/collections-query.ts#L36

Implement the functionalities as [spec-ed](https://identity.foundation/decentralized-web-node/spec/#collectionsquery).

thehenrytsai:
implemented

PR #174, opened 2022-12-18
Implement DateSort #167

gtaylor5:
### Problem
`dateSort` has not yet been implemented for `CollectionsQuery` message types. Today, if a `CollectionQuery` contains the `dateSort` field the request will throw an error. We want to implement this behavior. ref: #167 

### Solution

The solution to this problem entails the following:

1. Add a new `enum DateSortName` to encapsulate the raw string values that can be passed in.
2. Implement a `handleDateSort` function that accepts a `collectionQuery` and an array of `BaseMessage`.

The `handleDateSort` function extracts the `dateSort` value, and casts the BaseMessage type to a `CollectionWriteMessage` to gain access to the properties required to sort / filter entries. Then, we switch on `dateSort` and sort accordingly.

### Result
We can now support a `CollectionQuery` that contains a `dateSort` field.

### DIFF
--- a/src/interfaces/collections/handlers/collections-query.ts
+++ b/src/interfaces/collections/handlers/collections-query.ts
@@ -170,43 +170,39 @@ async function handleDateSort(
   entries: BaseMessage[]
 ): Promise<BaseMessage[]> {
 
-  const publishedFilter = (message): boolean => {
-    return message.descriptor.published;
-  };
-
-  const compareDateCreatedAscending = (left, right): number => {
-    return Number(left.descriptor.dateCreated) - Number(right.descriptor.dateCreated);
-  };
-
-  const compareDateCreatedDescending = (left, right): number => {
-    return Number(right.descriptor.dateCreated) - Number(left.descriptor.dateCreated);
-  };
-
-  const compareDatePublishedAscending = (left, right): number => {
-    return Number(left.descriptor.datePublished) - Number(right.descriptor.datePublished);
-  };
-
-  const compareDatePublishedDescending = (left, right): number => {
-    return Number(right.descriptor.datePublished) - Number(left.descriptor.datePublished);
-  };
-
   const { dateSort } = collectionsQuery.message.descriptor;
   const collectionMessages = entries as CollectionsWriteMessage[];
 
   switch (dateSort) {
   case DateSortName.CreatedAscending:
-    return collectionMessages.sort(compareDateCreatedAscending);
+    return collectionMessages.sort(getCompareByPropertyFn('dateCreated', 'asc'));
   case DateSortName.CreatedDescending:
-    return collectionMessages.sort(compareDateCreatedDescending);
-  case DateSortName.PublishedAscending:
+    return collectionMessages.sort(getCompareByPropertyFn('dateCreated', 'desc'));
+  case DateSortName.PublishedAscending: 
     return collectionMessages
-      .filter(publishedFilter)
-      .sort(compareDatePublishedAscending);
-  case DateSortName.PublishedDescending:
+      .filter(m => m.descriptor.published)
+      .sort(getCompareByPropertyFn('datePublished', 'asc'));
+  case DateSortName.PublishedDescending: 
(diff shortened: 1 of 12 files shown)


thehenrytsai (changes requested):
@gtaylor5, this is super cool! A couple general comments:

1. Make sure to run `npm run lint` to find and fix linter errors.
2. Need to add tests to cover the tests added.

I will try to update this PR myself if you can't get to the above in time. But I will first want to fix #145 first which I will do next.

gtaylor5:
Okay, great I'll run the lints and update the PR. I'll also look into adding tests as well. Thanks!

gtaylor5:
One thing my PR assumes is that the `created` and `published` dates are Unix timestamps. If that's not the case, then I need to make some updates.

td-bn:
I was taking a look at this issue and noticed that the dateSort field is not nested within the filter property in the current implementation
https://github.com/TBD54566975/dwn-sdk-js/blob/7b33a61fabb61c8d7f1de5404475bc1ba03d8243/src/interfaces/collections/types.ts#L37

Whereas it is [in the spec](https://identity.foundation/decentralized-web-node/spec/#collectionsquery)

thehenrytsai:
> dateSort field is not nested within the filter property in the current implementation... Whereas it is [in the spec](https://identity.foundation/decentralized-web-node/spec/#collectionsquery)

Good observation @td-bn, I believe the spec needs to be updated as `dateSort` is not a filter. Adding @csuwildcat the spec owner for awareness.

thehenrytsai (approved):

thehenrytsai:
The code is now good to merge. @gtaylor5 thanks again for implementing the feature, your contribution is highly appreciated!! 🥇 🙏 👏 

@csuwildcat, for spec work, the implementation filters out records that are not published if `dateSort` is on `datePublished`.

src/utils/string.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Compares two string given in lexicographical order.
+ * @returns 1 if `a` is larger than `b`; -1 if `a` is smaller/older than `b`; 0 otherwise (same message)
+ */
+export function lexicographicalCompare(a: string, b: string): number {
```
dcrousso, line 5: any reason we can't use something like this?
```
a.localeCompare(b, undefined, { numeric: true })
```
(the `numeric: true` makes it so that `"1" < "2" < "10"` instead of `"1" < "10" < "2"`)

src/interfaces/collections/handlers/collections-query.ts
```diff
@@ -153,4 +159,42 @@ async function fetchUnpublishedRecordsByRequester(collectionsQuery: CollectionsQ
...
+  collectionsQuery: CollectionsQuery,
+  entries: BaseMessage[]
+): Promise<BaseMessage[]> {
+
+  const { dateSort } = collectionsQuery.message.descriptor;
+  const collectionMessages = entries as CollectionsWriteMessage[];
+
+  switch (dateSort) {
+  case DateSort.CreatedAscending:
+    return collectionMessages.sort((a, b) => lexicographicalCompare(a.descriptor.dateCreated, b.descriptor.dateCreated));
```
dcrousso, line 188: are `dateCreated` supposed to be a timestamp in string format?  if so, would it be easier/better to `parseInt` and compare numerically instead?

thehenrytsai: `dateCreated` is in ISO 8061 format, not unix epoch time. e.g.

`2022-12-21T10:20:30.405060`

Not sure what `{ numeric: true }` will do with `localCompare` but don't think we can use `parseInt`.

src/utils/string.ts
thehenrytsai: I didn't use `localeCompare` as the context of the string compare here is not locale dependent and I don't know if there are implications of using `localCompare` in different locales so choose to not bother with it.

thehenrytsai (approved):

PR #176, opened 2022-12-27
Fix did ion resolver

mistermoe:
# Summary
* Updated resolution endpoint to `beta.discover.did.microsoft.com`. `discover.did.microsoft.com` wasn't working
image

* Fixed an issue that was causing ION DID resolution to bork in browser extensions.
    * `cross-fetch` is a [ponyfill](https://github.com/sindresorhus/ponyfill) that uses `XMLHTTPRequest` under the hood in browser environments. Unfortunately, `XMLHTTPRequest` cannot be used in MV3 browser extension background service workers. browser extensions get even more strict with `fetch` in that it cannot be referenced indirectly. 

Screenshot is a console in the context of a background service worker. Running `XMLHTTPRequestTest` shows that use of `XMLHTTPRequest` is not allowed. `IndirectReferenceTest` shows inability to use fetch when referenced indirectly

image

### DIFF
--- a/src/did/did-ion-resolver.ts
+++ b/src/did/did-ion-resolver.ts
@@ -1,13 +1,18 @@
 import crossFetch from 'cross-fetch';
 import { DidMethodResolver, DidResolutionResult } from './did-resolver.js';
 
+// supports fetch in: node, browsers, and browser extensions.
+// uses native fetch if available in environment or falls back to a ponyfill.
+// 'cross-fetch' is a ponyfill that uses `XMLHTTPRequest` under the hood.
+// `XMLHTTPRequest` cannot be used in browser extension background service workers.
+// browser extensions get even more strict with `fetch` in that it cannot be referenced
+// indirectly.
+const fetch = globalThis.fetch ?? crossFetch;
+
 /**
  * Resolver for ION DIDs.
  */
 export class DidIonResolver implements DidMethodResolver {
-  // cross-platform fetch
-  private fetch = crossFetch;
-
   /**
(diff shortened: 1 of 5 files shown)


src/did/did-ion-resolver.ts
```diff
@@ -1,17 +1,22 @@
...
  */
 export class DidIonResolver implements DidMethodResolver {
-  // cross-platform fetch
-  private fetch = crossFetch;
-
   /**
    * @param resolutionEndpoint optional custom URL to send DID resolution request to
    */
-  constructor (private resolutionEndpoint: string = 'https://discover.did.microsoft.com/1.0/identifiers/') { }
+  constructor (private resolutionEndpoint: string = 'https://beta.discover.did.microsoft.com/1.0/identifiers/') { }
```
thehenrytsai, line 19: Thanks for the update @mistermoe, sanity: should we still use the GA endpoint here though? Ot it stopped working?

mistermoe: It stopped working 😢. Check out the screenshot i posted in the PR description of curl failing

thehenrytsai: Hm... just tried and it works again. I guess it's still best to reply on their GA endpoint instead of beta? Though you seemed to find the beta endpoint more reliable....

mistermoe: all good @thehenrytsai! thanks for figuring it out. went ahead and updated the resolution endpoint

thehenrytsai (approved):

PR #177, opened 2023-01-03
npm release automation from a github release

michaelneale:
this is the first pass at release automation for npm - using the off the shelf github action recommendation.

To use this - you "simply" create a release on the github side - it takes care of the rest.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,33 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - run: npm test
+
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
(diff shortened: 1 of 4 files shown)


michaelneale:
cc @thehenrytsai @frankhinek for your consideration

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
```
thehenrytsai, line 18: Sanity: there isn't a `ci` script

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+      - run: npm test
```
thehenrytsai, line 19: Sanity:
```suggestion
      - run: npm run test:node:ci
```

thehenrytsai (changes requested):
Looks like this is the first attempt at the automated release and needing for the code to be checked in to actually test it.

michaelneale:
I need to make some changes: don't really want to run tests as part of this - @thehenrytsai what command do you run to npm publish today?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
```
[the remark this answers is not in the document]
michaelneale: > Looks like this is the first attempt at the automated release and needing for the code to be checked in to actually test it.

yes - I was testing this in a different repo with a dummy package first.

thehenrytsai:
I do an `npm version path/minor` followed by `npm publish` for "official" release. For "unstable" release, I manually update the version in `pacakge.json` to look like [`0.0.13-unstable-20221208-1`](https://www.npmjs.com/package/@tbd54566975/dwn-sdk-js/v/0.0.13-unstable-20221208-1), then `npm publish`.

As of now, I am more interested in the automation for the unstable release (say after each commit to `main` branch), but this is a great first step!

thehenrytsai:
For reference, [this](https://github.com/decentralized-identity/sidetree/blob/master/.github/workflows/unstable-release.yml) is what `sidetree` repo did for unstable release a couple of years back, the state of art might have changed now.

michaelneale:
ok @thehenrytsai please take another look - borrowed the sidetree approach so that main deploys will be deployed as unstable but with a handy timestamp (and proper releases can be created from github releases - but it will just use the package.json version in that case) - as I think that releases can be "press the button" for now

thehenrytsai (approved):
Let's give this a go! 🎉 🎊

PR #178, opened 2023-01-04
Example of documentation using Typedoc

nearlyjuly:
(The @types dependency needs to be updated for the following to work)
I've added some entries to the tsconfig.json file so that running Typedoc will create a set of pages such as the example here. So it could be run after each commit (simplest way) or there are more complex options as per https://typedoc.org/guides/options/ 

The documentation was generated by installing dwn-sdk locally, updating @types and then npx typedoc.


thehenrytsai:
Hey @nearlyjuly, took a look at the generated TS Docs, and it looks fantastic. As you suggested, it would be a great starting point if it can be automatically updated after each commit!

nearlyjuly:
Thank you so much! I will work on it some more tomorrow and this weekend and see how to do this

pdxjohnny (commented):
@nearlyjuly I poked around out of curiosity (I got excited when I saw your open docs PR and couldn't way to read the docs) https://github.com/nearlyjuly/dwn-sdk-js/pull/5

🥳

image

nearlyjuly:
Thank you! I'm not sure how to proceed, do I just approve the merge request?

michaelneale:
super cool @nearlyjuly  - looks good to me. In theory if @thehenrytsai gives a 👍 we can merge. I am not sure how to best test the gh-pages github action but it looks good (I did something similar recently).

Issue #182, opened 2023-01-05
introduce mutable `dateModified` to disambiguate from immutable `dateCreated` property


thehenrytsai:

thehenrytsai:
Implemented.

PR #189, opened 2023-01-11
#182 - introduced dateModified to disambiguate from dateCreated

thehenrytsai:
1. introduced dateModified to disambiguate from dateCreated
2. added convenience method `createLineageChild()`


thehenrytsai:
@csuwildcat, this is now in. Spec might need to be updated.

PR #196, opened 2023-01-16
add pages link for api docs

michaelneale:


michaelneale:
minor README change so I am merging.

PR #197, opened 2023-01-17
renamed Collections interface to Records

thehenrytsai:
1. The word "collection" no longer exists in the code base. Not even once.
2. Added a few more useful exports.


thehenrytsai:
@csuwildcat, this is now in. Be sure to not have the word "collection" in spec also!

Issue #199, opened 2023-01-24
Don't mix operations on BigInts and numbers in one arithmetical expression or use unsigned right shift operators on BigInts


shamilovtim:
### Problem:
Runtime exceptions present in Hermes and JSC with errors: `TypeError: Conversion from 'BigInt' to 'number' is not allowed.`

## Solution:
Don't mix operations on BigInts and numbers in one arithmetical expression
or use unsigned right shift operators on BigInts

[More info here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cant_convert_BigInt_to_number)

These operations cause incompatibility with:
- JSC
- Hermes
- Firefox

PR #200, opened 2023-01-24
Browser Bundle Improvements

mistermoe:
Addresses #199


mistermoe:
halved the bundle size by importing the 1 lodash function we're using instead of the entire lib
image

mistermoe:
included `firefox108` and `safari16` as build targets for browser bundle.

mistermoe:
configured `karma` to run tests in `WebkitHeadless` and `FirefoxHeadless`

mistermoe:
switched node stdlib polyfill lib from [@esbuild-plugins/node-globals-polyfill](https://www.npmjs.com/package/@esbuild-plugins/node-globals-polyfill) to [node-stdlib-browser](https://github.com/niksy/node-stdlib-browser). 

`node-globals-polyfill` had us stuck on `esbuild` version `0.14.39`. esbuild is on `0.17.4` now with some relatively large improvements including support for `hermes` as a build target. this is particularly useful for us as `hermes` is on its way to becoming the defacto JS engine for react native. 

Still blocked on bumping all the way up to `0.17.4` as it's not supported by [karma-esbuild](https://github.com/marvinhagemeister/karma-esbuild) yet which we use to test our browser bundle in headless browsers. [karma-esbuild PR](https://github.com/marvinhagemeister/karma-esbuild/pull/55) to support `esbuild` `0.17` is up though and approved. just waiting to be merged.

additionally, `node-globals-polyfill` seems largely unmaintained. [relevant issue here](https://github.com/remorses/esbuild-plugins/issues/1)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tbd54566975/dwn-sdk-js",
-  "version": "0.0.17",
+  "version": "0.0.18",
```
shamilovtim, line 3: is the `main` entrypoint still on the table to add?

mistermoe:
published `v0.0.18-unstable.2c05feb-2023.1.24-11-24-27` to npm so that we can test. need to test the following environments:
* cjs (node)
* esm (node)
* browser
* react-native

@thehenrytsai can you pull this branch down and run `npm run test:browser` to make sure it works on your end?

package.json
mistermoe: just added! need to test and make sure it works now

mistermoe:
fixed an issue. new unstable version is `0.0.18-unstable.c04518c-2023.1.24-11-57-22`

thehenrytsai:
> @thehenrytsai can you pull this branch down and run `npm run test:browser` to make sure it works on your end?
@mistermoe all tests passed on my end, with a nit that only the last batch of WebkitHeadless test results seemed to have printed out to console.

package.json
```diff
@@ -1,15 +1,17 @@
...
   "description": "A reference implementation of https://identity.foundation/decentralized-web-node/spec/",
   "type": "module",
   "types": "./dist/esm/src/index.d.ts",
   "engines": {
     "node": ">= 16"
   },
+  "main": "./dist/cjs/src/index.js",
```
thehenrytsai, line 10: For my learning: what's the reason for bringing back building cjs? Is this something required by react native?

thehenrytsai (approved):

package.json
shamilovtim: @thehenrytsai related:

https://github.com/facebook/metro/issues/670

shamilovtim: RN **does** use ESM but for whatever reason just doesn't support the `exports` field yet

shamilovtim, line 10: just wanted to comment that i'm not sure we necessarily need a `cjs` bundle. the esm ones should work in theory, as RN uses esm. you could try pointing the `main` field to the esm, or if that's a non starte you could try adding the [react-native directive](https://github.com/facebook/metro/issues/670#issuecomment-1099926849) instead of `main`

src/store/message-store-level.ts
```diff
@@ -3,7 +3,7 @@ import type { BaseMessage, DataReferencingMessage } from '../core/types.js';
 
 import * as block from 'multiformats/block';
 import * as cbor from '@ipld/dag-cbor';
-import _ from 'lodash';
+import isPlainObject from 'lodash/isPlainObject.js';
```
shamilovtim, line 6: good call

shamilovtim: we can add an eslint rule in the future that prohibits importing all of lodash:

```json
   "no-restricted-imports": [
      "error",
      {
        "name": "lodash",
        "message": "Dont import entire lodash. Use a module instead, for example: import _uniqBy from 'lodash/uniqBy';"
      },
      "error",
      {
        "name": "lodash-es",
        "message": "Dont import entire lodash. Use a module instead, for example: import _uniqBy from 'lodash/uniqBy';"
      }
    ]
```

package.json
mistermoe: @thehenrytsai i felt bad for including a `main` field that wasn't `cjs` because that is a widely accepted convention.

package.json
shamilovtim: I can confirm the `main` entrypoint works however wonder if you definitely want cjs given that react native should work with esm and via the `react-native` directive instead of `main`

package.json
mistermoe: nice find @shamilovtim! i'll def do that instead. @thehenrytsai am going to yank out the cjs before merging this

shamilovtim (approved):

package.json
mistermoe: actually going to leave the cjs as electron doesn't yet support esm. @shamilovtim any preference on whether you want the react native directive to point to esm or cjs? i imagine ejs but just wanted to be sure

shamilovtim: probably esm. no idea about ejs file types specifically

thehenrytsai: K, based on the parallel conversations, I take it that `cjs` is now needed not because of metro/react, but because of electron. Got it.

thehenrytsai (approved):
Seriously, what do we do without you! 🥇

shamilovtim (approved):

Issue #179, opened 2023-01-04
Remove `target` from message data model


csuwildcat:
We can get rid of `target` and the message reformulation it inflicts by moving to this tweak to authn/storage evaluation logic:

```js

// Is the signer a tenant, and does the signer want the message to be store-able in their own DWN?
if (messageSigner === signerIsTenant && selfStorage === true) {
  storeThatShiz();
}

// Is a permission present, and is the permission grantor a tenant?
if (hasPermission && permissionGrantedBy === grantorIsTenant) {
  storeThatShiz();
}

// Is it a Protocols message, and does it meet any of the actor requirements?
if (isProtocolMessage) {
  if (signerIsTenant && signerHasProtocol && isValidAgainstProtocolRules) {
   storeThatShiz();
  }
  if (recipientIsTenant && recipientHasProtocol && isValidAgainstProtocolRules) {
   storeThatShiz();
  }
}
```

thehenrytsai:
There have been a change to direction of the original post. Instead of opting for an implied logic, `target` will be required but passed to DWN external to the messages themselves.

thehenrytsai:
Implemented.

PR #201, opened 2023-01-25
#179 - removed 'target' property from message data model

thehenrytsai:


thehenrytsai:
@csuwildcat, this is now in, be sure to reflect in spec if needed.

PR #204, opened 2023-01-28
split method into interface and method

thehenrytsai:


thehenrytsai:
@csuwildcat, this is now in, the spec will need an update to reflect this.

Issue #209, opened 2023-02-02
Support `attestation` property for RecordsWrite and RecordsQuery


thehenrytsai:

thehenrytsai:
Implementation merged.

PR #210, opened 2023-02-02
#209 - added support for attestation property in records

thehenrytsai:
1. RecordsWrite can now include an `attestation` property
2. `attestation` property is returned in results of a RecordsQuery

Will add support for `attester` in RecordsQuery filter in the next PR.

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -1,4 +1,4 @@
-import type { GeneralJws, SignatureInput } from '../jose/jws/general/types.js';
+import type { GeneralJws } from '../jose/jws/general/types.js';
 
 /**
  * Intersection type for all concrete message types.
@@ -41,8 +41,4 @@ export type DataReferencingMessage = {
   };
 
   encodedData: string;
-};
-
-export type AuthCreateOptions = {
-  signatureInput: SignatureInput
 };
--- a/src/dwn.ts
+++ b/src/dwn.ts
@@ -68,7 +68,7 @@ export class Dwn {
       message = JSON.parse(requestString);
     } catch (error) {
       return new MessageReply({
-        status: { code: 400, detail: error.message }
+        status: { code: 400, detail: 'unable to JSON parse request' }
       });
     }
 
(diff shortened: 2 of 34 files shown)


src/dwn.ts
```diff
@@ -68,7 +68,7 @@ export class Dwn {
       message = JSON.parse(requestString);
     } catch (error) {
       return new MessageReply({
-        status: { code: 400, detail: error.message }
+        status: { code: 400, detail: 'unable to JSON parse request' }
```
andresuribe87, line 71: Should the bytes to string be moved outside the try catch block?

src/core/types.ts
```diff
@@ -42,7 +42,3 @@ export type DataReferencingMessage = {
 
   encodedData: string;
 };
-
-export type AuthCreateOptions = {
```
andresuribe87, line 46: Intentional? Might break users of the sdk

src/interfaces/records/messages/records-write.ts
```diff
@@ -28,14 +28,18 @@ export type RecordsWriteOptions = AuthCreateOptions & {
   published?: boolean;
   datePublished?: string;
   dataFormat: string;
+  authorizationSignatureInput: SignatureInput;
+  attestationSignatureInputs?: SignatureInput[];
```
andresuribe87, line 32: I find it strange to have an optional type over an array. Is this standard practice in TS ?

andresuribe87 (approved):

src/interfaces/records/messages/records-write.ts
thehenrytsai: This tries to make dev experience extra nicer, so that when they don't care about attestation, they are don't have to give and empty array just to make the API happy.

src/core/types.ts
thehenrytsai: Yes, intentional. This is not used by users, though even if it is, we break users consistently currently right now while we undergo rapid changes and development. Better to break now than later when there are "actually" users by the thousands.

src/dwn.ts
thehenrytsai: Not worth it at the moment, if it's not able to convert bytes to string, it's not able to JSON parse.

Issue #198, opened 2023-01-23
Finalize and implement data ingestion mechanism


thehenrytsai:
Currently data is ingested as base64url encoded property in the JSON message. This will not be acceptable if/when DWN is used to store large data.

A new proposal is to:
- [x] Remove support for multiple messages in one DWN request in favor of just one message per request
- [x] move the JSON metadata to an HTTP header
- [x] Add a new optional parameter `stream` to `processMessage(tenant, message, stream)`
- [x] Remove `encodedData` from being allowed in "write" messages
- [x] Return `encodedData` in message if it is under a certain size limit
- [x] Make `DataStore` a peer of `MessageStore`, another thin class/method `StorageController` that will manage the set/get of message and data
- [x] Change indexed key of data to tenant scoped [tenant + `messageCid` + (`commitId`, but this is more TODO for the future)]
- [x] Introduce `RecordsRead` message that will look something like:
```
{
  descriptor: {
    recordId: 'abc'
  }
}
```
- [x] Update the return type of `processMessage(...)` `MessageReply` to allow an optional stream so that `RecordsRead` can return data stream
- [x] Change the Node server code to 
   - Move the JSON metadata to an HTTP header
   - Leave the body for exclusive use to stream data

andorsk:
@thehenrytsai totally agree with this. Great proposal. Related to "streaming", could eventually other streaming protocols ( like grpc ) be supportable? No rush or immediate need on this, but just want to know if it's possible down the road.

mistermoe:
need to change API signature of `dwn.processMessage(message: any)` to `dwn.processMessage(message: any, data?: ReadableStream)`

[ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) looks to be supported across all major browsers and node. @shamilovtim is `ReadableStream` supported in react native?

mistermoe:
@andorsk entirely possible _if_ we can figure out a transport agnostic serialization scheme. currently it is transport agnostic. the moment we end up using http headers or anything specific to http (aka multipart uploads) we'll sort of be tying ourselves to http and each individual transport will need to figure out how to support DWNs.

andorsk:
@mistermoe thanks! I would be happy to help work on efforts to figure out a transport agnostic approach. I'm guessing there will be many implementors that will not want to bind their services to HTTP ( for various reasons ) and it would be great if we can not lock them out of DWN adoption cycles. 

[...]

PR #220, opened 2023-02-11
#198 - changed processMessage(...) to accept data using a readable stream

thehenrytsai:
This is the first PR of multiple iterations to revamp how data is ingested and returned in queries. This PR includes:

1. Added a new optional `dataStream` parameter stream to `processMessage(...)`.
1. Removed encodedData from being allowed in "write" messages.
1. Added a new optional `dataStream` parameter stream to `MessageStore.put(...)` method, this maybe temporary.
1. Added optional parameter `dataCid` to `RecordsWrite`'s `create(...)` and `createFrom(...)` to improve dev experience.
1. Introduced `DwnError` and `DwnErrorCode` as it was needed for the processing logic.
1. Minor change to `MethodHandler` interface.


andresuribe87 (approved):
Should the new release be backward incompatible?

src/core/dwn-error.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * A class that represents a DWN error.
+ */
+export class DwnError extends Error {
+  constructor (public code: string, message: string) {
```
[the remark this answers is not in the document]
thehenrytsai: I am opinionated on this, deliberately chose string because number is meaningless.

src/index.ts
```diff
@@ -18,6 +18,7 @@ export { RecordsDelete, RecordsDeleteOptions } from './interfaces/records/messag
 export { RecordsQuery, RecordsQueryOptions } from './interfaces/records/messages/records-query.js';
 export { RecordsWrite, RecordsWriteOptions, CreateFromOptions } from './interfaces/records/messages/records-write.js';
 export { DateSort } from './interfaces/records/messages/records-query.js';
+export { DataStream } from './utils/data-stream.js';
```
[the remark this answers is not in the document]
thehenrytsai: This is the file that declares lib classes what not for external usage, and I imagine `DataStream` utility will be a very useful one going forward.

src/interfaces/records/handlers/records-write.ts
```diff
@@ -74,7 +72,19 @@ export const handleRecordsWrite: MethodHandler = async (
...
     const indexes = await constructRecordsWriteIndexes(tenant, recordsWrite, isLatestBaseState);
 
-    await messageStore.put(incomingMessage, indexes);
+    try {
+      await messageStore.put(incomingMessage, indexes, dataStream);
+    } catch (error) {
+      if (error.code === DwnErrorCode.MessageStoreDataCidMismatch ||
+          error.code === DwnErrorCode.MessageStoreDataNotFound) {
+        return new MessageReply({
+          status: { code: 400, detail: error.message }
```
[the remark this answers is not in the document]
thehenrytsai: Yes there is, but quite low in priority at the moment, you will be doing something like `StatusCode.400` or `StatusCode.BadRequest` instead of 400...

src/core/dwn-error.ts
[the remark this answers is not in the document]
andresuribe87: Thinking more about this... I think it makes sense since it could be something like `EX1991` or similar.

CONTRIBUTING.md
```diff
@@ -39,8 +39,13 @@ We plan on including a Docker container to support all local development soon.
   * This is run via CI whenever a pull request is opened, or a commit is pushed to a branch that has an open PR
 * Running the `npm run test:browser` command from the root of the project will run the tests in a browser environment
   * Please make sure there are no failing tests before switching your PR to ready for review! We hope to have this automated via a github action very soon.
+
 ### Code Style
 Our preferred code style has been codified into `eslint` rules. Feel free to take a look [here](https://github.com/TBD54566975/dwn-sdk-js/blob/main/.eslintrc.cjs). Running `npm run lint` will auto-format as much as `eslint` can. Everything it wasn't able to will be printed out as errors or warnings. Please make sure to run `npm run lint` before switching your PR to ready for review! We hope to have this automated via a github action very soon.
+
+### Code Guidelines
+1. A `TODO` in comment must always link to a GitHub issue.
```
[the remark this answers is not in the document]
thehenrytsai: Great idea! Definitely would be onboard if there's a rule for that. Looking...

src/core/dwn-error.ts
shamilovtim, line 5: just an unimportant nit but going with exotic JS stuff like private/public constructors, public/private constructor variables or static functions in classes increases the amount of exotic code we have to transpile with babel with little value gained for it

CONTRIBUTING.md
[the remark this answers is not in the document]
thehenrytsai: Done. Found a custom plugin that does it!

src/utils/data-stream.ts
```diff
@@ -0,0 +1,48 @@
+import { Encoder } from './encoder.js';
+import { Readable } from 'readable-stream';
```
shamilovtim, line 2: so on slack I thought we were using `ReadableStream` the global. however this seems to be a package, presumably it's a polyfill for/from node envs? i'm not sure how this is going to work on alternative runtimes. but presumably since it's here, it works with web? i think I need some more context to understand

src/store/message-store.ts
```diff
@@ -1,5 +1,6 @@
 import type { BaseMessage } from '../core/types.js';
 import { RangeCriterion } from '../interfaces/records/types.js';
+import { Readable } from 'readable-stream';
```
shamilovtim, line 3: `import type` here

src/utils/data-stream.ts
shamilovtim, line 2: `import type` might be more clear here

src/utils/data-stream.ts
```diff
@@ -0,0 +1,48 @@
+import { Encoder } from './encoder.js';
```
shamilovtim, line 1: i've had trouble when consuming libraries that have file extensions on imports like this. is it possible to leave off file extensions on our imports for JS?

shamilovtim: looks like that's outside the scope of this PR because it's done everywhere else. will file a separate ticket

src/dwn.ts
```diff
@@ -4,13 +4,12 @@ import type { MessageStore } from './store/message-store.js';
...
 
 import { DidResolver } from './did/did-resolver.js';
-import { Encoder } from './utils/encoder.js';
 import { Message } from './core/message.js';
 import { MessageReply } from './core/message-reply.js';
 import { MessageStoreLevel } from './store/message-store-level.js';
-
 import { PermissionsInterface } from './interfaces/permissions/permissions-interface.js';
 import { ProtocolsInterface } from './interfaces/protocols/protocols-interface.js';
+import { Readable } from 'readable-stream';
```
shamilovtim, line 12: `import type`

src/core/dwn-error.ts
thehenrytsai: This is more of a syntax-sugar from TypeScript so that you can declare members more concisely:

https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

But I can sense your view in general on exotic add-ons to JS.

src/utils/data-stream.ts
thehenrytsai: You are right, it is a package: https://www.npmjs.com/package/readable-stream

I did not dig in too much on how they work their magic, I only know that:
1. It's meant to work in browsers by design
1. DWN tests (now run on chrome, firefox, and safari) confirms above
1. polyfill is not required from v4+
1. got 100M+ downloads a week

The above were good enough of an indicator for me to depend on it in this PR. But not sure about any other gotchas specific to RN etc.

src/store/message-store.ts
thehenrytsai: Done.

src/utils/data-stream.ts
thehenrytsai: Can't import just type here, because we do use it beyond just a type declaration, e.g. the constructor call.

src/dwn.ts
thehenrytsai: Done.

src/interfaces/records/messages/records-write.ts
```diff
@@ -78,15 +79,19 @@ export class RecordsWrite extends Message {
   public static async create(options: RecordsWriteOptions): Promise<RecordsWrite> {
     const currentTime = getCurrentTimeInHighPrecision();
 
-    const dataCid = await computeDagPbCid(options.data);
+    if (options.data !== undefined && options.dataCid !== undefined) {
+      throw new Error('only one parameter between `data` and `dataCid` is allowed');
+    }
+    const dataCid = options.dataCid ?? await computeDagPbCid(options.data);
```
dcrousso, line 85: is it an issue that above you only check for `undefined` but here you also check for `null`?

src/interfaces/records/messages/records-write.ts
```diff
@@ -198,12 +201,12 @@ export class RecordsWrite extends Message {
...
       dataFormat                  : unsignedMessage.descriptor.dataFormat,
-      // mutable properties below, if not given, inherit from message given
+      // mutable properties below
       dateModified                : options.dateModified ?? currentTime,
       published,
       datePublished,
-      // copying data from the given message may not be always right, there is probably opportunity for improvement here
-      data                        : options.data ?? Encoder.base64UrlToBytes(unsignedMessage.encodedData),
+      data                        : options.data,
+      dataCid                     : options.data ? undefined : unsignedMessage.descriptor.dataCid, // if data not given, use dataCid from base message
```
dcrousso, line 209: ditto :85

src/utils/data-stream.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Utility class for readable data stream, intentionally named to disambiguate from ReadableStream, readable-stream, Readable etc.
+ */
+export class DataStream {
+  /**
+   * Reads the entire readable stream given into array of bytes.
+   */
+  public static async toBytes(readableStream: Readable): Promise<Uint8Array> {
+    return new Promise((resolve, reject) => {
+      const chunks = [];
+      readableStream.on('data', chunk => {
```
dcrousso, line 14: does `ReadableStream` actually dispatch any events?

why not instead call `read()` in a loop (until the result is `null`)?

src/core/dwn-error.ts
```diff
@@ -0,0 +1,16 @@
...
+    super(`${code}: ${message}`);
```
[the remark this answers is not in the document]
thehenrytsai: Good call! Done.

src/interfaces/permissions/handlers/permissions-request.ts
```diff
@@ -5,12 +5,8 @@ import { canonicalAuth } from '../../../core/auth.js';
...
 import { PermissionsRequest } from '../messages/permissions-request.js';
 
-export const handlePermissionsRequest: MethodHandler = async (
-  tenant,
-  message,
-  messageStore,
-  didResolver
-): Promise<MessageReply> => {
+export const handlePermissionsRequest: MethodHandler = async (input): Promise<MessageReply> => {
+  const { tenant, message, messageStore, didResolver } = input;
```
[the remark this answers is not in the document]
thehenrytsai: Because not all props are used in this handler. So letting the handler choose what it cares about through destructuring.

src/interfaces/records/messages/records-write.ts
```diff
@@ -78,15 +79,19 @@ export class RecordsWrite extends Message {
   public static async create(options: RecordsWriteOptions): Promise<RecordsWrite> {
     const currentTime = getCurrentTimeInHighPrecision();
 
-    const dataCid = await computeDagPbCid(options.data);
+    if (options.data !== undefined && options.dataCid !== undefined) {
+      throw new Error('only one parameter between `data` and `dataCid` is allowed');
```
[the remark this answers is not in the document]
thehenrytsai: Keen eye!! This was one of the few lines that was pending code coverage. I've since added a test and made it mutually exclusive as intended.

src/interfaces/records/messages/records-write.ts
thehenrytsai: I think it's fine.

src/interfaces/records/messages/records-write.ts
thehenrytsai: I think it's fine.

src/interfaces/permissions/handlers/permissions-request.ts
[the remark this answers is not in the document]
dcrousso: im not suggesting that you destructure _all_ of the properties of `input`.  im suggesting that you do the destructure instead of even having an `input` parameter
```
export const handlePermissionsRequest: MethodHandler = async ({ tenant, message, messageStore, didResolver }): Promise<MessageReply> => {
```
that would make it _much_ easier for someone reading the code to immediately understand what (sub)parameters this function cares about instead of having to actually read any of the body (which is especially important for those that use code folding)

src/utils/data-stream.ts
thehenrytsai: It does.

You mean don't register any callbacks on any events and straight up `read()` in a `while` loop? Wouldn't it make this method blocking (not async) and thus not ideal?

Not too opinionated as this is currently purely surfaced in anticipation for SDK users to use as convenience, DWN itself doesn't use this, so if you want me to change it I will, just let me know!

dcrousso: i thought `read()` returned a `Promise`?

src/interfaces/permissions/handlers/permissions-request.ts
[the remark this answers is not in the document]
thehenrytsai: Done!

src/utils/data-stream.ts
thehenrytsai: I guess this is one (of many?) differences between browser-land and server-land when it comes to readable streams?! The abstraction that `readable-stream` provides adopts the node.js pattern, to which, wouldn't have worked with just a while loop based on their documentation anyway, still need to handle the `readable` event with the loop approach.

It's kinda a lose-lose situation in the sense that the web dev hates the SDK 'coz they can't use the web readable stream API directly, and I hate my job because I can only half-please everyone at best (thus please no one). But here we are, that's the requirement we got, at least it's cross platform.... I hope.

src/core/dwn-error.ts
mistermoe: @shamilovtim babel outright erases `public`

image

csuwildcat:
And on the JS feature front, I'll note private and static methods/variables
are now native to all JS environments, so you don't need to transpile for
them. Let's ensure we only target evergreen browsers and reasonably recent
Node envs so we don't have to care.

PR #221, opened 2023-02-13
clarifying readme examples

michaelneale:
@thehenrytsai @csuwildcat always bugged me this didn't have simple examples on the front page readme - is this ok?


README.md
```diff
@@ -21,14 +23,48 @@ npm install @tbd54566975/dwn-sdk-js
...
+export const dwn = await Dwn.create({});
+
+....
+const query = await RecordsWrite.create({
+  data                        : randomBytes(randomInt(50, 500)),
+  dataFormat                  : 'application/json',
+  published                   : true,
+  protocol                    : 'yeeter',
+  schema                      : 'yeeter/post',
+  authorizationSignatureInput : signatureMaterial
```
mistermoe, line 45: might be nice to include easiest way to generate did + `signatureMaterial` since both are basically necessary prerequisites to using dwns

mistermoe: thoughts? unrelated to this PR specifically, but it might be helpful to introduce utilities to this sdk so that people have everything they need to use DWNs.

README.md
```diff
@@ -21,14 +23,48 @@ npm install @tbd54566975/dwn-sdk-js
...
+});
+
+let result = await dwn.processMessage(didState.did, query.toJSON());
```
thehenrytsai, line 48: Method signature changed to take on data stream. Will pull down the branch and update the PR.

thehenrytsai (dismissed):
🎊

README.md
michaelneale: yeah I think it is lacking without it, so perhaps the readme should just point to example apps that use it instead?

thehenrytsai: @michaelneale, I think the intent is useful, let me do a refactor then update the README to meet @mistermoe's suggestion.

thehenrytsai: Also, let me know if you think `SignatureMaterial` makes more sense than `SignatureInput`, I can do an opportunistic rename. As a convenience reminder:

```typescript
export type SignatureInput = {
  protectedHeader: JwsHeaderParameters
  privateJwk: PrivateJwk
};
```

thehenrytsai: Done.

thehenrytsai (approved):
🎊 🎊

Issue #225, opened 2023-02-16
Define a "tenant gate" interface


thehenrytsai:
Currently messages are accepted regardless of what DID a message is targeted for.

Define a "tenant gate" interface so that the operator of the DWN can block messages routed to unknown tenant.

thehenrytsai:
Done

Issue #224, opened 2023-02-15
Don't require `{}` when initializing the DWN with defaults


thehenrytsai:
`Dwn.create({})` -> `Dwn.create()`

thehenrytsai:
Done

PR #226, opened 2023-02-16
#225 - defined a "tenant gate" interface and an allow-all default implementation

thehenrytsai:
Also opportunistically worked on #224, to not require empty object `{ }` for default DWN instantiation using `Dwn.create()`.


mistermoe:
Thank you

src/dwn.ts
```diff
@@ -1,9 +1,11 @@
 import type { BaseMessage } from './core/types.js';
 import type { DidMethodResolver } from './did/did-resolver.js';
 import type { MessageStore } from './store/message-store.js';
+import type { MethodHandler } from './interfaces/types.js';
 import type { Readable } from 'readable-stream';
-import type { Interface, MethodHandler } from './interfaces/types.js';
+import type { TenantGate } from './core/tenant-gate.js';
 
+import { AllowAllTenantGate } from './core/tenant-gate.js';
```
andresuribe87, line 8: Curious why this doesn't have `import type` ?

andresuribe87 (approved):

src/dwn.ts
thehenrytsai: @andresuribe87, because it is not a "type" declaration (e.g. such as an interface). It is an actual implementation (class). If no member of `AllowAllTenantGate` was ever accessed in this file, then I could add the `type` keyword as a best practice to help treeshaking by JS bundlers.

PR #229, opened 2023-02-16
UTC ISO-8601  timestamps

mistermoe:
change format of `dateCreated` and `dateUpdated` to UTC ISO-8601  timestamp

### DIFF
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -10,9 +10,9 @@ export function sleep(durationInMillisecond): Promise<void> {
 }
 
 /**
- * returns system time as a plain ISO date string with microsecond precision
+ * returns an UTC ISO-8601 timestamp with microsecond precision
  * using @js-temporal/polyfill
  */
 export function getCurrentTimeInHighPrecision(): string {
-  return Temporal.Now.plainDateTimeISO().toString({ smallestUnit: 'microseconds' });
+  return Temporal.Now.instant().toString({ smallestUnit: 'microseconds'  });
 }
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -14,5 +14,5 @@ export function sleep(durationInMillisecond): Promise<void> {
  * using @js-temporal/polyfill
  */
 export function getCurrentTimeInHighPrecision(): string {
-  return Temporal.Now.instant().toString({ smallestUnit: 'microseconds'  });
+  return Temporal.Now.instant().toString({ smallestUnit: 'microseconds' });
 }


dcrousso:
would something like `(new Date()).toISOString()` work instead?

src/utils/time.ts
```diff
@@ -10,9 +10,9 @@ export function sleep(durationInMillisecond): Promise<void> {
...
 /**
- * returns system time as a plain ISO date string with microsecond precision
+ * returns an UTC ISO-8601 timestamp with microsecond precision
  * using @js-temporal/polyfill
  */
 export function getCurrentTimeInHighPrecision(): string {
-  return Temporal.Now.plainDateTimeISO().toString({ smallestUnit: 'microseconds' });
-}
\ No newline at end of file
+  return Temporal.Now.instant().toString({ smallestUnit: 'microseconds'  });
```
dcrousso, line 17: would something like `(new Date()).toISOString()` work instead?

seems a bit unnecessary to load a polyfill for `Temporal` just to do something this basic 😕

mistermoe: agreed. @thehenrytsai might have context here. was the reason behind using `Temporal` to get microsecond precision?

dcrousso: also why not just use a timestamp instead of a string?  that's immediately more sortable and is significantly simpler to parse

mistermoe: deferring to @thehenrytsai here for context

mistermoe:
@thehenrytsai i have no idea why the CI linter check is saying i have extra spaces between curly braces. running `npm run lint` locally produces 0 errors

src/utils/time.ts
thehenrytsai: @mistermoe  and @dcrousso it is indeed the microsecond precision requirement.

thehenrytsai (approved):

src/utils/time.ts
dcrousso: ok that explains `Temporal`, but why not store the timestamp instead of a string?

thehenrytsai: @dcrousso, what timestamp format are you suggesting?

dcrousso: like a raw number

mistermoe: how important is this timestamp w.r.t ordering @thehenrytsai ? is it relied upon in other parts of the SDK?

dcrousso: the format shouldn't really matter for ordering (i.e. whether it's a string or a number is kinda irrelevant since theyre both just different ways of expressing the same underlying value)

also wouldn't a number be easier to compare?

thehenrytsai: @dcrousso, what raw number though? It needs to represent a timestamp in with microsecond precision.
@mistermoe, it is how message are compared for a given so it's important. Though I personally don't care about precision, milli, micro, nano, whatever the DWN spec wants.

dcrousso: maybe do it like JS does with `Date.now()` (which has millisecond precision) except represent the nanoseconds as a decimal (e.g. `1234567890123.456` would be the same as `"2009-02-13T23:31:30.123456Z"`)?  or maybe we just have it be a nanosecond precision value (e.g. `1234567890123456`)?

so long as it's clear in the documentation i think we can make this whatever format we want

thehenrytsai: @dcrousso, my 1st preference is to use a well-known standard/convention and over coming up with our own. ISO 8601 seems reasonable to me still.

thehenrytsai: Agreed to file issue to track: #230.

PR #232, opened 2023-02-17
Support all common variants of importing/requiring

mistermoe:
newer version of node use the `exports` property in `package.json` instead of `main` to figure out which file to import/require. Currently, the `package.json` in this repo only exports an `esm` distribution. This causes using the `cjs` distribution in newer versions of node to bork. 

This PR changes `exports` to support `cjs` and `esm`. Also added the `module` property because some bundlers look for it

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,11 @@
     "node": ">= 16"
   },
   "main": "./dist/cjs/src/index.js",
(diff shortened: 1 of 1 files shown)


thehenrytsai (approved):
🙏

Issue #244, opened 2023-03-03
Refactor `MessageStore`


mistermoe:
change `MessageStore` [`put`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store.ts#L19), [`get`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store.ts#L25), [`query`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store.ts#L34-L37), [`delete`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/message-store.ts#L42) method signatures to accept a `ctx: Context` as last argument.

```typescript
export interface Context {
  tenant: string
  signal?: AbortSignal
}
```

> 💡 This is a tee-up to address a separate call-out below. technically don't _need_ to provide `AbortSignal`. does seem nice to provide the caller with the ability to cancel an async operation.

---

replace usage of [`search-index`]() lib with handrolled search index. 

> 💡 `search-index` lib is the cause of all sorts of bundling woes particularly in react native. also tanks performance.

Hand-rolled:

```typescript
import type { AbstractBatchDelOperation, AbstractBatchOperation, AbstractLevel } from 'abstract-level';

import flat from 'flat';
import { Level } from 'level';

const { flatten } = flat;

export type Record = {
  _id: string,
  [prop: string]: unknown
};

export type RangeFilter = {
  gt?: unknown,
  gte?: unknown,
  lt?: unknown,
  lte?: unknown
};

export type TermFilter = string | number | boolean;

export type Filter = {
  [prop: string]: TermFilter | TermFilter[] | RangeFilter
};

export class IndexLevel {
  level: AbstractLevel<string, string>;

  constructor(level?) {
    this.level = level ?? new Level('__INDEX__');
  }

  async delete(_id: string): Promise<void> {
    const ops: AbstractBatchDelOperation<typeof this.level, string>[] = [];

    let keyPrefixes: string | string[] = await this.level.get(`__${_id}__meta`);
    if (!keyPrefixes) {
      return;
    }

    keyPrefixes = keyPrefixes.split('@');

    for (const keyPrefix of keyPrefixes) {
      const op = { type: 'del' as const, key: `${keyPrefix}~${_id}` };
      ops.push(op);
    }

    ops.push({ type: 'del' as const, key: `__${_id}__meta` });

    await this.level.batch(ops);
  }

[...]

PR #247, opened 2023-03-05
#244 - add a `Options` to each method of `MessageStore`

dcrousso:
currently it only contains a `signal: AbortSignal` that can be used to stop the activity of the called method

it's worth noting that this will only go so far, as `search-index`/`level`/etc. do not support anything like this, so ultimately it will only prevent future operations from happening

for example, given the following
```js
abortOr(signal, other())
```
even if the `signal` is already aborted, `other()` has already been called and therefore will not be stopped (unless `other` is changed to accept the `signal` as an argument)

### DIFF
--- a/src/store/blockstore-level.ts
+++ b/src/store/blockstore-level.ts
@@ -1,6 +1,7 @@
 import type { AwaitIterable, Batch, KeyQuery, Pair, Query } from 'interface-store';
 import type { Blockstore, Options } from 'interface-blockstore';
 
+import { abortOr } from '../utils/abort.js';
 import { CID } from 'multiformats';
 import { Level } from 'level';
 import { sleep } from '../utils/time.js';
@@ -54,13 +55,17 @@ export class BlockstoreLevel implements Blockstore {
     return this.db.close();
   }
 
-  put(key: CID, val: Uint8Array, _ctx?: Options): Promise<void> {
-    return this.db.put(key.toString(), val);
+  put(key: CID, val: Uint8Array, options?: Options): Promise<void> {
+    options?.signal?.throwIfAborted();
+
+    return abortOr(options?.signal, this.db.put(key.toString(), val));
   }
 
-  async get(key: CID, _options?: Options): Promise<Uint8Array> {
+  async get(key: CID, options?: Options): Promise<Uint8Array> {
+    options?.signal?.throwIfAborted();
+
     try {
-      const val = await this.db.get(key.toString());
+      const val = await abortOr(options?.signal, this.db.get(key.toString()));
       return val;
     } catch (e) {
       // level throws an error if the key is not present. Return undefined in this case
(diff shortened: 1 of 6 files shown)


src/store/blockstore-level.ts
```diff
@@ -54,13 +55,17 @@ export class BlockstoreLevel implements Blockstore {
     return this.db.close();
   }
 
-  put(key: CID, val: Uint8Array, _ctx?: Options): Promise<void> {
-    return this.db.put(key.toString(), val);
+  put(key: CID, val: Uint8Array, options?: Options): Promise<void> {
+    options?.signal?.throwIfAborted();
```
thehenrytsai, line 59: For my learning: I see this pattern throughout, I could guess, but what would be an example or scenario of why an abort signal would already be aborted upon the start of a method executing?

src/utils/abort.ts
```diff
@@ -0,0 +1,23 @@
...
+      return;
+    }
+
+    signal.addEventListener('abort', () => {
+      reject(signal.reason);
+    });
+  });
+}
+
+export async function abortOr<Type>(signal: AbortSignal, promise: Promise<Type>): Promise<Type> {
```
thehenrytsai, line 14: Add basic documentation above the function.

src/utils/abort.ts
```diff
@@ -0,0 +1,23 @@
+function promisifySignal<Type>(signal: AbortSignal): Promise<Type> {
+  return new Promise((resolve, reject) => {
+    if (signal.aborted) {
+      reject(signal.reason);
+      return;
+    }
```
thehenrytsai, line 6: This portion untested, please add a test for it.

thehenrytsai, line 6: This portion untested, please add a test for it.

src/store/blockstore-level.ts
dcrousso: perhaps something like this?
```js
let controller = new AbortController;

try {
    await doThisThing({ signal: controller.signal });
} catch {
    controller.abort();
}

doThatThing({ signal: controller.signal });
```
i.e. if the same `AbortSignal` is used by multiple things


also it's worth mentioning that it's possible to create an already aborted `AbortSignal` via `AbortSignal.abort([reason])` (and im pretty sure the other uses of `AbortSignal` (e.g. `fetch`) do it this way too, so i was trying to be consistent)

thehenrytsai: What would be the reason for continuing with `doThatThing()` if an abort signal is already detected in the `catch` in this case? Is the `throwIfAborted()` check in every function more of a defensive programming pattern in case the caller needlessly call `doSomething({ signal: controller.signal });`?

thehenrytsai (approved):
🎊

src/store/blockstore-level.ts
dcrousso: you're assuming that it's not possible to recover from a rejection of `doThisThing`.  `doThatThing` could be totally independent and/or the containing function could allow for that failure and still continue (and it might be easier to use the `AbortSignal` instead of having a `throw` since it centralizes where the exception can come from, instead of having more than one spot)

i think it'd be more realistic if i rewrote it like this
```js
async function doFileStuff(path, callback, { signal } = { }) {
    let file;

    try {
        file = await fileOpen(path, { signal });
        await callback(file);
     } catch { }

     if (file)
         await fileClose(file, { signal });
 }
```
we always want to call `fileClose` (assuming that `fileOpen` succeeded) unless the entire operation was aborted (e.g. if `controller.abort()` was called inside the `callback` due to some error case)

... tho thinking a bit more perhaps filesystem stuff is also not the best example, but hopefully this better demonstrates an example scenario 😅

but yeah generally speaking it's definitely more of a defensive approach 🙂

dcrousso: sorry i wasnt entirely clear.  i meant that `doThatThing` was independent of `doThisThing`, not the `signal`

to use the `doFileStuff` example, what if it was called like so
```
let controller = new AbortController;
doFileStuff("foo.txt", (file) => {
    if (file.empty())
        controller.abort();
}, { signal: controller.signal })
```
then we'd explicitly want a `options?.signal?.throwIfAborted();` as the first line inside `fileClose` as otherwise we'd potentially do _some_ work, which kinda defeats the purpose of `AbortSignal`

Issue #218, opened 2023-02-09
Use tenant-scoped IDs in `MessageStore`


thehenrytsai:
With the removal of `target` from the `descriptor` in #201, messages are no longer natively DID/tenant scoped. This means that the exactly same message maybe stored across multiple tenants in the same DWN.

Currently, if a record delete occurs in one of the tenant, it is deleted in all tenants.

thehenrytsai:
This is now done.

PR #251, opened 2023-03-08
#244 - add a `tenant` parameter to each `MessageStore` method

dcrousso:
we should not allow anyone to manipulate the entire `MessageStore`, especially things that they do not have control over

requiring a `tenant` for each operation ensures that `MessageStore` will only operate on that "slice" of the data

for `MessageStoreLevel`, this is implemented via `AbstractLevel.prototype.sublevel`


src/store/message-store-level.ts
```diff
@@ -87,46 +90,52 @@ export class MessageStoreLevel implements MessageStore {
     const messages: BaseMessage[] = [];
 
     // parse criteria into a query that is compatible with the indexing DB (search-index) we're using
-    const exactTerms = MessageStoreLevel.buildExactQueryTerms(exactCriteria);
+    const exactTerms = MessageStoreLevel.buildExactQueryTerms({ ...exactCriteria, tenant });
```
thehenrytsai, line 93: With this change, I think we can/should remove the explicit indexing of tenant in everywhere `put()` is invoked:

* `.../handlers/protocols-configure.ts` - line 63
* `.../handlers/records-delete.ts` - line 94
* `.../handlers/records-write.ts` - line 125
* `.../handlers/permissions-request.ts` - line 33

I'd attempt to make the changes, but also don't want to screw up your workflow.

thehenrytsai (changes requested):
A freaking awesome PR!! Definitely love the nice clean separation of tenant data, as long as sublevel is readily available feature for all level implementations I guess.

Also love how isolated/target the PR is and not "polluted" with other S.

The change paves the way for the work outlined in #218 overall, I do wonder:

1. if same technique can be applied at the `record` level in data store
1. perf implication if any

But that's out of scope for this PR.

A couple of comments and the PR is good to go.

Also I noticed the workflow (rebase whatnot?) that you are using is making the PR reviewers' life more difficult: ie. the very useful feature of just looking at diffs between iterations of code after each review is lost every time you push, which means the reviewer is forced to look over the entire PR again to look for the changes corresponding to the comments being addressed and new changes introduced. Not a big deal for small PRs but definitely a toll for larger PRs.

dcrousso:
> Also I noticed the workflow (rebase whatnot?) that you are using is making the PR reviewers' life more difficult: ie. the very useful feature of just looking at diffs between iterations of code after each review is lost every time you push, which means the reviewer is forced to look over the entire PR again to look for the changes corresponding to the comments being addressed and new changes introduced. Not a big deal for small PRs but definitely a toll for larger PRs.

FYI there should be a "Compare" button next to the commit each time I push that'll show the changes in the push, even if it was a force push
Screenshot 2023-03-08 at 14 14 50

Issue #252, opened 2023-03-08
Introduce `dataSize` property in `RecordsWrite`


thehenrytsai:
Related to #198.

This further allows features such as:

1. Returning data as `encodedData` only if the data size falls below a certain threshold
1. Protocol rules that limits data size of a message

thehenrytsai:
Done.

PR #253, opened 2023-03-08
#252 - introduced dataSize property in RecordsWrite

thehenrytsai:


mistermoe (commented):
instinctively, i don't think providing `dataSize` should be the responsibility of the message author. we're already expecting author to provide data CID and having the author sign over that, so there's not much gained from an integrity perspective.   

i see the benefit of `dataSize` but it feels like something that can be calculated, stored, and returned without putting the burden on a message author without losing integrity

frankhinek:
> instinctively, i don't think providing `dataSize` should be the responsibility of the message author. we're already expecting author to provide data CID and having the author sign over that, so there's not much gained from an integrity perspective.
> 
> i see the benefit of `dataSize` but it feels like something that can be calculated, stored, and returned without putting the burden on a message author without losing integrity

I tend to agree with @mistermoe's comments that `dataSize` should be _computed_ by the DWN SDK, particularly since the dataSize is being checked anyway to ensure the value specified in the message matches the actual size of the data.

@thehenrytsai What do you view as the primary benefit of requiring `dataSize` be specified in the event a `dataCid` is defined in the `RecordsWrite.create()`?

thehenrytsai:
@mistermoe, @frankhinek, and @csuwildcat,

To not have the PR sit in limbo, I could abandon/morph the PR to introduce a `ancillaryData` container property (other names welcome) accompanying each message returned by the `RecordsQuery`, where it will contain the `dataSize`.

I believe this would tick all boxes from everyone in the scope of this work item. What do you think?

frankhinek:
> The main problems with not including `dataSize` in the message (which was discusses in the office hours) were:
> 
> 1. We need to further discuss the mechanism/schema in which computed property is returned to the client, as this would be the first precedence of such need. Do you have recommendation on what that should look like?
> 2. We have intentionally sidestepped precisely pt. 1 with `recordId` and `contextId` by requiring them in message, if were to take pt. 1 for `dataSize` then it begs the question should we are not omitting properties such as `recordId` and `contextId` also?
> 3. Assuming data stream is the original raw stream for `RecordsRead` for example, the server could claim a large file size and trick client into wasting bandwidth, but maybe we could say this is not a case we worry about.

Given the points raised and the minimal change in developer experience in the simple case where data is passed when calling `RecordsWrite.create()`, it seems like moving forward with adding `dataSize` to the descriptor is a reasonable option to enable the DWN to return only those messages that are of a "small" size (e.g., <100KB).

And in comparison to adding an `ancillaryData` container to RecordsQuery results seems like a bigger change that wouldn't be worth introducing simply for this scenario.

PR #254, opened 2023-03-09
#249- only import `'level'` when it's actually used

dcrousso:
i.e. if a developer provides their own `createLevelDatabase` then we will use that instead of `'level'`

in order to support this for both node and browser, we need to switch to using dynamic `import` (instead of `require`, as that's not defined for node when `"type": "module"`)

most of this commit is changing sync logic to `async` for that purpose (though it also allows for more flexibility when providing a custom `createLevelDatabase`)

### DIFF
--- a/src/store/blockstore-level.ts
+++ b/src/store/blockstore-level.ts
@@ -22,16 +22,16 @@ export class BlockstoreLevel implements Blockstore {
    * the {@link https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase IDBDatabase}
    * to be opened.
    */
-  constructor(location: string, config: BlockstoreLevelConfig = {}) {
+  constructor(config: BlockstoreLevelConfig) {
     this.config = {
       createLevelDatabase,
       ...config
     };
-
-    this.db = this.config.createLevelDatabase<string, Uint8Array>(location, { keyEncoding: 'utf8', valueEncoding: 'binary' });
   }
 
   async open(): Promise<void> {
+    await this.createLevelDatabase();
+
     while (this.db.status === 'opening' || this.db.status === 'closing') {
       await sleep(200);
     }
@@ -52,6 +52,10 @@ export class BlockstoreLevel implements Blockstore {
    * releases all file handles and locks held by the underlying db.
    */
   async close(): Promise<void> {
+    if (!this.db) {
+      return;
+    }
+
     while (this.db.status === 'opening' || this.db.status === 'closing') {
       await sleep(200);
     }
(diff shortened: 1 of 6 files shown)


shamilovtim (approved):

shamilovtim (approved):

src/store/blockstore-level.ts
```diff
@@ -52,6 +52,8 @@ export class BlockstoreLevel implements Blockstore {
    * releases all file handles and locks held by the underlying db.
    */
   async close(): Promise<void> {
+    await this.createLevelDatabase();
```
thehenrytsai, line 55: for my understanding: why is it not easier/better to just not do anything if `db` is not defined?

src/store/create-level.ts
```diff
@@ -1,11 +1,10 @@
...
 
-import { Level } from 'level';
-
 export type LevelDatabase<K, V> = AbstractLevel<string | Buffer | Uint8Array, K, V>;
 
 export type LevelDatabaseOptions<K, V> = AbstractDatabaseOptions<K, V>;
 
-export function createLevelDatabase<K, V>(location: string, options?: LevelDatabaseOptions<K, V>): LevelDatabase<K, V> {
+export async function createLevelDatabase<K, V>(location: string, options?: LevelDatabaseOptions<K, V>): Promise<LevelDatabase<K, V>> {
+  const { Level } = await import('level');
```
thehenrytsai, line 8: Can you add the summary you had on this PR as comment, feels like important context not trivial to reader of the code.

thehenrytsai (commented):
Thanks! Looks great over all, just a couple of comments.

src/store/create-level.ts
shamilovtim: should `level` be a peer dep to really cement this boundary and prevent it from slipping back in to the codebase?

src/store/blockstore-level.ts
dcrousso: oh lol true didnt think of that

src/store/create-level.ts
mistermoe: what's your understanding of peer dependency @shamilovtim ? i just realized i didn't really know _what_ exactly the distinction meant. after reading [this](https://nodejs.org/ko/blog/npm/peer-dependencies/#the-solution-peer-dependencies)

> Some way of saying, "I only work when plugged in to version 1.2.x of my host package, so if you install me, be sure that it's alongside a compatible host." We call this relationship a peer dependency.

i feel like making `level` a peer dependency is the opposite of what we want. but maybe i'm not understanding `peerDependencies` correctly

mistermoe:
@thehenrytsai can you publish an unstable version of this so @shamilovtim can consume?

shamilovtim:
> @thehenrytsai can you publish an unstable version of this so @shamilovtim can consume?

This one can probably just get merged. I need the unstable of the other PR and this rebased into that PR first

thehenrytsai:
@mistermoe and @shamilovtim, approved. Letting @dcrousso have the honor of merging. Feel free to merge if you can't wait.

Issue #245, opened 2023-03-03
Update IPFS related dependencies to latest version


thehenrytsai:
Latest state of this issue:

✅ "@ipld/dag-cbor": "9.0.0",
✅ "interface-blockstore": "4.0.1"
✅ "ipfs-unixfs": "11.0.0",
✅ "ipfs-unixfs-exporter": "13.1.0",
✅ "ipfs-unixfs-importer": "14.0.1",
✅ "multiformats": "11.0.2",

Some dependencies are not backwards compatible thus will break build/compilation so will require update to code.

More annoying is that it appears some of the latest IPFS packages aren't exporting types correctly/fully which may require changes to IPFS repos. For reference, following errors are thrown when attempting to compile:

```
node_modules/@ipld/dag-pb/dist/src/index.d.ts:22:29 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

22 export type PBLink = import('./interface').PBLink;
                               ~~~~~~~~~~~~~

node_modules/@ipld/dag-pb/dist/src/index.d.ts:23:29 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

23 export type PBNode = import('./interface').PBNode;
                               ~~~~~~~~~~~~~

node_modules/@ipld/dag-pb/dist/src/util.d.ts:15:61 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

15 export function createNode(data: Uint8Array, links?: import("./interface").PBLink[] | undefined): PBNode;
                                                               ~~~~~~~~~~~~~

node_modules/@ipld/dag-pb/dist/src/util.d.ts:23:29 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

23 export type PBLink = import('./interface').PBLink;
                               ~~~~~~~~~~~~~

node_modules/@ipld/dag-pb/dist/src/util.d.ts:24:29 - error TS2834: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Consider adding an extension to the import path.

[...]

PR #255, opened 2023-03-10
#245 - updated multiformats, @ipld/dag-cbor, and interface-blockstore dependencies

thehenrytsai:


src/utils/cid.ts
```diff
@@ -29,7 +29,7 @@ const codecs = {
  * @throws {Error} encoding fails
  * @throws {Error} if hasher is not supported
  */
-export async function computeCid(payload: any, codecCode = cbor.code, multihashCode = sha256.code): Promise<string> {
+export async function computeCid(payload: any, codecCode: number = cbor.code, multihashCode: number = sha256.code): Promise<string> {
```
diehuxx, line 32: Nit: this line is on the long side. do you mind linebreaking it somewhere?

src/store/message-store-level.ts
```diff
@@ -73,7 +73,7 @@ export class MessageStoreLevel implements MessageStore {
       return undefined;
     }
 
-    const decodedBlock = await abortOr<block.Block>(options?.signal, block.decode({ bytes, codec: cbor, hasher: sha256 }));
```
diehuxx, line 76: I think usages of `block` are all removed, but we're still importing. mind removing?

While on the subject, what do you think of[ adding a plugin](https://www.npmjs.com/package/eslint-plugin-unused-imports) to the linter to catch unused imports? I can file an issue and add a PR.

diehuxx (approved):
A couple non-blocking comments. Otherwise LGTM

src/utils/cid.ts
thehenrytsai: Done. Note that we do have a linter rule on max 150 chars.

diehuxx: ah didn't realize. i'm used to closer to 80-100 max chars

src/store/message-store-level.ts
thehenrytsai: I think it does get used, `block.decode()` in same line. Also we already have a linter rule `@typescript-eslint/no-unused-vars` includes this check, I get nagged all the time when imports are not used.

diehuxx (approved):

src/utils/cid.ts
thehenrytsai: 80 to 100?!?! That was the standard length limit back when I was in college in 2000 when the screen size was 15" and resolution was 1024x768..... we now have 32"+ monitors with crazy resolutions!! = ]

PR #257, opened 2023-03-11
#244 - use the same parameter types for `MessageStoreLevel.prototype.query` and `IndexLevel.prototype.query`

dcrousso:
this makes the API simpler to understand as `ExactCriterion` and `RangeCriterion` are expressed in the types rather than two separate parameters

it also entirely removes any chance of clashing if the same property is provided in both `exactCriteria` and `rangeCriteria`

### DIFF
--- a/src/interfaces/records/messages/records-query.ts
+++ b/src/interfaces/records/messages/records-query.ts
@@ -1,5 +1,6 @@
 import type { SignatureInput } from '../../../jose/jws/general/types.js';
-import type { ExactCriterion, RangeCriterion, RecordsQueryDescriptor, RecordsQueryFilter, RecordsQueryMessage } from '../types.js';
+import type { Filter, RangeFilter } from '../../../core/types.js';
+import type { RecordsQueryDescriptor, RecordsQueryFilter, RecordsQueryMessage } from '../types.js';
 
 import { getCurrentTimeInHighPrecision } from '../../../utils/time.js';
 import { Message } from '../../../core/message.js';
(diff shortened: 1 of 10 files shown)


dcrousso:
i was tempted to change the query JSON schema to use `gte` and `lte` as well, but i felt that was probably more of a larger (and possibly contentious) change so decided to hold off :)

src/interfaces/records/messages/records-query.ts
```diff
@@ -70,25 +71,28 @@ export class RecordsQuery extends Message {
...
-    const rangeCriteria = filter.dateCreated ? { dateCreated: filter.dateCreated } : { };
-    return rangeCriteria;
+  public static convertFilter(filter: RecordsQueryFilter): Filter {
+    const result: Filter = { };
+    for (const key in filter) {
+      switch (key) {
+      case 'dateCreated':
+        var rangeFilter: RangeFilter = {
+          gte : filter.dateCreated.from,
+          lte : filter.dateCreated.to
```
thehenrytsai, line 81: Can we do `lt` here now?

```suggestion
          lt : filter.dateCreated.to
```

thehenrytsai (approved):
🙏

src/interfaces/records/messages/records-query.ts
dcrousso: id prefer to do this in a followup just so that it doesn't look like an accidental typo to someone who doesn't read the commit message

PR #258, opened 2023-03-11
#250 and #257 - change `to` of query JSON schema to represent `lt` instead of `lte`

dcrousso:
this better matches many other APIs that operate with a [inclusive, exclusive) system (e.g. `String.prototype.slice`)


thehenrytsai (approved):
🙏 🙏 🙏

PR #259, opened 2023-03-13
create a concrete wrapper for `level`

dcrousso:
the behavior adjustments made inside `BlockstoreLevel` (e.g. `get` not throwing if the key is not found) should be used everywhere so that we have a consistent pattern for using `level`


package.json
```diff
@@ -86,6 +86,7 @@
     "@types/varint": "6.0.0",
     "@typescript-eslint/eslint-plugin": "5.37.0",
     "@typescript-eslint/parser": "5.37.0",
+    "array-from-async": "^3.0.0",
```
thehenrytsai, line 89: nit: sharing you past sentiment (I think), is it possible to avoid adding a new dependency for use in once place?

dcrousso: this is a prollyfill for the [array-from-async](https://github.com/tc39/proposal-array-from-async) TC39 proposal, so i wouldnt necessarily call it the same kind of dependency as the others

but yes that's a fair point i shouldnt be a hypocrite :P

dcrousso: oh tho actually this is a `devDependency`, so i personally dont have as much of a concern about it since its not part of shipping code

but ill still remove it :)

thehenrytsai (dismissed):
🥇 🎸 🚀

thehenrytsai (approved):

PR #260, opened 2023-03-13
#198 - added size check when returning encodedData

thehenrytsai:
* added size check when returning encodedData
* opportunistically added some documentation

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -49,8 +49,19 @@ export type EqualFilter = string | number | boolean;
 
 export type OneOfFilter = EqualFilter[];
 
-type GT = ({ gt: string } & { gte?: never }) | ({ gt?: never } & { gte: string });
-type LT = ({ lt: string } & { lte?: never }) | ({ lt?: never } & { lte: string });
+/**
+ * "greater than" or "greater than or equal to" range condition. `gt` and `gte` are mutually exclusive.
+ */
+
+/**
+ * "less than" or "less than or equal to" range condition. `lt`, `lte` are mutually exclusive.
+ */
+export type GT = ({ gt: string } & { gte?: never }) | ({ gt?: never } & { gte: string });
+export type LT = ({ lt: string } & { lte?: never }) | ({ lt?: never } & { lte: string });
+
+/**
+ * Ranger filter. 1 condition is required.
+ */
 export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
 
 export type Filter = {
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -49,14 +49,15 @@ export type EqualFilter = string | number | boolean;
 
 export type OneOfFilter = EqualFilter[];
 
+
 /**
  * "greater than" or "greater than or equal to" range condition. `gt` and `gte` are mutually exclusive.
  */
+export type GT = ({ gt: string } & { gte?: never }) | ({ gt?: never } & { gte: string });
 
 /**
  * "less than" or "less than or equal to" range condition. `lt`, `lte` are mutually exclusive.
  */
-export type GT = ({ gt: string } & { gte?: never }) | ({ gt?: never } & { gte: string });
 export type LT = ({ lt: string } & { lte?: never }) | ({ lt?: never } & { lte: string });
 
 /**
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -35,14 +35,14 @@ export type TimestampedMessage = BaseMessage & {
 };
 
 /**
- * Message that references `dataCid`.
+ * Message returned in a query result.
+ * NOTE: the message structure is a modified version of the message received, the most notable differences are:
+ * 1. does not contain `authorization`
+ * 2. may include encoded data
  */
-export type DataReferencingMessage = {
-  descriptor: {
-    dataCid: string;
-  };
-
-  encodedData: string;
+export type QueryResultEntry = {
+  descriptor: Descriptor;
+  encodedData?: string;
 };
 
 export type EqualFilter = string | number | boolean;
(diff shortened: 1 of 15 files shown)


src/store/storage-controller.ts
```diff
@@ -84,12 +85,11 @@ export class StorageController {
...
     const messages = await messageStore.query(tenant, filter);
 
+    // for every message, only include the data as `encodedData` if the data size is equal or smaller than the size threshold
     for (const message of messages) {
       const dataCid = message.descriptor.dataCid;
-      if (dataCid !== undefined) {
-        // TODO: #219 (https://github.com/TBD54566975/dwn-sdk-js/issues/219)
-        // temporary placeholder for keeping status-quo of returning data in `encodedData`
-        // once #219 is implemented, `encodedData` may or may not exist directly as part of the returned message here
+      const dataSize = message.descriptor.dataSize;
```
diehuxx, line 91: why do we assume `dataCid` might be undefined but we assert that `dataSize` is defined? Does `dataCid` being defined guarantee that `dataSize` is also defined? If so, where does that check happen?

src/core/types.ts
```diff
@@ -35,22 +35,34 @@ export type TimestampedMessage = BaseMessage & {
...
+ */
+export type GT = ({ gt: string } & { gte?: never }) | ({ gt?: never } & { gte: string });
+
+/**
+ * "less than" or "less than or equal to" range condition. `lt`, `lte` are mutually exclusive.
+ */
+export type LT = ({ lt: string } & { lte?: never }) | ({ lt?: never } & { lte: string });
+
+/**
+ * Ranger filter. 1 condition is required.
```
diehuxx, line 64: 🙏 much appreciate all the comments you're adding in this PR

src/store/storage-controller.ts
thehenrytsai: > Does `dataCid` being defined guarantee that `dataSize` is also defined?

Yes.

> If so, where does that check

https://github.com/TBD54566975/dwn-sdk-js/blob/main/json-schemas/records/records-write.json#L74

src/core/types.ts
```diff
@@ -35,22 +35,34 @@ export type TimestampedMessage = BaseMessage & {
...
+ * 1. does not contain `authorization`
+ * 2. may include encoded data
  */
-export type DataReferencingMessage = {
-  descriptor: {
-    dataCid: string;
-  };
-
-  encodedData: string;
+export type QueryResultEntry = {
```
diehuxx, line 43: +1 to the name change from DataReferencingMessage to QueryResultEntry.

Question: Is `entry` a specific term, defined in the DWN spec or somewhere? Looks like we are using it to mean something like "the part of a message response that contains the requested data blob, plus a descriptor."

While on the subject, what is a descriptor? The DWN spec has some hyperlinks to `https://identity.foundation/decentralized-web-node/spec/#message-descriptors`, but doesn't look like that section exists (anymore?).

tests/interfaces/records/handlers/records-query.spec.ts
```diff
@@ -98,6 +99,38 @@ describe('RecordsQueryHandler.handle()', () => {
...
+
+      const writeReply = await dwn.processMessage(alice.did, write.message, write.dataStream);
+      expect(writeReply.status.code).to.equal(202);
+
+      const messageData = await TestDataGenerator.generateRecordsQuery({ requester: alice, filter: { recordId: write.message.recordId } });
+      const reply = await dwn.processMessage(alice.did, messageData.message);
+
+      expect(reply.status.code).to.equal(200);
+      expect(reply.entries?.length).to.equal(1);
+      expect((reply.entries[0] as any).encodedData).to.equal(Encoder.bytesToBase64Url(data));
```
diehuxx, line 115: why `as any`? My typescript isn't super deep, but I would assume that it can pick up that `entries` is `QueryResultEntry[]`, which have property `encodedData`.

src/core/types.ts
thehenrytsai: Honestly, I didn't understanding exactly the range filter type, but went with 100% trust here.

src/core/types.ts
thehenrytsai: > Is entry a specific term

Yes, it's defined in spec. such as https://identity.foundation/decentralized-web-node/spec/#example-response-object

> what is a descriptor

Probably easiest to think of it as a container property that "describes" the metadata of a message. The practical reason for having such a container is so that we can perform operations on the group of properties as a whole, such as computing the `recordId`.

> doesn't look like that section exists (anymore?).

Feel free to file spec issues against @csuwildcat

diehuxx (dismissed):

tests/interfaces/records/handlers/records-query.spec.ts
thehenrytsai: Thanks for bringing this up! This was a copy and past from existing tests before I decided to refactor the code. It was needed before the the intro of `QueryResultEntry`, but no longer needed now.

Went ahead and removed all the `as any` in tests.

diehuxx: Thanks for the extra clean up!
Not to punish a good deed, but I see a few more in records-query.spec.ts that could be cleaned up. Found them in VSCode CMD-F with regex `entries\[.\] as any\).encodedData`. Do you mind cleaning up those too?

thehenrytsai: I am not seeing any more, can you point me to an example?

diehuxx: Ope, sorry I think my VSCode was showing me the wrong commit. Apologies, still getting used to this editor.

diehuxx (approved):

Issue #84, opened 2022-08-11
Implement data deletion in message store


thehenrytsai:
Currently the message store does not have the ability to delete data referenced by the CollectionsWrite messages. The implementation needs to take into account that:

1. Records within the same tenant can reference the same block.
2. Records across multiple tenants can reference the same block.

thehenrytsai:
@dcrousso, I just did a sanity test of writing a 100MB data blob, then deleting it again. And the datastore folder still has ~100MB after deletion.

Can you sanity verify and investigate and fix if true?

thehenrytsai:
Hey @dcrousso I did the same sanity test with a 100MB data blob before closing this issue, the level folder still appears to be 100MB in size... can you sanity verify on your end to make sure I am not dreaming?

dcrousso:
im pretty sure that's how `leveldb` works, as a sort of performance optimization to prevent having to read the entire DB, figure out what to get rid of, and then (re)write it all back to disk (i.e. a kinda sorta somewhat operation similar to defragmentation)

if you do a test where you

1. `put` 100MB
2. `delete` 100MB
3. call `level.prototype.open` (repeat)

you should see that most of the time you call `level.prototype.open` the size of the folder on disk will go down (im pretty sure this is because one of the places where (partial) compaction happens is during `open`)

if it's a _hard_ requirement that we delete the data on disk _immediately_ then we might be able to leverage `ClassicLevel.prototype.compactRange` (though we'd have to figure out what to provide for `start` and `end`) which should trigger a compaction (of the given range), but IMO since the actual `leveldb` API doesn't expose this then i feel it's probably OK (i.e. you'd need access to the machine to get the data, or to have found some bug in `leveldb`) and it might utterly _tank_ the overall performance of the `Dwn`

lmk what you want me to do :)

thehenrytsai:
This is now done.

Issue #230, opened 2023-02-17
Revisit `dateCreated`, `dateModified`, and `datePublished` format


thehenrytsai:
Currently it is expected to be ISO 8601 UTC format, revisit to see if we should change to a custom format of nanoseconds since unix epoch or some other format.

dcrousso:
so i tried taking a stab at this by changing the JS representation to a `Temporal.Instant` and having the JSON be a stringified `BigInt` (e.g. `"1234567890123456789"` vs `"2009-02-13T23:31:30.123456789Z"`), but quickly ran into some issues.

the current architecture of having the `message` be both the argument to construct a `RecordsWrite`/`RecordsQuery`/`RecordsDelete` means that we're kinda constrained to only having JSON types in our JS API, which IMO feels pretty constrained and isnt allowing us to take advantage of all the powerful JS APIs at our fingertips (e.g. we'd have to convert the ISO string to a `Temporal.Instant` each time we wanted to do a `Tempora.Instant.compare` instead of being able to have the property be a `Temporal.Instant` from the getgo).  id suggest we consider an architecture that actually pulls apart (and repacks) the JSON so that we can be much more OOP about all of this, instead of having things basically just be wrappers around JSON.

putting all that aside tho, here are some of the benefits and drawbacks of using other formats

-----

#### ISO-8601
##### Pros
- standardized format for creation and parsing
- very human readable
- extremely well known and supported (at least with only milliseconds precision)
##### Cons
- more verbose with lots of repeated characters (e.g. `-`, `T`, `:`, `Z`, etc.)
- most things I can find only support up to milliseconds precision \(even the [official standards page](https://www.iso.org/iso-8601-date-and-time-format.html) only lists milliseconds\)
- stringifying a `Date` to JSON automatically uses this, but parsing from JSON will not, so would still need a transformer if we wanted to manipulate a `Date` instead of a raw string
- manipulation and comparison are not very easy and likely require transforming to a different form to be done efficiently (or at all)

#### `BigInt`
##### Pros
- fantastic JS API for creation, manipulation, comparison, stringification, etc.
- easy interop with `Temporal.Instant`
- very standardized and fully builtin
- timestamps are a super understood concept (at least with only seconds/milliseconds precision) and have lots of benefits over date strings
##### Cons
[...]

PR #263, opened 2023-03-16
#84 - implement a reference counting system for data management to avoid duplication

dcrousso:
this is done by reworking how data is stored inside `DataStoreLevel`

it has the following structure (`+` represents a sublevel and `->` represents a key->value pair):
```
'data' + <dataCid> -> <data>
'host' + <dataCid> + <tenant> + <messageCid> -> PLACEHOLDER_VALUE
```
this allows for the `<data>` to be shared for everything that uses the same `<dataCid>` while also making sure that the `<data>` can only be deleted if there are no `<messageCid>` for any `<tenant>` still using it

note we are currently doing this using `sublevel`, but that has the restriction of only allowing keys to contain characters after `34` and before `127`, meaning that if `<dataCid>`/`<tenant>`/`<messageCid>` ever allow for characters outside that range we will need to adjust

also switch from `logicalId`/`recordId` to `messageCid` as the former can be reused by multiple things (e.g. two different `RecordsWrite` having the same `recordId`) whereas the latter is derived from the entire message

### DIFF
--- a/src/interfaces/records/handlers/records-delete.ts
+++ b/src/interfaces/records/handlers/records-delete.ts
@@ -75,7 +75,7 @@ export class RecordsDeleteHandler implements MethodHandler {
     }
 
     // delete all existing messages that are not newest, except for the initial write
-    await deleteAllOlderMessagesButKeepInitialWrite(tenant, existingMessages, newestMessage, this.messageStore);
+    await deleteAllOlderMessagesButKeepInitialWrite(tenant, existingMessages, newestMessage, this.messageStore, this.dataStore);
 
     return messageReply;
   };
--- a/src/interfaces/records/handlers/records-write.ts
+++ b/src/interfaces/records/handlers/records-write.ts
@@ -106,7 +106,7 @@ export class RecordsWriteHandler implements MethodHandler {
     }
 
     // delete all existing messages that are not newest, except for the initial write
-    await deleteAllOlderMessagesButKeepInitialWrite(tenant, existingMessages, newestMessage, this.messageStore);
+    await deleteAllOlderMessagesButKeepInitialWrite(tenant, existingMessages, newestMessage, this.messageStore, this.dataStore);
 
     return messageReply;
   };
--- a/src/interfaces/records/records-interface.ts
+++ b/src/interfaces/records/records-interface.ts
@@ -1,10 +1,12 @@
-import { MessageStore } from '../../store/message-store.js';
-import { RecordsWrite } from './messages/records-write.js';
+import type { DataStore } from '../../store/data-store.js';
+import type { MessageStore } from '../../store/message-store.js';
 import type { RecordsWriteMessage } from '../../interfaces/records/types.js';
 import type { TimestampedMessage } from '../../core/types.js';
 
 import { constructRecordsWriteIndexes } from './handlers/records-write.js';
 import { Message } from '../../core/message.js';
+import { RecordsWrite } from './messages/records-write.js';
+import { StorageController } from '../../store/storage-controller.js';
 
 /**
  * Deletes all messages in `existingMessages` that are older than the `comparedToMessage` in the given tenant,
@@ -13,19 +15,19 @@ import { Message } from '../../core/message.js';
 export async function deleteAllOlderMessagesButKeepInitialWrite(
   tenant: string,
   existingMessages: TimestampedMessage[],
-  comparedToMessage,
-  messageStore: MessageStore
+  comparedToMessage: TimestampedMessage,
+  messageStore: MessageStore,
+  dataStore: DataStore
 ): Promise<void> {
   // NOTE: under normal operation, there should only be at most two existing records per `recordId` (initial + a potential subsequent write/delete),
   // but the DWN may crash before `delete()` is called below, so we use a loop as a tactic to clean up lingering data as needed
   for (const message of existingMessages) {
-    const messageIsOld = await RecordsWrite.isOlder(message, comparedToMessage);
+    const messageIsOld = await Message.isOlder(message, comparedToMessage);
     if (messageIsOld) {
       // the easiest implementation here is delete each old messages
       // and re-create it with the right index (isLatestBaseState = 'false') if the message is the initial write,
       // but there is room for better/more efficient implementation here
-      const cid = await Message.getCid(message);
-      await messageStore.delete(tenant, cid);
+      await StorageController.delete(messageStore, dataStore, tenant, message);
 
       // if the existing message is the initial write
       // we actually need to keep it BUT, need to ensure the message is no longer marked as the latest state
--- a/src/store/blockstore-level.ts
+++ b/src/store/blockstore-level.ts
@@ -40,20 +40,24 @@ export class BlockstoreLevel implements Blockstore {
     return new BlockstoreLevel({ ...this.config, location: '' }, db);
   }
 
-  async put(key: CID, val: Uint8Array, options?: Options): Promise<void> {
-    return this.db.put(key.toString(), val, options);
+  async put(key: CID | string, val: Uint8Array, options?: Options): Promise<void> {
+    return this.db.put(String(key), val, options);
   }
 
-  async get(key: CID, options?: Options): Promise<Uint8Array> {
-    return this.db.get(key.toString(), options);
+  async get(key: CID | string, options?: Options): Promise<Uint8Array> {
+    return this.db.get(String(key), options);
   }
 
-  async has(key: CID, options?: Options): Promise<boolean> {
-    return !! await this.get(key, options);
+  async has(key: CID | string, options?: Options): Promise<boolean> {
+    return this.db.has(String(key), options);
   }
 
-  async delete(key: CID, options?: Options): Promise<void> {
-    return this.db.delete(key.toString(), options);
+  async delete(key: CID | string, options?: Options): Promise<void> {
+    return this.db.delete(String(key), options);
+  }
+
+  async isEmpty(options?: Options): Promise<boolean> {
+    return this.db.isEmpty(options);
   }
 
   async * putMany(source: AwaitIterable<Pair<CID, Uint8Array>>, options?: Options):
--- a/src/store/data-store-level.ts
+++ b/src/store/data-store-level.ts
@@ -2,16 +2,28 @@ import type { ImportResult } from 'ipfs-unixfs-importer';
 import type { PutResult } from './data-store.js';
 
 import { BlockstoreLevel } from './blockstore-level.js';
-import { CID } from 'multiformats/cid';
 import { createLevelDatabase } from './level-wrapper.js';
 import { DataStore } from './data-store.js';
 import { exporter } from 'ipfs-unixfs-exporter';
 import { importer } from 'ipfs-unixfs-importer';
 import { Readable } from 'readable-stream';
 
+const DATA_PARTITION = 'data';
+const HOST_PARTITION = 'host';
+
+// `BlockstoreLevel` doesn't support being a `Set` (i.e. it always requires a value), so use a placeholder instead.
+const PLACEHOLDER_VALUE = new Uint8Array();
+
 /**
  * A simple implementation of {@link DataStore} that works in both the browser and server-side.
  * Leverages LevelDB under the hood.
+ *
+ * It has the following structure (`+` represents a sublevel and `->` represents a key->value pair):
+ *   'data' + <dataCid> -> <data>
+ *   'host' + <dataCid> + <tenant> + <messageCid> -> PLACEHOLDER_VALUE
+ *
+ * This allows for the <data> to be shared for everything that uses the same <dataCid> while also making
+ * sure that the <data> can only be deleted if there are no <messageCid> for any <tenant> still using it.
  */
 export class DataStoreLevel implements DataStore {
   config: DataStoreLevelConfig;
@@ -39,10 +51,10 @@ export class DataStoreLevel implements DataStore {
     await this.blockstore.close();
   }
 
-  async put(tenant: string, recordId: string, dataStream: Readable): Promise<PutResult> {
-    const partition = await this.blockstore.partition(tenant);
+  async put(tenant: string, messageCid: string, dataStream: Readable): Promise<PutResult> {
+    const dataPartition = await this.blockstore.partition(DATA_PARTITION);
 
-    const asyncDataBlocks = importer([{ content: dataStream }], partition, { cidVersion: 1 });
+    const asyncDataBlocks = importer([{ content: dataStream }], dataPartition, { cidVersion: 1 });
 
     // NOTE: the last block contains the root CID as well as info to derive the data size
     let block: ImportResult;
@@ -51,24 +63,38 @@ export class DataStoreLevel implements DataStore {
     const dataCid = block.cid.toString();
     const dataSize = block.unixfs ? Number(block.unixfs!.fileSize()) : Number(block.size);
 
+    const tenantsForData = await this.blockstore.partition(HOST_PARTITION);
+    const messagesForTenant = await tenantsForData.partition(dataCid);
+    const messages = await messagesForTenant.partition(tenant);
+
+    await messages.put(messageCid, PLACEHOLDER_VALUE);
+
     return {
       dataCid,
       dataSize
     };
   }
 
-  public async get(tenant: string, recordId: string, dataCid: string): Promise<Readable | undefined> {
-    const partition = await this.blockstore.partition(tenant);
+  public async get(tenant: string, messageCid: string, dataCid: string): Promise<Readable | undefined> {
+    const tenantsForData = await this.blockstore.partition(HOST_PARTITION);
+    const messagesForTenant = await tenantsForData.partition(dataCid);
+    const messages = await messagesForTenant.partition(tenant);
 
-    const cid = CID.parse(dataCid);
-    const bytes = await partition.get(cid);
+    const allowed = await messages.has(messageCid);
+    if (!allowed) {
+      return undefined;
+    }
+
+    const dataPartition = await this.blockstore.partition(DATA_PARTITION);
+
+    const bytes = await dataPartition.get(dataCid);
 
     if (!bytes) {
       return undefined;
     }
 
     // data is chunked into dag-pb unixfs blocks. re-inflate the chunks.
-    const dataDagRoot = await exporter(dataCid, partition);
+    const dataDagRoot = await exporter(dataCid, dataPartition);
     const contentIterator = dataDagRoot.content()[Symbol.asyncIterator]();
 
     const readableStream = new Readable({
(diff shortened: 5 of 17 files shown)


thehenrytsai:
hey @dcrousso, thanks for the PR, saying getting issue done is critical is an understatement IMO! I will leave the intense studying of the PR until tomorrow. But just to start, regarding the statement:

> switch from logicalId/recordId to messageCid as the former can be reused by multiple things (e.g. two different RecordsWrite having the same recordId) whereas the latter is derived from the entire message

To be clear, not necessarily pushing back as I have not read the code yet (it might even be better), just trying to understand clearly: Is this a functional incorrectness thing or an implementation convenience/consistency thing? I imagine the latter, because `recordId` should work since `RecordsWrite` is a clobber/replace operation, so only one version of the data needs to exist in the datastore per `recordId`. The thinking when I created the interface was for the datastore to just overwrite the existing data if it's there when `put()`, I can see why this might not be the best idea, but the original hope was that this would allow less "management code". Am I wrong anywhere?

dcrousso:
> > switch from logicalId/recordId to messageCid as the former can be reused by multiple things (e.g. two different RecordsWrite having the same recordId) whereas the latter is derived from the entire message
> 
> To be clear, not necessarily pushing back as I have not read the code yet (it might even be better), just trying to understand clearly: Is this triggered by a functional incorrectness thing or more of an implementation convenience/consistency/improvement thing? I imagine the latter, because `recordId` should work since `RecordsWrite` is a clobber/replace operation, so only one version of the data needs to exist in the datastore per `recordId`. The thinking when I created the interface was for the datastore to just overwrite the existing data if it's there when `put()`, I can see why this might not be the best idea, but the original hope was that this would allow less "management code". Am I wrong anywhere?

a bit of both? 😅

i tried playing around with `recordId` a bunch before switching to `CID` and found that it made things way uglier (especially in `deleteAllOlderMessagesButKeepInitialWrite`)

`recordId` makes the logic of keeping the data alive _much_ harder in the case that there are different `RecordsWrite` with the same `recordId` (specifically the code in `deleteAllOlderMessagesButKeepInitialWrite`)

`CID` was much easier to use as it represents a message in its entirety, rather than just a single component of it (i.e. i think you'd have to send the _exact_ same message in order for it to have the same `CID`, which im pretty sure isn't possible due to things like `dateCreated` and whatnot) meaning that two messages with the same `recordId` would each have a reference to the same data (based on the `dataCid`) thereby allowing us to drop the reference from the old message instead of having to "transfer" it to the new message without destroying the data

also, would different data result in a different `dataCid`?  not entirely sure how `ipfs-unixfs-importer` works, but if that's the case then we wouldn't be overriding the existing data anyways (since it'd be keyed on a different/"old" `dataCid`)

anyways, ill keep experimenting and maybe will find something just as clean using `recordId`

thehenrytsai:
@dcrousso, thanks for the context, with the above comments in mind, will start reviewing. Responding to a couple of things:

> specifically the code in `deleteAllOlderMessagesButKeepInitialWrite`

I would've thought the method would be left untouched (mostly?), since it deals with message, not data.

>  would different data result in a different `dataCid`?

Yes, for sure different data always results in a different `dataCid`. My point was that it is fine to override the existing data if it's keyed off of `recordId`, since we only want one/latest version anyway.

Using `messageCid` would work, but it seems like it'll necessitate the need for reference count (haven't looked into the PR yet), where using `recordId` it seems you might get away with not doing reference count.

dcrousso:
> > specifically the code in `deleteAllOlderMessagesButKeepInitialWrite`
> 
> I would've thought the method would be left untouched (mostly?), since it deals with message, not data.

i was trying to modify it as a "delete all older messages _and their data_ (but keep the initial write without data)" but have since done it slightly differently

> > would different data result in a different `dataCid`?
> 
> Yes, for sure different data always results in a different `dataCid`. My point was that it is fine to override the existing data if it's keyed off of `recordId`, since we only want one/latest version anyway.
> 
> Using `messageCid` would work, but it seems like it'll necessitate the need for reference count (haven't looked into the PR yet), where using `recordId` it seems you might get away with not doing reference count.

FWIW this entire system is kinda an attempt at doing reference counting, so i think it wouldve worked (unless somehow someone crafted a message that computed to the same `messageCid`)

i did some more experimenting (and asked some questions offline) and managed to make it all work using `recordId` :)

src/core/message.ts
```diff
@@ -113,6 +113,13 @@ export abstract class Message {
     return currentNewestMessage;
   }
 
+  /**
+   * Gets the recordId of the given message.
+   */
+  public static getRecordId(message: BaseMessage): string {
+    return message['recordId'] ?? message.descriptor['recordId'];
```
thehenrytsai, line 120: `recordId` is a  concept only to `Records` message, so to have a method for getting the `recordId` here is odd. Also if `message['recordId']` is `undefined, it doesn't mean `message.descriptor['recordId']` will be defined, maybe the reason for doing this is just a "best effort" to get the `recordId`?

I feel the alternatives below maybe "better" given the lack of options:
1. declare `recordId?: string` in `BaseMessage`, or if you feel this has practically the same issue
2. straight up doing `message['recordId']` in the places that invokes `getRecordId(...)` or
3. maybe express the true intent by having an abstract instanced method `getLogicalId()` where it is the `recordId` for `RecordsWrite`, and it's the `protocol` (URL) for `ProtocolConfigure`, but this change is more involved because you'll need to touch all interface methods, and logical ID does not apply to all methods (e.g. `RecordsQuery`) so in that case you'll probably have to return `undefined`... just more involved than you'd care for.

Just publishing my thought as I go (so that you don't get all comments in a big bang) and might revisit as I gain more context by reviewing further, I can already see by skimming through that maybe your idea of using `messageCid` is much better after all as it probably can sidestep all this discussion.

src/interfaces/records/records-interface.ts
```diff
@@ -1,10 +1,56 @@
...
+  for (const message of existingMessages) {
+    const messageIsOld = await Message.isOlder(message, comparedToMessage);
+    if (!messageIsOld) {
+      continue;
+    }
+
+    // ignore anything that's not the same record
+    if (Message.getRecordId(message) !== recordId) {
+      continue;
+    }
```
thehenrytsai, line 35: Consider either:

- remove this check as this code path can't be hit since all messages given will be in the scope of the same `recordId` already; or
- if you intended for this method to be more general purpose, add a test to cover the path

src/core/message.ts
dcrousso: yeah this was by best effort attempt at getting some sort of `recordId`

there isn't a `BaseRecord` class that `RecordsWrite`/`RecordsDelete`/etc. inherit from, so i felt this was the next best option (tho i could add that)

the reason im doing both is because `recordId` is defined on

- `RecordsWriteMessage` (_not_ `RecordsWriteDescriptor`)
- `RecordsDeleteDescriptor` (_not_ `RecordsDeleteMessage`)

so there's no consistency between them as to where it will be, and `StorageController`/`deleteAllOlderData`/etc. only know about `TimestampedMessage` so they have to handle both scenarios (btw this would be improved by [having actual `class` instead of manipulating the raw JSON objects](https://github.com/TBD54566975/dwn-sdk-js/issues/230#issuecomment-1435385854))

src/interfaces/records/records-interface.ts
```diff
@@ -1,10 +1,56 @@
...
+
+    // we can only delete if we have a dataCid
+    if (message.descriptor.dataCid === undefined) {
+      continue;
+    }
```
thehenrytsai, line 40: Add a test to cover this path. This would be the case of a `RecordsWrite` after a `RecordsDelete`.

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -58,6 +58,8 @@ export class RecordsDeleteHandler implements MethodHandler {
       newestMessage = newestExistingMessage;
     }
 
+    await deleteAllOlderData(tenant, existingMessages, newestMessage, this.dataStore);
```
thehenrytsai, line 61: Can we move `deleteAllOlderData()` call after `deleteAllOlderMessagesButKeepInitialWrite()` call below (as the last call).

Reason being mainly in case `put()` blows up, the record wouldn't be in a bad state, and secondarily it's nice that data state follows the message state.

src/interfaces/records/handlers/records-write.ts
```diff
@@ -75,6 +74,8 @@ export class RecordsWriteHandler implements MethodHandler {
       newestMessage = newestExistingMessage;
     }
 
+    await deleteAllOlderData(tenant, existingMessages, newestMessage, this.dataStore);
```
thehenrytsai, line 77: Same comment as the other `deleteAllOlderData()` placement.

src/store/blockstore-level.ts
```diff
@@ -40,20 +40,24 @@ export class BlockstoreLevel implements Blockstore {
...
-  async put(key: CID, val: Uint8Array, options?: Options): Promise<void> {
-    return this.db.put(key.toString(), val, options);
+  async isEmpty(options?: Options): Promise<boolean> {
+    return this.db.isEmpty(options);
   }
 
-  async get(key: CID, options?: Options): Promise<Uint8Array> {
-    return this.db.get(key.toString(), options);
+  async put(key: CID | string, val: Uint8Array, options?: Options): Promise<void> {
+    return this.db.put(String(key), val, options);
```
thehenrytsai, line 48: for my learning: is there a functional difference here with switching to `String(key)`, or it's just a preference thing?

dcrousso: its a safety precaution in case someone accidentally provides `null`/`undefined` as an argument, as neither have a `.toString()`

src/store/blockstore-level.ts
```diff
@@ -40,20 +40,24 @@ export class BlockstoreLevel implements Blockstore {
...
   }
 
-  async has(key: CID, options?: Options): Promise<boolean> {
-    return !! await this.get(key, options);
+  async get(key: CID | string, options?: Options): Promise<Uint8Array> {
```
thehenrytsai, line 51: for my understanding: I don't understand how adding `| string` would work (though it clearly does), the type declaration for `Blockstore` says:

`export interface Blockstore extends Store<CID, Uint8Array>`

so I though the key has to be a `CID`? Is the type definition out of date, or am I understanding things incorrectly?

src/store/blockstore-level.ts
thehenrytsai: Got it, chance is low since there are type checks yeah? But no push back on precautionary coding, it's better.

dcrousso: the types only help you if you're using TypeScript :)

src/store/blockstore-level.ts
dcrousso: im pretty sure the `|` is an "or", so it still satisfies the requirement from `Blockstore` that the `key` is a `CID` while also allowing for `string` (which ultimately is what it gets turned into anyways inside `BlockstoreLevel` for use with `LevelWrapper`)

src/store/data-store-level.ts
```diff
@@ -86,21 +134,48 @@ export class DataStoreLevel implements DataStore {
...
+    const recordForCid = await this.blockstore.partition(HOST_PARTITION);
+    const tenantForRecord = await recordForCid.partition(dataCid);
+    const dataForTenant = await tenantForRecord.partition(recordId);
+    const tenants = await dataForTenant.partition(tenant);
 
-    const cid = CID.parse(dataCid);
-    const rootBlockBytes = await partition.get(cid);
+    const allowed = await tenants.has(tenant);
+    if (!allowed) {
+      return undefined;
```
thehenrytsai, line 144: sanity: `undefined` is not an expected return type. Did you mean to return `false` here? Also consider to just remove this method completely as it appears to be not referenced. If you think we should keep it for whatever reason, add tests for it as it is not tested.

src/store/data-store-level.ts
```diff
@@ -85,22 +133,32 @@ export class DataStoreLevel implements DataStore {
     return readableStream;
   }
 
-  public async has(tenant: string, recordId: string, dataCid: string): Promise<boolean> {
-    const partition = await this.blockstore.partition(tenant);
+  public async delete(tenant: string, recordId: string, dataCid: string): Promise<void> {
+    const recordForCid = await this.blockstore.partition(HOST_PARTITION);
+    const tenantForRecord = await recordForCid.partition(dataCid);
+    const dataForTenant = await tenantForRecord.partition(recordId);
+    const tenants = await dataForTenant.partition(tenant);
```
thehenrytsai, line 140: sanity: the plural `tenants` currently throws me off, as `dataForTenant.partition(tenant)` already selects a partition base on a tenant, why would this be plural, same question mark for occurrences in above methods. Thus why would `tenants.isEmpty()` ever be `false` on line 144 after `tenants.delete(tenant)? Code coverage result also hints at the same thing as the `if (!wasLastTenant)` condition being `true` is never covered, or maybe that's coincidental and that we are just missing (thus ought to add) a test to cover the case of deleting data that is referenced in multiple tenants, if the test doesn't exist already. Will keep reading.

dcrousso: oh lol yup im pretty sure we can just remove that

i think it was a leftover from an earlier design 😅

src/store/storage-controller.ts
```diff
@@ -28,15 +29,17 @@ export class StorageController {
     indexes: { [key: string]: string },
     dataStream?: Readable
   ): Promise<void> {
+    const recordId = Message.getRecordId(message);
```
thehenrytsai, line 32: Looking at this, your idea of using `messageCid` as opposed to the original intent of "logical ID' probably ends up being a better idea after all..... 

The problem with directly referencing of `recordId` here is that the storage controller is intended be generic and thus able to handle storing of data regardless of message being a Record, or a Protocol Definition, or a PermissionGrant or HookDefinition whatnot. Which is why I named the ID `logicalId` and not `recordId`, but if we follow that pattern, we'd need to have a method somewhere called `getLogicalId()`.

But your idea of using `messageCid` would sidestep the above and still work.... do you mind switching to `messageCid` given the above reasoning?

dcrousso: LOL sure yeah i can go back to using `messageCid` :)

src/store/storage-controller.ts
```diff
@@ -28,15 +29,17 @@ export class StorageController {
...
+
     // if `dataCid` is given, it means there is corresponding data associated with this message
     // but NOTE: it is possible that a data stream is not given in such case, for instance,
     // a subsequent RecordsWrite that changes the `published` property, but the data hasn't changed,
     // in this case requiring re-uploading of the data is extremely inefficient so we take care allow omission of data stream
-    if (message.descriptor.dataCid !== undefined) {
+    if (recordId !== undefined && message.descriptor.dataCid !== undefined) {
       if (dataStream === undefined) {
         // the message implies that the data is already in the DB, so we check to make sure the data already exist
         // TODO: #218 - Use tenant + record scoped IDs - https://github.com/TBD54566975/dwn-sdk-js/issues/218
```
thehenrytsai, line 41: Remove the TODO since you've done it!

src/store/data-store-level.ts
```diff
@@ -85,22 +111,47 @@ export class DataStoreLevel implements DataStore {
...
-    return (rootBlockBytes !== undefined);
+    await tenants.put(tenant, PLACEHOLDER_VALUE);
+
+    return true;
   }
 
-  public async delete(tenant: string, recordId: string, dataCid: string): Promise<void> {
-    const partition = await this.blockstore.partition(tenant);
+  public async delete(tenant: string, messageCid: string, dataCid: string): Promise<void> {
+    const messagesForData = await this.blockstore.partition(HOST_PARTITION);
```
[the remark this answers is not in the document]
dcrousso: FWIW this is always a possibility with an `AbortSignal`, so not sure there's really much we can do if we allow for things to be cancelled midway

tests/interfaces/records/handlers/records-delete.spec.ts
```diff
@@ -120,6 +121,141 @@ describe('RecordsDeleteHandler.handle()', () => {
...
+      const aliceQueryAfterAliceWriteReply = await dwn.processMessage(alice.did, aliceQueryAfterAliceWriteData.message);
+      expect(aliceQueryAfterAliceWriteReply.status.code).to.equal(200);
+      expect(aliceQueryAfterAliceWriteReply.entries.length).to.equal(1);
+      expect(aliceQueryAfterAliceWriteReply.entries[0].encodedData).to.equal(encodedData);
+
+      // data store verify
+      await expect(fromAsync(dataPartition.db.keys())).to.eventually.eql([ dataCid ]);
+
+      // bob write
+      const bobWriteData = await TestDataGenerator.generateRecordsWrite({ requester: bob, dataCid, dataSize: 4 });
```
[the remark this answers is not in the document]
dcrousso: in order for us to determine the `dataCid` of the stream we'd have to read the stream in its entirety, which is almost all the way to doing what `put` does right now (the difference being that we wouldn't have to actually write it to disk)

or are you specifically referring to the scenario where _both_ a `dataCid` and stream are provided?  in that case, wouldn't that invalidate the check in `StorageController.put` that makes sure the `dataCid` of the stream matches the provided `dataCid`?

src/store/data-store-level.ts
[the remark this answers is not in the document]
dcrousso: also wouldn't it be theoretically possible for the client to send another `RecordsWrite` to clear up any data if something bad happened while trying to handle the first/previous `RecordsWrite`?

tests/interfaces/records/handlers/records-delete.spec.ts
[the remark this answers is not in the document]
thehenrytsai: `dataCid` is alway there in the message at processing time, it's only optional for `create()` method for dev convenience. But thinking about it more, bob being able to just reference the `dataCid` and not need to supply a stream if it exists in another tenant (as shown in the test) is bad right? 'coz it means bob can gain access to any data by just knowing its CID?! Am I wrong here?! If not, I think it's important to check that the data stream already exists under the tenant (bob) already to allow them to bypass passing the data stream.

Same condition goes to my proposed optimization, but yeah, it's really optional and low pri, and I am okay for you to not file an issue on this.

src/store/data-store-level.ts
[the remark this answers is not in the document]
thehenrytsai: >  possible for the client to send another RecordsWrite

Did you mean `RecordsDelete`? If so, yes, they can submit another one to fix the linger data not being deleted. But the problem is that they probably wouldn't know there is a lingering piece of data, because it would be essentially in a "detatched state" if I understand correctly.

> this is always a possibility with an `AbortSignal`

Haven't thought deeply about this, but if we didn't bother with reference counting and go with the dumb data duplication per tenant approach, we might be able to perform an atomic delete at this datastore level. Orthogonally this might also make it simple to count total storage usage per tenant.

Again not really asking you to change the implementation, just thought it's worthwhile calling out limitation so we know what's up if/when we see strange behaviors relating to this.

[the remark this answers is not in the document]
dcrousso: oop yes i meant `RecordsDelete`

yeah fair point it would require things to be perfect in order to know that something went wrong

ill create an issue to look into ways of making storage stuff atomic (or at least safer)

tests/interfaces/records/handlers/records-delete.spec.ts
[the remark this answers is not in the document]
dcrousso: oho good catch yeah this would allow for that

im pretty sure i can fix that ill take a look

src/store/data-store-level.ts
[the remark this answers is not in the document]
dcrousso: created <https://github.com/TBD54566975/dwn-sdk-js/issues/267>

tests/interfaces/records/handlers/records-delete.spec.ts
```diff
@@ -120,6 +121,376 @@ describe('RecordsDeleteHandler.handle()', () => {
...
+
+      await expect(fromAsync(dataPartition.db.keys())).to.eventually.eql([ dataCid ]);
+
+      const bobAssociateData = await TestDataGenerator.generateRecordsWrite({
+        requester : bob,
+        dataCid,
+        dataSize  : 4
+      });
+      const bobAssociateReply = await dwn.processMessage(bob.did, bobAssociateData.message, bobAssociateData.dataStream);
+      expect(bobAssociateReply.status.code).to.equal(400);
```
thehenrytsai, line 326: Just to make sure it's not some other error throwing 400 due to buggy test whatnot (happens more than I'd admit when I write tests)

```suggestion
      expect(bobAssociateReply.status.code).to.equal(400);
      expect(bobAssociateReply.status.detail).to.contain(DwnErrorCode.MessageStoreDataNotFound);
```

thehenrytsai (commented):
Thanks for adding a whole bunch of tests on data references what not... it's probably time to consider refactoring that file in a future PR soon... Also, I would've tried to take the dumber approach of data duplication/separation at the tenant level, but thanks for going the extra mile, can't wait to see this code running in the wild!

Just one last comment (and a rebase) and we are good to go!

PR #271, opened 2023-03-21
various minor documentation fixes plus a minor function rename

thehenrytsai:


diehuxx (approved):

src/utils/array.ts
```diff
@@ -1,5 +1,9 @@
-export async function fromAsync<T>(iterator: AsyncGenerator<T>): Promise<Array<T>> {
```
diehuxx, line 1: I think this was trying to mirror [this polyfill](https://github.com/tc39/proposal-array-from-async). In a PR last week, Devin added that polyfill, then removed to avoid importing one-off libraries, and added this lil implementation. Personally I don't feel that strongly either way, but in my crystal ball I foresee this function maybe getting re-renamed back to `fromAsync`.

PR #275, opened 2023-03-22
#84 - deleting a `dataCid` should delete all blocks, not just the root

dcrousso:
`ipfs-unixfs-importer` will split the `DataStream` into blocks, storing each in the `Blockstore` under a separate `CID`

unfortunately, we can only get the `CID` of the root block, so instead just leverage `level` to create a separate `partition` for the given `dataCid` and use that to delete everything in one go

ideally we'd have a way to get the `CID` of every block so that we can have the same `tenant`+`messageCid` association logic with each block instead of just the root (though this could result in a much larger metadata storage cost, but could also allow for deduplication of individual blocks)

### DIFF
--- a/src/store/data-store-level.ts
+++ b/src/store/data-store-level.ts
@@ -1,5 +1,5 @@
 import type { ImportResult } from 'ipfs-unixfs-importer';
-import type { DataStore, PutResult } from './data-store.js';
+import type { AssociateResult, DataStore, GetResult, PutResult } from './data-store.js';
 
 import { BlockstoreLevel } from './blockstore-level.js';
 import { createLevelDatabase } from './level-wrapper.js';
@@ -50,31 +50,29 @@ export class DataStoreLevel implements DataStore {
     await this.blockstore.close();
   }
 
-  async put(tenant: string, messageCid: string, dataStream: Readable): Promise<PutResult> {
-    const dataPartition = await this.blockstore.partition(DATA_PARTITION);
-
-    const asyncDataBlocks = importer([{ content: dataStream }], dataPartition, { cidVersion: 1 });
-
-    // NOTE: the last block contains the root CID as well as info to derive the data size
-    let block: ImportResult;
-    for await (block of asyncDataBlocks) { ; }
-
-    const dataCid = block.cid.toString();
-    const dataSize = block.unixfs ? Number(block.unixfs!.fileSize()) : Number(block.size);
-
+  async put(tenant: string, messageCid: string, dataCid: string, dataStream: Readable): Promise<PutResult> {
     const tenantsForData = await this.blockstore.partition(HOST_PARTITION);
     const messagesForTenant = await tenantsForData.partition(dataCid);
     const messages = await messagesForTenant.partition(tenant);
 
     await messages.put(messageCid, PLACEHOLDER_VALUE);
 
+    const blocksForData = await this.blockstore.partition(DATA_PARTITION);
+    const blocks = await blocksForData.partition(dataCid);
+
+    const asyncDataBlocks = importer([{ content: dataStream }], blocks, { cidVersion: 1 });
+
+    // NOTE: the last block contains the root CID as well as info to derive the data size
+    let dataDagRoot: ImportResult;
+    for await (dataDagRoot of asyncDataBlocks) { ; }
+
     return {
-      dataCid,
-      dataSize
+      dataCid  : String(dataDagRoot.cid),
+      dataSize : Number(dataDagRoot.unixfs?.fileSize() ?? dataDagRoot.size)
     };
   }
 
-  public async get(tenant: string, messageCid: string, dataCid: string): Promise<Readable | undefined> {
+  public async get(tenant: string, messageCid: string, dataCid: string): Promise<GetResult | undefined> {
     const tenantsForData = await this.blockstore.partition(HOST_PARTITION);
     const messagesForTenant = await tenantsForData.partition(dataCid);
     const messages = await messagesForTenant.partition(tenant);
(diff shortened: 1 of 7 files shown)


src/store/data-store-level.ts
```diff
@@ -50,27 +50,25 @@ export class DataStoreLevel implements DataStore {
     await this.blockstore.close();
   }
 
-  async put(tenant: string, messageCid: string, dataStream: Readable): Promise<PutResult> {
-    const dataPartition = await this.blockstore.partition(DATA_PARTITION);
+  async put(tenant: string, messageCid: string, dataCid: string, dataStream: Readable): Promise<PutResult> {
+    const blocksForData = await this.blockstore.partition(DATA_PARTITION);
+    const blocks = await blocksForData.partition(dataCid);
```
thehenrytsai, line 55: If I read the code correctly (regardless of moving CID check logic in the other comment or not), when `dataCid` mismatches, the "real" existing `dataCid` partition can be filled with junk blocks by the "fake" data claiming to have the same `dataCid`. If someone repeatedly exploit this then the data partition for that `dataCid` can grow until storage cap is reached (if there is such a cap) 😭 

A quick workaround could be that we adopt the optional optimization (now it becomes necessity) I was talking about in the previous PR: ignore data stream entirely if data already exists with the claimed `dataCid`, update the `put()` description in data store interface accordingly for implementors of the data store.

thehenrytsai (changes requested):
Looks like a viable implementation. A couple of changes suggested to handle CID mismatch cases.

src/store/data-store-level.ts
```diff
@@ -50,27 +50,25 @@ export class DataStoreLevel implements DataStore {
...
     const tenantsForData = await this.blockstore.partition(HOST_PARTITION);
     const messagesForTenant = await tenantsForData.partition(dataCid);
     const messages = await messagesForTenant.partition(tenant);
 
     await messages.put(messageCid, PLACEHOLDER_VALUE);
 
     return {
-      dataCid,
-      dataSize
+      dataCid  : block.cid.toString(),
```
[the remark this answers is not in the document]
dcrousso: perhaps it was a mistake on my part also calling this `dataCid`.  i almost see it more as a "give this big blob of data a name" rather than "what is the actual `CID` of this data"

im not necessarily against moving the exceptions into this, but i think the problem there is that if the developer provides their own `DataStore` then suddenly we'd lose these safety checks (and along those lines id kinda rather `DataStoreLevel` assume everyone is a good actor so that it's logic can stay clean)

src/store/data-store-level.ts
dcrousso: technically yes, but im pretty sure those blocks would immediately get destroyed by `StorageController.put` since the input `dataCid` wouldn't match the output `dataCid` (tho this would present a different issue in that when it would delete the added data, the entire bucket for `dataCid` would get deleted instead of just the newly added data)

the problem with the optimization you describe is that we'd also need to store the `dataSize` somewhere (or figure out a way of extracting that from `blocks`) in order to maintain the same `PutResult`.  ill investigate how easy it would be to do that

thehenrytsai: > im pretty sure those blocks would immediately get destroyed by StorageController.put since the input dataCid wouldn't match the output dataCid

It will try, but it currently looks into the wrong CID (the CID that is returned by the datastore), if it used the alleged CID, it would try to delete, but when other references to the same `dataCid` exists, data-store will leave it alone, that's my understanding.

> figure out a way of extracting that from blocks
That's would be ideal.

dcrousso: ah oops yes `StorageController.put` uses the returned `dataCid` instead of the `message.descriptor.dataCid`.  will change/fix :)

src/store/data-store-level.ts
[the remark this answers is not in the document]
thehenrytsai: Agree with you that it'd be ideal that we can lift any burden away from the data store implementation AMAP, but here we are, requiring `dataCid` as part of the interface mainly because level impl needs it, this kinda is going against that spirit, so it was sorta of a tradeoff situation. In the same spirit, the "bypassing data stream if data of CID exists check" logic would also be better to be outside the data store implementation?

I want to be agreeable to the PR and there are better hills to die on. If you can add the documentation of expectation of `dataCid` param in the `put` method signature of the `Datastore` interface (along with the condition for ignoring data stream if required), then we can call this resolved.

The original thinking was that a "thorough/defensive" programmer would end up doing the CID check as it is odd to not check it given it is the expected `dataCid` after all. Which then begs the question why have it in two places if they were gonna check it.

src/store/data-store-level.ts
```diff
@@ -127,6 +124,14 @@ export class DataStoreLevel implements DataStore {
       return false;
     }
 
+    const blocksForData = await this.blockstore.partition(DATA_PARTITION);
+    const blocks = await blocksForData.partition(dataCid);
+
+    const exists = await blocks.has(dataCid);
+    if (!exists) {
+      return false;
```
thehenrytsai, line 132: Add a test for this path.

src/store/data-store-level.ts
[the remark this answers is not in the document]
dcrousso: yeah i suppose it's somewhat expected that a developer building their own `DataStore` would do the right thing if given both a `dataCid` and `dataSize` (in addition to the `dataStream`) so i suppose we could move the checks inside `DataStoreLevel`.  put another way, there's no guarantee the developer wouldnt lie when returning the `dataCid` or `dataSize` (though if we dont provide the `dataSize` as an argument then they'd at least have to get that part right, which i suppose was part of the idea when neither `dataCid` or `dataSize` were provided)

lemme try moving the exceptions into `DataStoreLevel` and see how it looks/feels 🤔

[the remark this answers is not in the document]
dcrousso: yeah IMO it felt kinda funky to put it in the `DataStoreLevel`, especially since the check for `associate` was not, which made me want to put that check inside too

instead i felt it was better to unify everything and have all the methods return both a `dataCid` and `dataSize` for verification at every step

Issue #261, opened 2023-03-15
add stricter TypeScript check in tsconfig.json


thehenrytsai:
Add `"strict": true"` to `tsconfig.json/tsconfig.cjs.json` with `"strictNullChecks": true` being the fallback option if `"strict": true"` is too painful to enforce.

This will allow projects having stricter TS config rules to use the library without hitting errors.

duncanmak:
Hello @thehenrytsai and @diehuxx, thanks for merging in my PRs.

Do you think this issue should be closed now? And keep any follow-ups in #298?

thehenrytsai:
@duncanmak , was contemplating if we should wait until the "FIXME" are fixed. But I'll  close and open a separate issue for those too.

Issue #242, opened 2023-03-02
Add participants actor to Protocol rules


csuwildcat:

diehuxx:
Since we added the new protocol `inGroup` rules, should we close this or move it to the backlog? The existing rules cover most of the use cases that we wanted out of `participants` AFAIU.

thehenrytsai:
Superseded by #489.

PR #277, opened 2023-03-22
Stricten typescript in protocol-authorization.ts

diehuxx:
Taking a small step towards getting this repo onto strict typescript rules (#261). Starting with the file `protocol-authorization.ts` in preparation for #242.

Typescript doesn't support per-file strictness out of the box, but `typescript-strict-plugin` does. In future PRs, if you edit a file, please add it to the list of strict files in the tsconfig 🙏.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,7 +76,6 @@
         "search-index": "3.4.0",
         "sinon": "13.0.1",
         "typescript": "4.8.2",
-        "typescript-strict-plugin": "^2.1.0",
         "util": "0.12.4"
       },
       "engines": {
@@ -1873,27 +1872,6 @@
         "node": ">=12"
       }
     },
-    "node_modules/cli-cursor": {
-      "version": "3.1.0",
-      "resolved": "https://artifactory.global.square/artifactory/api/npm/square-npm/cli-cursor/-/cli-cursor-3.1.0.tgz",
-      "integrity": "sha512-I/zHAwsKf9FqGoXM4WWRACob9+SNukZTd94DWF57E4toouRulbCxcUh6RKUEOQlYTHJnzkPMySvPNaaSLNfLZw==",
-      "dev": true,
-      "dependencies": {
-        "restore-cursor": "^3.1.0"
-      },
-      "engines": {
-        "node": ">=8"
-      }
-    },
-    "node_modules/cli-spinners": {
-      "version": "2.7.0",
-      "resolved": "https://artifactory.global.square/artifactory/api/npm/square-npm/cli-spinners/-/cli-spinners-2.7.0.tgz",
-      "integrity": "sha512-qu3pN8Y3qHNgE2AFweciB1IfMnmZ/fsNTEE+NOFjmGB2F/7rLhnhzppvpCnN4FovtP26k8lHyy9ptEbNwWFLzw==",
-      "dev": true,
-      "engines": {
-        "node": ">=6"
-      }
-    },
     "node_modules/cliui": {
       "version": "7.0.4",
       "resolved": "https://registry.npmjs.org/cliui/-/cliui-7.0.4.tgz",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -115,6 +115,7 @@
     "search-index": "3.4.0",
     "sinon": "13.0.1",
     "typescript": "4.8.2",
+    "typescript-strict-plugin": "^2.1.0",
```
thehenrytsai, line 118: Are you adding this dependency temporarily until we have moved everything to stricter checks?

An alternative approach to make progress for #261 would be to turn `"strict": true` on in `tsconfig` and fix what you can/want then comment out the rule again.

I just do now want:
1. for this dependency to be permanent; and
1. for us to add every file in the `paths` in both `tsconfig` files in order to complete #261, turning on the strict rule itself seems to be easier to be sure and know the actual progress.

src/core/protocol-authorization.ts
```diff
@@ -147,7 +155,11 @@ export class ProtocolAuthorization {
       }
 
       // else we keep going down the message chain
-      allowedRecordsAtCurrentLevel = allowedRecordsAtCurrentLevel[currentRecordType].records;
+      const nextAllowedRecords = allowedRecordsAtCurrentLevel[currentRecordType].records;
+      if (nextAllowedRecords === undefined) {
+        throw new Error(`No record types found in schema '${currentRecordSchema}' in structure level ${currentMessageIndex}`);
```
thehenrytsai, line 160: Seems legitimate case caught by TS. Please add a test for it.

package-lock.json
```diff
@@ -1872,6 +1873,27 @@
         "node": ">=12"
       }
     },
+    "node_modules/cli-cursor": {
+      "version": "3.1.0",
+      "resolved": "https://artifactory.global.square/artifactory/api/npm/square-npm/cli-cursor/-/cli-cursor-3.1.0.tgz",
```
thehenrytsai, line 1878: Sanity: are you sure all these tens of `https://artifactory.global.square` square specific reference won't break install? They break for me.

src/core/protocol-authorization.ts
```diff
@@ -207,7 +219,10 @@ export class ProtocolAuthorization {
       allowedActions = allowRule.anyone.to;
     } else if (allowRule.recipient !== undefined) {
       allowedActions = allowRule.recipient.to;
-    } // not possible to have `else` because of same check already done by verifyAllowedRequester()
+    } else {
+      // not possible to have `else` because of same check already done by verifyAllowedRequester()
+      allowedActions = [];
+    }
```
thehenrytsai, line 225: instead just do:
`let allowedActions: string[] = [];`

src/core/types.ts
```diff
@@ -65,5 +70,5 @@ export type LT = ({ lt: string } & { lte?: never }) | ({ lt?: never } & { lte: s
 export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
 
 export type Filter = {
-  [property: string]: EqualFilter | OneOfFilter | RangeFilter
+  [property: string]: DwnInterfaceFilter | DwnMethodFilter | EqualFilter | OneOfFilter | RangeFilter
```
thehenrytsai, line 73: Sanity: is this really necessary? IMO this is much more confusing than what it was before.

src/core/protocol-authorization.ts
thehenrytsai, line 1: The code coverage dropped from 100% to ~92%. In general, when this happens it really begs a closer look, typically it means either the code added is redundant (unreachable), or there are missing tests. Either way makes the case to bring the code coverage back to 100%.

package.json
diehuxx: Yes, the intention was to use this temporarily until we finish migrating to strict. I don't feel very strongly about this approach, so turning on `"strict": true` while developing works for me.

package-lock.json
diehuxx: Ah, sorry didn't realize it was adding artifactory stuff. In any case, I removed the package from this PR

src/core/types.ts
diehuxx: Removed

src/core/protocol-authorization.ts
diehuxx: Took a closer look, and you're absolutely right. I added a test and made a slight change to how we're handling an error. Let me know what you think.

PR #278, opened 2023-03-23
Enable strict mode

duncanmak:
This begins the work on #261.

There are a few places where I'm not sure what to do, so I marked those with `FIXME`s.


duncanmak:
This is still a draft and doesn't compile yet. 

Here's a listing of places that need fixing:

* `protocol-authorization.ts`: `methodToAllowedActionMap` is not an instance of `Map`

* `dwn.ts`: `messageStore` and `dataStore` don't have string indexers, this is used by `dump`.

* `protocols-configure.ts`:
`permission-request.ts`: The type of `index` is problematic.

* `records-read.ts`: Cannot type the arguments to `handle`.

* `level-wrapper.ts`:
`data-store-level.ts`: Not sure what to return for the empty case.

* `level-wrapper.ts`: The return type of `get` doesn't allow for `undefined`, but the API needs to be changed?

src/interfaces/protocols/handlers/protocols-configure.ts
```diff
@@ -62,7 +58,9 @@ export class ProtocolsConfigureHandler implements MethodHandler {
         author,
         ... message.descriptor
       };
-      await StorageController.put(this.messageStore, this.dataStore, tenant, incomingMessage, indexes, dataStream);
+      // FIXME: indexes, Property 'dataSize' is incompatible with index signature.
```
diehuxx, line 61: Hmm `indexes` ultimately gets passed down to `IndexLevel#put` as an `Entry` which is very general.
```typescript
type Entry = {
  [property: string]: unknown
};
```
There's a couple ways we could go with this.

@dcrousso What's your recommendation here? Seems like an actual issue if someone loads their own message store and we pass them an object with an unexpected shape.

src/interfaces/records/handlers/records-write.ts
```diff
@@ -83,12 +77,11 @@ export class RecordsWriteHandler implements MethodHandler {
       try {
         await StorageController.put(this.messageStore, this.dataStore, tenant, incomingMessage, indexes, dataStream);
       } catch (error) {
-        if (error.code === DwnErrorCode.MessageStoreDataCidMismatch ||
-            error.code === DwnErrorCode.MessageStoreDataNotFound ||
-            error.code === DwnErrorCode.MessageStoreDataSizeMismatch) {
-          return new MessageReply({
-            status: { code: 400, detail: error.message }
-          });
+        const e = error as any; // FIXME, support code filters in MessageReply.fromError?
```
diehuxx, line 80: TBH I like that the error code check is explicit here. Makes it easy to cmd+F. Adding filters to `MessageReply.fromError` seems like overkill.

src/interfaces/records/messages/records-query.ts
```diff
@@ -77,8 +72,8 @@ export class RecordsQuery extends Message {
       switch (key) {
       case 'dateCreated':
         var rangeFilter: RangeFilter = {
-          gte : filter.dateCreated.from,
-          lt  : filter.dateCreated.to
+          gte : filter.dateCreated?.from!,
+          lt  : filter.dateCreated?.to!
```
diehuxx, line 76: Since we're in the `case 'dateCreated'`, we should be good to `filter.dateCreated!`

src/interfaces/records/messages/records-read.ts
```diff
@@ -55,12 +55,9 @@ export class RecordsRead {
       date      : options.date ?? currentTime
     };
 
-    const message: RecordsReadMessage = { descriptor };
-
-    if (options.authorizationSignatureInput) {
-      const authorization = await Message.signAsAuthorization(descriptor, options.authorizationSignatureInput);
-      message.authorization = authorization;
-    }
+    // FIXME: what if authorizationSignatureInput is undefined?
```
diehuxx, line 58: We should make `authorizationSignatureInput` required in `RecordsQueryOptions`. Every other message has it required, so it's probably a typo that it's optional here.

src/jose/algorithms/signing/secp256k1.ts
```diff
@@ -57,7 +57,7 @@ export const secp256k1: Signer = {
     validateKey(publicJwk);
 
     const xBytes = Encoder.base64UrlToBytes(publicJwk.x);
-    const yBytes = Encoder.base64UrlToBytes(publicJwk.y);
+    const yBytes = Encoder.base64UrlToBytes(publicJwk.y!);
```
diehuxx, line 60: Don't think it's safe to say `.y` is guaranteed here
```suggestion
    let yBytes = new Uint8Array([]);
    if (publicJwk?.y !== undefined) {
      yBytes = Encoder.base64UrlToBytes(publicJwk.y);
    }
```

src/store/blockstore-level.ts
```diff
@@ -45,7 +45,8 @@ export class BlockstoreLevel implements Blockstore {
   }
 
   async get(key: CID | string, options?: Options): Promise<Uint8Array> {
-    return this.db.get(String(key), options);
+    const result = await this.db.get(String(key), options);
+    return result!;
```
diehuxx, line 49: Instead of `result!`, let's update return type to `Promise<Uint8Array | undefined>

src/store/level-wrapper.ts
```diff
@@ -112,7 +113,7 @@ export class LevelWrapper<V> {
   async * keys(options?: LevelWrapperOptions): AsyncGenerator<string> {
     options?.signal?.throwIfAborted();
 
-    await abortOr(options?.signal, this.createLevelDatabase());
+    await abortOr(options?.signal!, this.createLevelDatabase());
```
diehuxx, line 116: `abortOr` actually checks if `signal` is undefined. We should update the method signature to allow undefined, instead of doing `.signal!` here.

src/store/data-store-level.ts
```diff
@@ -63,9 +63,11 @@ export class DataStoreLevel implements DataStore {
     const asyncDataBlocks = importer([{ content: dataStream }], blocks, { cidVersion: 1 });
 
     // NOTE: the last block contains the root CID as well as info to derive the data size
-    let dataDagRoot: ImportResult;
+    let dataDagRoot: ImportResult | undefined = undefined;
     for await (dataDagRoot of asyncDataBlocks) { ; }
 
+    if (dataDagRoot == undefined) { return { dataCid, dataSize: 0 }; } // FIXME: Is this a valid empty PutResult?
```
diehuxx, line 69: @thehenrytsai Curious what your thoughts are here. IIUC, `dataDagRoot` should only be undefined if there is an issue getting `asyncDataBlocks` from IPFS. In that case, we should throw an exception. Does that sound right to you?

diehuxx (commented):
Hi @duncanmak ! Thanks for the contribution. Here's my first round of comments, I'll finish reviewing later today or tomorrow.

Let me know if you have time to go through these this week. If not, I'm happy to make the changes for you.

src/interfaces/protocols/handlers/protocols-configure.ts
dcrousso: IMO we should allow for anything as the value.  i dont think it's super flexible for `MessageStore` to require that the `indexes` be a flat structure.  that's just how we do it :)

src/core/message.ts
```diff
@@ -25,14 +25,14 @@ export enum DwnMethodName {
   Delete = 'Delete'
 }
 
-export abstract class Message {
-  readonly message: BaseMessage;
+export abstract class Message<M extends BaseMessage> {
```
diehuxx, line 28: i like this. this is a good idea.

src/dwn.ts
```diff
@@ -113,11 +111,11 @@ export class Dwn {
     console.groupEnd();
 
     console.group('messageStore');
-    await this.messageStore['dump']?.();
+    await this.messageStore['dump']?.(); // FIXME: messageStore.get requires cid and other bits
```
diehuxx, line 114: See my other comment about `dump()`. This is fine as is. We should just ts-ignore

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "strict": true,
```
diehuxx, line 3: To compile with `"strict": true`, we'll need to stricten all the tests too. That's too much to do in this PR, so for now, let's remove this line.

diehuxx (commented):
Last few comments.

src/core/message-reply.ts
```diff
@@ -35,4 +35,14 @@ export class MessageReply {
     this.entries = entries;
     this.data = data;
   }
+
+  static fromError(e: unknown, code: number): MessageReply {
+
+    const isError = (e: unknown): e is { message: string } =>
+      typeof e === 'object' && e !== null && 'message' in e;
```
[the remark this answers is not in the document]
duncanmak: Done.

src/jose/algorithms/signing/secp256k1.ts
duncanmak: Done.

src/store/level-wrapper.ts
duncanmak: Done.

tsconfig.json
duncanmak: Done.

src/store/level-wrapper.ts
```diff
@@ -165,16 +166,18 @@ export class LevelWrapper<V> {
...
-    return abortOr(options?.signal, this.db.batch(operations));
+    return abortOr(options?.signal!, this.db.batch(operations));
   }
 
   private async createLevelDatabase(): Promise<void> {
-    this.db ??= await this.config.createLevelDatabase<V>(this.config.location, {
-      keyEncoding   : 'utf8',
-      valueEncoding : this.config.valueEncoding
-    });
+    if (this.config.createLevelDatabase) {
```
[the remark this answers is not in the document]
duncanmak: Done.

src/interfaces/records/messages/records-read.ts
duncanmak: Done.

src/store/blockstore-level.ts
duncanmak: Done.

duncanmak: Fixing this requires updating the `interface-blockstore` package.

diehuxx (approved):
Added some commits to fix the last few build, test, and merge issues, plus one or two other small things. Looks all good to me at this point. Thanks again @duncanmak !

diehuxx (approved):
fixed one more merge conflict

src/store/data-store-level.ts
diehuxx: Resolving since this can be addressed in a future PR if it becomes important.

diehuxx (approved):

PR #279, opened 2023-03-23
#84 - clearing all blocks for a `dataCid` should remove the data from disk right now

dcrousso:
`leveldb` seems to not immediately delete things from disk when calling `clear` (or `del`)

instead, `leveldb` will periodically remove nonreferenced data when it determines that it needs to (or if it's at an opportune moment, e.g. `open`)

this can be manually tested (we can't do an automated test because `approximateSize` isn't available on all platforms) by doing the following:

1. change any of the `DataStore` tests involving large amounts of data (e.g. `10_000_000`) to be `it.only`
2. repeatedly do `npm run test:node`
3. observe the size of `TEST-DATASTORE` on disk

the size of `TEST-DATASTORE` should not go above ~3× the size of the data (e.g. `30_000_000`)

furthermore, if you then changed that test to not do anything other than `open`, you should see the size of `TEST-DATASTORE` decrease on most runs until it reaches a small enough value that `leveldb` doesn't feel the need to do any compacting

if instead of all that we wanted to delete the data on `clear`, we need to leverage the additional method `compactRange` provided by `classic-level`

in order to make this the nicest API, we only allow this for `sublevel` as that's the easiest way to derive a range to give to `compactRange` (based on the `prefix` of the `sublevel`)

also, for now we're only using `compactRange` inside `clear` as the current structure of `DataStoreLevel` leverages it to delete data (and that's where the large majority of massive data will be)

drive-by: only call `#createLevelDatabase` if the `db` is not yet set to avoid having to `await`

### DIFF
--- a/src/store/level-wrapper.ts
+++ b/src/store/level-wrapper.ts
@@ -81,7 +81,7 @@ export class LevelWrapper<V> {
   async partition(name: string): Promise<LevelWrapper<V>> {
     await this.createLevelDatabase();
 
-    return new LevelWrapper({ ...this.config, location: '' }, this.db.sublevel(name, {
+    return new LevelWrapper(this.config, this.db.sublevel(name, {
       keyEncoding   : 'utf8',
       valueEncoding : this.config.valueEncoding
     }));
@@ -159,7 +159,9 @@ export class LevelWrapper<V> {
   async clear(): Promise<void> {
     await this.createLevelDatabase();
 
-    return this.db.clear();
+    await this.db.clear();
+
+    await this.compactUnderlyingStorage();
   }
 
   async batch(operations: Array<LevelWrapperBatchOperation<V>>, options?: LevelWrapperOptions): Promise<void> {
@@ -168,6 +170,42 @@ export class LevelWrapper<V> {
     await abortOr(options?.signal, this.createLevelDatabase());
 
     return abortOr(options?.signal, this.db.batch(operations));
+  }
+
+  private async compactUnderlyingStorage(options?: LevelWrapperOptions): Promise<void> {
+    options?.signal?.throwIfAborted();
+
+    await abortOr(options?.signal, this.createLevelDatabase());
+
+    const range = this.sublevelRange;
+    if (!range) {
+      return;
+    }
+
+    // additional methods are only available on the root API instance
+    const root = this.root;
+
+    if (root.db.supports.additionalMethods.compactRange) {
+      return abortOr(options?.signal, root.db['compactRange']?.(...range));
+    }
+  }
+
+  private get sublevelRange(): [ string, string ] | undefined {
+    const prefix = this.db['prefix'];
+    if (!prefix) {
+      return undefined;
+    }
+
+    // use the separator to derive an exclusive `end` that will never match to a key (which matches how `abstract-level` creates a `boundary`)
+    return [ prefix, prefix.slice(0, -1) + String.fromCharCode(prefix.charCodeAt(prefix.length - 1) + 1) ];
+  }
+
+  private get root(): LevelWrapper<V> {
+    let db = this.db;
+    for (const parent = db['db']; parent && parent !== db; ) {
+      db = parent;
+    }
+    return new LevelWrapper(this.config, db);
   }
 
   private async createLevelDatabase(): Promise<void> {
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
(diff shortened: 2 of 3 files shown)


src/store/level-wrapper.ts
```diff
@@ -170,6 +172,42 @@ export class LevelWrapper<V> {
...
+    // additional methods are only available on the root API instance
+    const root = this.root;
+
+    if (root.db.supports.additionalMethods.compactRange) {
+      return abortOr(options?.signal, root.db['compactRange']?.(...range));
+    }
+  }
+
+  private get sublevelRange(): [string, string] | undefined {
+    const prefix = this.db['prefix'] ?? '';
```
thehenrytsai, line 194: For my learning/understanding: is `prefix` an implementation specific detail of sublevel?

src/store/level-wrapper.ts
```diff
@@ -170,6 +172,42 @@ export class LevelWrapper<V> {
...
+    if (!range) {
+      return;
+    }
+
+    // additional methods are only available on the root API instance
+    const root = this.root;
+
+    if (root.db.supports.additionalMethods.compactRange) {
+      return abortOr(options?.signal, root.db['compactRange']?.(...range));
+    }
```
thehenrytsai, line 190: For my understanding, would the below not be sufficient (without the `if` check)? Are you trying to draw attention to the fact that `compactRange()` may not be available thinking that the `?.` is not explicit enough? No push back if that's the case, just trying to understand the intent.
```suggestion
    return abortOr(options?.signal, root.db['compactRange']?.(...range));
```

src/store/level-wrapper.ts
```diff
@@ -170,6 +172,42 @@ export class LevelWrapper<V> {
...
+      return undefined;
+    }
+
+    // use the separator to derive an exclusive `end` that will never match to a key (which matches how `abstract-level` creates a `boundary`)
+    return [ prefix, prefix.slice(0, -1) + String.fromCharCode(prefix.charCodeAt(prefix.length - 1) + 1) ];
+  }
+
+  private get root(): LevelWrapper<V> {
+    let db = this.db;
+    for (const parent = db['db']; parent && parent !== db; ) {
```
thehenrytsai, line 205: The `parent && parent !== db` throws me off... could `parent !== db` alone be enough?

Can you annotate in code why this check is the way it is? I can imagine why it "could" be needed, like:

- `db['db']` gets you the parent if there is a parent, which is kinda odd already, but fair enough I guess...
- when there is no parent, the `db['db']` can sometimes be `undefined` or sometimes be the same as the child db (this is the odd part for me)

without explicit comment, the code gives appearance of unnecessary check or potential bug.

thehenrytsai (commented):
A couple of comments and questions.

src/store/level-wrapper.ts
dcrousso: i was trying to follow the API design of `level` in that you should check for `supports` before doing the thing :)

src/store/level-wrapper.ts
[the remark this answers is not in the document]
dcrousso: ah yeah sure i guess i check it below anyways so no need to `?? ''`

dcrousso: yeah `prefix` is only a thing for `Sublevel`

src/store/level-wrapper.ts
dcrousso: the interface of `AbstractSublevel` is where the `db` is, but some implementations of `AbstractLevel` will also provide `db` as a backwards compatibility thing

the `parent &&` is really just a defensive check since we dont want to try to create a `LevelWrapper` with a `db` that is `undefined`

thehenrytsai: Repeating back what I understood, let me know if I got anything wrong:

- `db['db']` get you the level DB, it's the same instance of the level DB regardless of which "layer" of sublevel you are at
- at the root level (not sublevel), there is no guarantee that `db[`'db']` would return an instance (since it's an `AbstractSublevel` prop.

dcrousso: i _think_ that `db['db']` will always get you the root `level`, but i'm not entirely sure so i wrote that code just in case `db['db']` only returns the parent.  e.g.
```
let db = new Level(...);
let a = db.sublevel('a');
let b = a.sublevel('b');
let c = b.sublevel('c');

console.log(c.db)  // should be `db`, but could be `b`
console.log(b.db)  // should be `db`, but could be `a`
console.log(a.db)  // should be `db`
console.log(db.db) // should be `undefined`, but could be `db` (see below)
```

according to `AbstractLevel` it shouldn't have a `db` property, but some implementationsprovide it for backwards compat reasons (e.g. [`browser-level`](https://github.com/Level/browser-level/blob/main/index.js#L67-L70))

TL;DR: i did this to be extra defensive against known scenarios where things aren't as described by the "types"

thehenrytsai (approved):

PR #280, opened 2023-03-25
add a way to do chunked file uploading

dcrousso:
rather than try to shove everything into a single `RecordsWrite`, allow for clients to send data in multiple parts that are assembled on the receiving end

the following new messages are added:

- `RecordsUploadStart` indicates that an upload is now beginning
- `RecordsUploadPart` contains a chunk of data and an `index` indicating where in the overall whole it belonds
- `RecordsUploadComplete` gives a final `count` of the number of `RecordsUploadPart` and the `dataFormat` it should use

with these new messages, there are new expectations:

- all `RecordsUpload*` for the same upload must have the same `recordId`
- there can never be more than one `RecordsUploadStart` for a `recordId`
- each `RecordsUploadPart` must provide a `dataCid` and `dataSize` corresponding to the `dataStream`
- each `index` must be unique across all `RecordsUploadPart`
- there can not be any `RecordsUpload*` after a `RecordsUploadComplete` for a `recordId`
- the `RecordsUploadComplete` must provide a `dataCid` and `dataSize` computed by combining all of the `RecordsUploadPart`
- any extra `index` beyond the `count` are deleted when handling the `RecordsUploadComplete`
- `RecordsWrite` cannot use the same `recordId` as `RecordsUpload*`

the default storage mechanism for `RecordsUpload*` is `UploadStoreLevel`, which has the following structure (`+` represents a sublevel and `->` represents a key->value pair):

```
  '<tenant>' + <recordId> + 'data' + <index> -> <data>
  '<tenant>' + <recordId> + 'root' + <index> -> <dataCid>
```

this allows for the `<data>` to be stored without having to provide it's `<dataCid>` upfront, which is especially necessary when getting _all_ the data for a `<recordId>` for a `<tenant>`, as the root `<dataCid>` can be retrieved for each `<index>` internally (i.e. not needing the client to provide each `<dataCid>` for each part).

### DIFF
--- a/src/interfaces/records/handlers/records-upload.ts
+++ b/src/interfaces/records/handlers/records-upload.ts
@@ -1,0 +1,208 @@
+import type { MethodHandler } from '../../types.js';
+import type { DidResolver, MessageStore, UploadStore } from '../../../index.js';
+import type { RecordsUploadCompleteMessage, RecordsUploadPartMessage, RecordsUploadStartMessage } from '../types.js';
+
+import { authenticate } from '../../../core/auth.js';
+import { MessageReply } from '../../../core/message-reply.js';
+import { RecordsUpload } from '../messages/records-upload.js';
+import { DwnInterfaceName, DwnMethodName, DwnStateName } from '../../../core/message.js';
+
+type RecordsUploadMessageVariant = RecordsUploadCompleteMessage | RecordsUploadPartMessage | RecordsUploadStartMessage;
+
+export class RecordsUploadHandler implements MethodHandler {
+
+  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private uploadStore: UploadStore) { }
+
+  public async handle({
+    tenant,
+    message,
+    dataStream
+  }): Promise<MessageReply> {
+    const incomingMessage = message as RecordsUploadMessageVariant;
+
+    let recordsUpload: RecordsUpload;
+    try {
+      recordsUpload = await RecordsUpload.parse(incomingMessage);
+    } catch (e) {
+      return new MessageReply({
+        status: { code: 400, detail: e.message }
+      });
+    }
+
+    // authentication & authorization
+    try {
+      await authenticate(incomingMessage.authorization, this.didResolver);
+      await recordsUpload.authorize(tenant, this.messageStore);
+    } catch (e) {
+      return new MessageReply({
+        status: { code: 401, detail: e.message }
+      });
+    }
+
+    // get existing upload messages matching the `recordId`
+    const query = {
+      interface : DwnInterfaceName.Records,
+      method    : DwnMethodName.Upload,
+      recordId  : incomingMessage.recordId
+    };
+    const existingMessages = await this.messageStore.query(tenant, query) as RecordsUploadMessageVariant[];
+
+    try {
+      const existingUploadComplete = await RecordsUpload.getUploadComplete(existingMessages);
+      if (existingUploadComplete) {
+        return new MessageReply({
+          status: { code: 400, detail: 'upload already complete' }
+        });
+      }
+    } catch (e) {
+      // ignore error if upload complete cannot be found
+    }
+
+    const newMessageIsUploadStart = await recordsUpload.isUploadStart();
+
+    try {
+      const existingUploadStart = await RecordsUpload.getUploadStart(existingMessages);
+      if (existingUploadStart) {
+        if (newMessageIsUploadStart) {
+          return new MessageReply({
+            status: { code: 400, detail: 'cannot start an upload more than once' }
+          });
+        }
+
+        RecordsUpload.verifyEqualityOfImmutableProperties(existingUploadStart, incomingMessage);
+      }
+    } catch (e) {
+      // it's ok to not find the upload start if the new message is the upload start
+      if (!newMessageIsUploadStart) {
+        return new MessageReply({
+          status: { code: 400, detail: e.message }
+        });
+      }
+    }
+
+    try {
+      for (const existingMessage of existingMessages) {
+        RecordsUpload.verifyExclusivityOfUniqueProperties(existingMessage, incomingMessage);
+      }
+    } catch (e) {
+      return new MessageReply({
+        status: { code: 400, detail: e.message }
+      });
+    }
+
+    const newMessageIsUploadComplete = await recordsUpload.isUploadComplete();
+    if (newMessageIsUploadComplete) {
+      const uploadComplete = incomingMessage as RecordsUploadCompleteMessage;
+
+      const indexes = new Set<number>();
+
+      for (const existingMessage of existingMessages) {
+        if (existingMessage.descriptor.state === DwnStateName.Part) {
+          const uploadPart = existingMessage as RecordsUploadPartMessage;
+
+          // we already checked above that each index is unique
+          indexes.add(uploadPart.descriptor.index);
+        }
+      }
+
+      // make sure we have all the parts
+      for (let index = 0; index < uploadComplete.descriptor.count; ++index) {
+        if (!indexes.has(index)) {
+          return new MessageReply({
+            status: { code: 400, detail: `missing index '${index}'` }
+          });
+        }
+
+        indexes.delete(index);
+      }
+
+      // make sure we don't have any extra parts
+      for (const index of indexes) {
+        return new MessageReply({
+          status: { code: 400, detail: `extra index '${index}'` }
+        });
+      }
+    }
+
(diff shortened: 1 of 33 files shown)


src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+import { RecordsUpload } from '../messages/records-upload.js';
+import { DwnInterfaceName, DwnMethodName, DwnStateName } from '../../../core/message.js';
+
+type RecordsUploadMessageVariant = RecordsUploadCompleteMessage | RecordsUploadPartMessage | RecordsUploadStartMessage;
+
+export class RecordsUploadHandler implements MethodHandler {
+
+  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private uploadStore: UploadStore) { }
+
+  public async handle({
```
dcrousso, line 16: havent written any tests for this yet.  wanted to get the PR up so folks could give feedback on the API :)

src/interfaces/records/messages/records-upload.ts
```diff
@@ -0,0 +1,516 @@
...
+};
+
+export type RecordsUploadCompleteOptions = RecordsUploadOptions & {
+  recordId: string;
+  count: number;
+  dataCid: string;
+  dataSize: number;
+};
+
+export class RecordsUpload extends Message {
```
dcrousso, line 49: havent written any tests for this yet.  wanted to get the PR up so folks could give feedback on the API :)

src/interfaces/records/messages/records-upload.ts
```diff
@@ -0,0 +1,516 @@
...
+   * Gets the upload complete from the given list or record upload.
+   */
+  public static async getUploadComplete(messages: BaseMessage[]): Promise<RecordsUploadCompleteMessage>{
+    for (const message of messages) {
+      if (await RecordsUpload.isUploadComplete(message)) {
+        return message as RecordsUploadCompleteMessage;
+      }
+    }
+
+    throw new Error(`upload complete is not found`);
```
diehuxx, line 477: Strikes me as a bit odd to throw on what feels like valid input in an unexceptional situation. Esp since the only usage of this function swallows the error. Thoughts on returning `undefined`?

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    const newMessageIsUploadStart = await recordsUpload.isUploadStart();
+    const newMessageIsUploadComplete = await recordsUpload.isUploadComplete();
+
+    const parts = new Set<number>();
+
+    try {
+      const existingUploadStart = await RecordsUpload.getUploadStart(existingMessages);
+      if (existingUploadStart) {
```
diehuxx, line 68: `getUploadStart` throws when there is no `existingUploadStart`, so this won't address the valid case where `newMessageIsUploadStart` and there is no `existingUploadStart`.

I mentioned in another comment above that it's a bit odd that the fact that `getUploadComplete` throws is odd to me. Same goes for `getUploadStart`. Personal preference for returning `undefined`.

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+type RecordsUploadMessageVariant = RecordsUploadCompleteMessage | RecordsUploadPartMessage | RecordsUploadStartMessage;
+
+export class RecordsUploadHandler implements MethodHandler {
+
+  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private uploadStore: UploadStore) { }
+
+  public async handle({
+    tenant,
+    message,
+    dataStream
```
diehuxx, line 19: nit: add types pls

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+        if (!parts.has(index)) {
+          return new MessageReply({
+            status: { code: 400, detail: `missing index '${index}'` }
+          });
+        }
+
+        parts.delete(index);
+      }
+
+      for (const index of parts) {
```
diehuxx, line 108: having trouble parsing what this does. `index` is defined in the outer loop. but also again here. Is that intentional? If so, it might be a little too clever for me.

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+    } catch (e) {
+      // ignore error if upload complete cannot be found
+    }
+
+    const newMessageIsUploadStart = await recordsUpload.isUploadStart();
+    const newMessageIsUploadComplete = await recordsUpload.isUploadComplete();
+
+    const parts = new Set<number>();
+
+    try {
```
diehuxx, line 66: Around this `try/catch` this method gets a little hard to follow for me, with many `if` checking what RecordsUpload state the message is. What do you think of refactoring with `handleStart`, `handlePart`, and `handleComplete`?

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+          }
+        }
+      }
+    } catch (e) {
+      return new MessageReply({
+        status: { code: 400, detail: e.message }
+      });
+    }
+
+    if (newMessageIsUploadComplete) {
```
diehuxx, line 95: We check that we have unique indexed parts up to `uploadComplete.descriptor.count`. We do not do this index check when we are handling a `RecordsUploadPart`. I'm guessing the reason is that we may receive parts out of order? If so, is it also possible that we may receive the `Complete` message before we finish receiving the `Part`s?

src/interfaces/records/handlers/records-upload.ts
```diff
@@ -0,0 +1,196 @@
...
+      result = await this.uploadStore.part(tenant, uploadPart.recordId, uploadPart.descriptor.index, dataStream);
+      break;
+
+    case DwnStateName.Start:
+      var uploadStart = incomingMessage as RecordsUploadStartMessage;
+      result = await this.uploadStore.start(tenant, uploadStart.recordId, uploadStart.descriptor.dataFormat);
+      break;
+    }
+
+    if (!result) {
```
diehuxx, line 134: Could this go as a `default` case in the `switch` statement?

diehuxx (commented):
Thinking through updating/overwriting a record. I'm still not confident in all my knowledge here, so please do correct me if my assumptions are wrong.

In the current design:
* The first `RecordsWrite` for a given `recordId` creates the record in the receiving DWN. That `recordId` is deterministic based on (among other things) the `CID` of the data.
* Each subsequent `RecordsWrite` with the same `recordId` will overwrite the record, and the data store will keep the same `recordId` even though the underlying CID has changed.

With `RecordsUpload` how do we update the record? My guess would be that a record can be updated only after the `Start`+`Parts`+`Complete` cycle has completed. At which point, a new cycle may be initiated for a given `recordId`. This seems to be prohibited based your notes in the PR description and the code itself, which allows for at most one `Start`+`Complete` per `recordId`.

src/store/upload-store-level.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  async open(): Promise<void> {
+    await this.blockstore.open();
+  }
+
+  async close(): Promise<void> {
+    await this.blockstore.close();
+  }
+
+  async start(_tenant: string, _recordId: string, _dataFormat: string, options?: UploadStoreOptions): Promise<boolean> {
```
diehuxx, line 58: What is the use case for having `start()` at all if it's no-op? I see that a developer may want to supply their own `upload-store` with a custom `start()`, but what would they use it for? If there are some ideas you have in mind, I'm down to keep this, but otherwise I'm partial to omitting it until that need pops up.

src/interfaces/records/handlers/records-upload.ts
dcrousso: no.  `result` is checking that the `UploadStore` operation did what was expected (e.g. the `dataCid` and `dataSize` match), not that we're handling some unknown `DwnStateName` value

thehenrytsai:
Would be nice it there was a short technical proposal document to go over the motivation, design, tradeoffs etc first. But please consult with the TPMs (@csuwildcat and @frankhinek) first before even consider doing a TP so that you don't waste your time as I am not aware that they want us to introduce new DWN methods right now. I'll pause on reviewing the PR unless they give green light on the approach.

diehuxx:
Closing this, since it's now pretty far behind main and we haven't gotten around to a tech proposal. If we revisit, probably easiest to open a fresh PR.

Issue #276, opened 2023-03-22
Support data encryption


thehenrytsai:
Tracking proposal https://github.com/TBD54566975/technical-proposals/pull/4

thehenrytsai:
Encryption is now supported, will create separate issue tracking specific encryption features, such as specific key derivation schemes.

PR #282, opened 2023-03-29
#276 - added methods for AES encryption and decryption

thehenrytsai:


csuwildcat:
Seems straightforward, lgtm

diehuxx (approved):
LGTM. IIRC you said this was working in browser but `crypto` shouldn't work in browser. Did we figure out what's the deal there? But hey if it works, works for me.

PR #283, opened 2023-03-29
Support author protocols rule

diehuxx:
Addresses #241. `recipient` is already implemented. This PR adds support for `author`


src/core/protocol-authorization.ts
```diff
@@ -155,43 +158,17 @@ export class ProtocolAuthorization {
   }
 
   /**
-   * Verifies the requester of the given message is allowed actions based on the rule set.
-   * @throws {Error} if requester not allowed.
+   * Verifies the actions specified in the given message matches the allowed actions in the rule set.
+   * @throws {Error} if action not allowed.
    */
-  private static verifyAllowedRequester(
```
thehenrytsai, line 161: For my understanding: what's the reason that prompted the decision to merge the logic into `verifyAllowedActions()`? Was it because it's hard to understand? Or redundant code? Or?

The intention behind having two separate methods here were:
1. It is easier to know what the method is doing (in theory) without reading the implementation.
2. It gives precise error message when error does occur.

The new implementation removes both benefits above, hence want to discuss it.

json-schemas/protocol-rule-set.json
```diff
@@ -32,6 +32,35 @@
...
+                "to": {
+                  "type": "array",
+                  "minItems": 1,
+                  "items": {
+                    "type": "string",
+                    "enum": [
+                      "write"
+                    ]
+                  }
+                }
```
thehenrytsai, line 55: Consider filing a good-first-issue for the community to refactor the repeated definition into its own schema so we can just reference the same schema.

thehenrytsai (commented):
Thanks @diehuxx, looks good overall!

Have a design question for discussion. Will wait for response before reviewing further!

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -1,6 +1,8 @@
 import chaiAsPromised from 'chai-as-promised';
 import credentialIssuanceProtocolDefinition from '../../../vectors/protocol-definitions/credential-issuance.json' assert { type: 'json' };
 import dexProtocolDefinition from '../../../vectors/protocol-definitions/dex.json' assert { type: 'json' };
+import socialMebiaProtocolDefinition from '../../../vectors/protocol-definitions/social-mebia.json' assert { type: 'json' };
+
```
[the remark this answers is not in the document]
diehuxx: Yes, I was being a goofball since it's just a test fixture. I'm down to correct the misspelling if it's confusing/annoying/too silly.

src/core/protocol-authorization.ts
diehuxx: I tried keeping `verifyAllowedRequester` and `verifyAllowedActions` separate, but I found that they were becoming redundant. Plus, I felt that the implementation was more intuitive by combining them. I'm down to jam on keeping them separate if you feel that's best. Here was my thought process:

We had these lines originally:
```typescript
    if (allowRule.anyone !== undefined) {
      allowedActions = allowRule.anyone.to;
    } else if (allowRule.recipient !== undefined) {
      allowedActions = allowRule.recipient.to;
    }
    const inboundMessageAction = methodToAllowedActionMap[incomingMessageMethod];
    if (!allowedActions.includes(inboundMessageAction)) {
      throw new Error(`inbound message action '${inboundMessageAction}' not in list of allowed actions ${allowedActions}`);
    }
```

Consider an `allowRule` with both `anyone` and `recipients` defined. It's possible that `recipients` have more allowed actions than `anyone`, and that the `inboundMessageAction` is in `recipient.to`. The above code would erroneously block authorization because it would hit the `if (allowRule.anyone !== undefined)`.

We could fix this by turning the `else if` into a regular `if`, and adding allowed rules into a set of `allowedActions`. But it gets odd when you add `author`.
```typescript
    if (allowRule.anyone !== undefined) {
      allowRule.anyone.to.forEach(action => allowedActions.add(action));
    }
    if (allowRule.recipient !== undefined) {
      allowRule.recipient.to.forEach(action => allowedActions.add(action));
    }
    if (allowRule.author !== undefined) {
      allowRule.author.to.forEach(action => allowedActions.add(action));
    }
```

This is bad because we don't want to allow all the actions in all the rules. We need to check that the message is actually from an `author` or a `recipient`.

```typescript
    if (allowRule.anyone !== undefined) {
      allowRule.anyone.to.forEach(action => allowedActions.add(action));
    }
    if (allowRule.recipient !== undefined) {
      if message is from a recipient {
        allowRule.recipient.to.forEach(action => allowedActions.add(action));
      }
    }
    if (allowRule.author !== undefined) {
      if message is from an author {
        allowRule.recipient.to.forEach(action => allowedActions.add(action));
      }
    }
```
We already do those checks in `verifyAllowedRequester`, so `verifyAllowedRequester` is redundant.

json-schemas/protocol-rule-set.json
diehuxx: Good call. #286

src/core/protocol-authorization.ts
thehenrytsai: Thanks @diehuxx for the full context, I did not at all consider the possibility of having multiple "roles" under the same `allow` property! I will continue with the review with this new knowledge!

tests/interfaces/records/handlers/records-write.spec.ts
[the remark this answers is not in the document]
thehenrytsai: I have spell checker on so the one dozen occurrences stick out like a sore thumb, but knowing this is intentional I'll let you decide if you want to keep the fun or not!

Don't be surprised if someone in the community submits a PR to fix the "typos"!

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -570,6 +572,60 @@ describe('RecordsWriteHandler.handle()', () => {
           .to.equal(base64url.baseEncode(encodedCredentialResponse));
       });
 
+      it('should allow write with author rule', async () => {
+        // scenario: alice posts an image on the social mebia protocol, then she adds a caption
+        const protocol = 'https://tbd.website/decentralized-web-node/protocols/social-mebia';
+        const protocolDefinition: ProtocolDefinition = socialMebiaProtocolDefinition;
+
+        const alice = await TestDataGenerator.generatePersona();
```
thehenrytsai, line 580: The test is not valid because Alice is writing to her own DWN so even without the `author` rule, the test will pass.

So you'll need have to involve Bob somehow in the scenario.

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -813,6 +869,53 @@ describe('RecordsWriteHandler.handle()', () => {
...
+          schema    : socialMebiaProtocolDefinition.labels.image.schema,
+          data      : encodedImage
+        });
+        const imageReply = await dwn.processMessage(alice.did, imageRecordsWrite.message, imageRecordsWrite.dataStream);
+        expect(imageReply.status.code).to.equal(202);
+
+        const encodedCaption = new TextEncoder().encode('bad vibes! >:(');
+        const imageContextId = await imageRecordsWrite.recordsWrite.getEntryId();
+        const captionRecordsWrite = await TestDataGenerator.generateRecordsWrite({
+          requester : aliceImposter,
```
thehenrytsai, line 907: Consider merge this the negative test with the positive test above, tagging on the imposter test at the end, as the test is so verbosely similar.

diehuxx: Merged

tests/interfaces/records/handlers/records-write.spec.ts
diehuxx: Fixed

tests/interfaces/records/handlers/records-write.spec.ts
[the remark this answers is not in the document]
diehuxx: I switched to the correct spelling. I've had my fun, that's enough for today :)

thehenrytsai (approved):
Good to go!!! 🙏 🎖️

Issue #14, opened 2022-03-11
Implement Sync


mistermoe:
related proposal issue: decentralized-identity/identity-hub#128

mistermoe:
# DWN Sync Proposal

Synchronizing a tenant's data across more than 1 DWN can be achieved by introducing three mechanisms:
* An append-only event log
* The ability to get events from event log
* the ability to get a message by CID

using the above mechanisms in addition to existing ones we can break sync into 2 phases:
* **Push**: here are all the messages i have for you
* **Pull**: give me all the messages i haven't already gotten from you

## Event Log
Within the context of a DWN, An append-only event log is simply a ordered K/V store where the key is an ordinal number and the value is the CID of a message. 

Each DWN would maintain it's own individual event log. There's no requirement for an individual DWN's event log to be in the same order as another DWN

Whenever a message is stored by a DWN, the CID of that message is added to its event log.

## `EventsGet`

message type used to get events from a DWN

### Message Contents

| property name   | type              | required (y/n) | description                                                        |
| --------------- | ----------------- | -------------- | ------------------------------------------------------------------ |
| `method`        | const `EventsGet` | Y              |                                                                    |
| `watermark`     | string            | N              | providing no watermark returns all events for a given tenant       |
| `authorization` | GeneralJws        | Y              | same expected authorization format required by other DWN messages. |


### Response
all events that occurred between the watermark and the time at which the message is processed
[`MessageReply`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/core/message-reply.ts#L13) where `entries` is an array of event objects

### Notes
* Should not be usable in protocol definitions


## `MessagesGet`

message typed used to get a message by CID from a DWN

### Message Contents

| property name   | type                | required (y/n) | description                                                        |
| --------------- | ------------------- | -------------- | ------------------------------------------------------------------ |
| `method`        | const `MessagesGet` | Y              |                                                                    |
| `cid`           | string              | Y              | the CID of the message you want                                    | 
| `authorization` | GeneralJws          | Y              | same expected authorization format required by other DWN messages. |

## Pull
Introducing `EventsGet` and `MessagesGet` makes it possible to pull all messages from a given DWN

```mermaid
sequenceDiagram
autonumber

participant A as DWN A
participant B as DWN B

A->>A: grab pull-specific watermark for DWN B
A->>B: EventsGet{ watermark, tenant_sig }
B->>B: fetch all events between watermark and now
[...]

PR #290, opened 2023-03-31
Event log

mistermoe:
First of 3 PRs to implement sync. Introduces an `EventLog` interface along with a concrete level-based implementation. Events are added to the EventLog for:
* `RecordsWrite`
* `RecordsDelete`
* `ProtocolsConfigure`

The PR that follows this one will introduce the `EventsGet` interface method


src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+  location: string,
+  createLevelDatabase?: typeof createLevelDatabase,
+};
+
+export class EventLogLevel implements EventLog {
+  config: EventLogLevelConfig;
+  level: LevelWrapper<string>;
+  ulid: ULID;
+
+  constructor(config: EventLogLevelConfig = { location: 'EVENTLOG' }) {
```
dcrousso, line 25: NIT: i feel like we should put this as
```
this.config = {
  location: 'EVENTLOG',
  createLevelDatabase,
  ...config
};
```
as that'll handle the scenario where `{ }` is provided to `new EventLogLevel` (since default values for properties only get applied when the property is not provided or `undefined`)

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+      createLevelDatabase,
+      ...config,
+    };
+
+    this.level = new LevelWrapper<string>({ ...this.config, valueEncoding: 'utf8' });
+    this.ulid = monotonicFactory();
+  }
+
+  async open(): Promise<void> {
+    // `db.open()` is automatically called by the database constructor.  We're calling it explicitly
```
dcrousso, line 36: is this needed?  we already have this comment in `IndexLevel`

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+    // in order to explicitly catch an error that would otherwise not surface until another method
+    // like `db.get()` is called.  Once `db.open()` has then been called, any read & write
+    // operations will again be queued internally until opening has finished.
+    return this.level.open();
+  }
+
+  close(): Promise<void> {
```
dcrousso, line 43: no `async`?

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+    return this.level.close();
+  }
+
+  clear(): Promise<void> {
```
dcrousso, line 47: ditto (:43)

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+   *  LevelDB will store its files, or in browsers, the name of the
+   * {@link https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase IDBDatabase} to be opened.
+  */
+  location: string,
+  createLevelDatabase?: typeof createLevelDatabase,
+};
+
+export class EventLogLevel implements EventLog {
+  config: EventLogLevelConfig;
+  level: LevelWrapper<string>;
```
dcrousso, line 22: to be consistent i think we should call this `db` (or rename all the others)

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+      events.push(event);
+    }
+
+    return events;
+  }
+
+  hashTenant(tenant: string): string {
+    const tenantBytes = Encoder.stringToBytes(tenant);
+
+    return base32crockford.encode(tenantBytes);
```
dcrousso, line 78: why do we need to hash the `tenant`?  isn't it already a DID/CID?

if this is desired/necessary, should we be doing this for all the other places too (e.g. `MessageStoreLevel`)?

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+  async append(tenant: string, messageCid: string): Promise<string> {
+    const hashedTenant = this.hashTenant(tenant);
+    const tenantEventLog = await this.level.partition(hashedTenant);
+
+    const watermark = this.ulid();
+    await tenantEventLog.put(watermark, messageCid);
+
+    return watermark;
+  }
+  async getEventsAfter(tenant: string, watermark?: string | undefined): Promise<Event[]> {
```
dcrousso, line 60: is the `| undefined` necessary since you've already marked it as optional?

Style: missing newline before

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const opts = watermark ? { gt: watermark } : {};
+    const events: Array<Event> = [];
+
+    for await (const [key, value] of tenantEventLog.iterator(opts)) {
+      const event = { watermark: key, messageCid: value };
+      events.push(event);
+    }
+
+    return events;
```
dcrousso, line 72: NIT: if you wanted to you could instead modify `asyncGeneratorToArray` to accept a `mapper` parameter that does
```
array.push(mapper ? mapper(value) : value);
```
(which would match the `Array.fromAsync` proposal too)

src/event-log/event-log.ts
```diff
@@ -0,0 +1,32 @@
...
+  append(tenant: string, messageCid: string): Promise<string>
+
+  /**
+   * retrieves all of a tenant's events that occurred after the watermark provided.
+   * If no watermark is provided, all events for a given tenant will be returned.
+   *
+   * @param tenant
+   * @param watermark
+   */
+  getEventsAfter(tenant: string, watermark?: string): Promise<Array<Event>>
```
dcrousso, line 31: NIT: what about just `get`?  the `After` part is a bit misleading IMO in the case where someone just calls `getEventsAfter(tenant)` since there is no `watermark` argument

src/store/storage-controller.ts
```diff
@@ -78,6 +80,9 @@ export class StorageController {
     }
 
     await messageStore.put(tenant, message, indexes);
+
+    const messageCid = await Message.getCid(message);
```
dcrousso, line 84: we already calculate this above, but only `if (message.descriptor.dataCid !== undefined)`

could we instead move that outside the `if` and avoid the repeated calculation?

tests/event-log/event-log-level.spec.ts
```diff
@@ -0,0 +1,130 @@
+import chaiAsPromised from 'chai-as-promised';
+import { computeCid } from '../../src/utils/cid.js';
+import { EventLogLevel } from '../../src/event-log/event-log-level.js';
+import { TestDataGenerator as TDG } from '../utils/test-data-generator.js';
```
dcrousso, line 4: NIT: as much as I appreciate why you did this, I personally try to avoid renaming things like this as it makes it harder to do a search (e.g. `grep`) for things in a codebase 😅

src/event-log/event-log.ts
```diff
@@ -0,0 +1,32 @@
...
+   * closes the connection to the underlying store
+   */
+  close(): Promise<void>;
+
+  /**
+   * adds an event to a tenant's event log
+   * @param tenant - the tenant's DID
+   * @param messageCid - the CID of the message
+   */
+  append(tenant: string, messageCid: string): Promise<string>
```
dcrousso, line 22: i wonder if we should have this take an actual `BaseMessage` instead of a `messageCid` to ensure that we're actually getting the entire message and not just some random CID (similar to how `MessageStore` requires the entire `BaseMessage` for `put`)

also, how does this handle stuff like `deleteAllOlderMessagesButKeepInitialWrite` where we remove any overridden `RecordsWrite` (except for the initial message and the newest one)?  all the intervening messages will no longer exist in the `MessageStore`, so having entries for them in the `EventLog` seems a bit odd (and possibly problematic if a new `RecordsWrite` is received in between an `append` and `getEventsAfter` as the other `Dwn` would have a `messageCid` that no longer exists in the `MessageStore` since it would've been deleted by that new `RecordsWrite`)

diehuxx (commented):
Codewise this looks generally good to me, with a couple small questions. But I have some broader design questions before I stamp.

Let me know if you'd prefer design discussion here or in [your tech proposal](https://github.com/TBD54566975/dwn-sdk-js/issues/14).

Trying to think through this scenario.
1. Say that DWN A and B both have record R. Initially they have the same most recent update of record R.
2. Then DWN A receives a RecordsWrite(A) updating R to be R(A). Slightly after that, before any sync occurs, DWN B also receives a RecordsWrite(B) updating R to be R(B).
3. Is it deterministic which version of R wins? If not, is that acceptable?

If DWN A initiates a pull from B, then it will send the CID of RecordsWrite(A) to DWN B in an EventsGet. DWN B has never seen this message, so I’m actually not sure what it will return, since it’s supposed to return events after the watermark. Does it return no events? Or all events from all time?

src/dwn.ts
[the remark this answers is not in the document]
csuwildcat: Because the system and data model are inherently convergent with a deterministic outcome for all record activities, sync shouldn't have to care, just simply ensure events make it to the other node, where processing them will naturally settle to a convergent state.

src/event-log/event-log-level.ts
mistermoe: adjusted

src/event-log/event-log-level.ts
mistermoe: na. was just a note to myself while implementing the event log. removed it. feel free to resolve after taking a look.

src/event-log/event-log-level.ts
mistermoe: i typically don't add `async` to a method signature unless i'm explicitly `await`ing in the method body. went ahead and added it as i'm guessing `async` appears in other places in the codebase whenever a `Promise` is returned. feel free to resolve after taking a look

src/event-log/event-log-level.ts
mistermoe: i typically don't add `async` to a method signature unless i'm explicitly `await`ing in the method body. went ahead and added it as i'm guessing `async` appears in other places in the codebase whenever a `Promise` is returned. feel free to resolve after taking a look

src/event-log/event-log-level.ts
mistermoe: changed. feel free to resolve after taking a look

src/event-log/event-log-level.ts
mistermoe: it isn't necessary. removed it. feel free to resolve comment after taking a look

src/event-log/event-log.ts
mistermoe: > the `After` part is a bit misleading IMO in the case where someone just calls

fair. though i think `get` is misleading as well because it i'm not sure whether i'm getting 1 or many. i suppose one could make the argument that the return signature would provide you with that detail.

tests/event-log/event-log-level.spec.ts
mistermoe: good point! changed.

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  async append(tenant: string, messageCid: string): Promise<string> {
+    const hashedTenant = this.hashTenant(tenant);
+    const tenantEventLog = await this.level.partition(hashedTenant);
+
+    const watermark = this.ulid();
+    await tenantEventLog.put(watermark, messageCid);
+
+    return watermark;
```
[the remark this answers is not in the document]
mistermoe: In the context of event streams, a watermark is a technique used to track the progress of events within a stream and to manage event processing.

In event processing, events are typically processed in batches rather than one at a time. A watermark is a timestamp that represents the last event in a batch of events that have been processed. As new events are added to the stream, the watermark is updated to represent the most recent processed event.

The purpose of a watermark is to provide a way to manage the progress of event processing, to prevent event duplication, and to provide a guarantee of event ordering. By using a watermark, event processing systems can ensure that they have processed all events up to a certain point in time before moving on to process newer events.

down to change the term/word to something else if you've got something in mind!

src/event-log/event-log-level.ts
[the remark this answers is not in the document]
mistermoe: good call. changed. feel free to resolve comment after taking a look

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,80 @@
...
+  close(): Promise<void> {
+    return this.level.close();
+  }
+
+  clear(): Promise<void> {
+    return this.level.clear();
+  }
+
+  async append(tenant: string, messageCid: string): Promise<string> {
+    const hashedTenant = this.hashTenant(tenant);
```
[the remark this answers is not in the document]
mistermoe: `hash` isn't even the correct terminology here so apologies for that. currently i'm encoding only. i'll update the method name to `encode`. but to answer your question, i had two specific reasons in mind:
1. DIDs can be very long and in general can be different lengths (ion short form vs. long form). `base32crockford` produces a fixed-length output, which makes it easier to handle and compare than the variable-length
2. `base32crockford` encoding also produces output that consists of a limited set of alphanumeric characters, making it more suitable for use in contexts where special characters or binary data are not allowed or need to be avoided. Specifically in the case of `level`, only ascii characters are allowed when creating sub-levels. @dcrousso ran into this and has an [issue open](https://github.com/Level/abstract-level/issues/70) in the `abstract-level` repo

src/event-log/event-log-level.ts
mistermoe: `hash` isn't even the correct terminology here so apologies for that. currently i'm encoding only. i'll update the method name to `encode`. but to answer your question, i had two specific reasons in mind:
1. DIDs can be very long and in general can be different lengths (ion short form vs. long form). `base32crockford` produces a fixed-length output, which makes it easier to handle and compare than the variable-length
2. `base32crockford` encoding also produces output that consists of a limited set of alphanumeric characters, making it more suitable for use in contexts where special characters or binary data are not allowed or need to be avoided. Specifically in the case of `level`, you figured out that only ascii characters are allowed when creating sub-levels.

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -12,7 +13,7 @@ import { RecordsWrite } from '../messages/records-write.js';
 
 export class RecordsDeleteHandler implements MethodHandler {
 
-  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private dataStore: DataStore) { }
+  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private dataStore: DataStore, private eventLog: EventLog) { }
```
thehenrytsai, line 16: Sanity: `eventLog.append` is not invoked for delete?

src/store/storage-controller.ts
```diff
@@ -78,6 +80,9 @@ export class StorageController {
...
+    await eventLog.append(tenant, messageCid.toString());
```
thehenrytsai, line 85: Nit: `messageCid` is already a string.
```suggestion
    await eventLog.append(tenant, messageCid);
```

src/event-log/event-log.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * closes the connection to the underlying store
+   */
+  close(): Promise<void>;
+
+  /**
+   * adds an event to a tenant's event log
+   * @param tenant - the tenant's DID
+   * @param messageCid - the CID of the message
```
thehenrytsai, line 20: Add the expected return type in documentation (watermark).

tests/event-log/event-log-level.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      for (let i = 0; i < 9; i += 1) {
+        const { message } = await TestDataGenerator.generateRecordsWrite({ requester });
+        const messageCid = await computeCid(message);
+        messageCids.push(messageCid);
+
+        await eventLog.append(requester.did, messageCid.toString());
+      }
+
+      const events = await eventLog.getEventsAfter(requester.did);
```
thehenrytsai, line 90: Would be nice if there is a test that makes sure watermark is returned correctly by `getEventsAfter()` also.

tests/event-log/event-log-level.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    it('appends a tenant namespaced entry into leveldb', async () => {
+      const { requester, message } = await TestDataGenerator.generateRecordsWrite();
+      const messageCid = await computeCid(message);
+
+      const watermark = await eventLog.append(requester.did, messageCid.toString());
+
+      for await (const [key, value] of eventLog.db.iterator()) {
+        expect(key).to.include(watermark);
+        expect(value).to.equal(messageCid.toString());
+      }
```
thehenrytsai, line 35: The use of loop confused me... consider:
```suggestion
      const iteratorResult = await eventLog.db.iterator().next();
      const [key, value] = iteratorResult.value;
      expect(key).to.include(watermark);
      expect(value).to.equal(messageCid);
```

tests/event-log/event-log-level.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    it('maintains order in which events were appended', async () => {
+      const { requester, message } = await TestDataGenerator.generateRecordsWrite();
+      const messageCid = await computeCid(message);
+
+      await eventLog.append(requester.did, messageCid.toString());
```
thehenrytsai, line 42: All occurrences of `messageCid.toString()` can just be `messageCid` in this file. (It probably threw you off because it used to be a CID object)

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  constructor(config?: EventLogLevelConfig) {
+    this.config = {
+      location: 'EVENTLOG',
+      createLevelDatabase,
+      ...config,
+    };
+
+    this.db = new LevelWrapper<string>({ ...this.config, valueEncoding: 'utf8' });
+    this.ulid = monotonicFactory();
```
thehenrytsai, line 33: Sanity: I am under the impression that the ID generated will be guaranteed monotonically increasing only if you use the same exact same instance created by monotonicFactory() by passing the same initialization/seed?

ie. I tried:
```
      for (let i = 0; i < 3; i++) {
        const ulid = monotonicFactory();
        console.log(ulid(), i);
      }
```
Output doesn't appear to be monotonically increasing.

src/dwn.ts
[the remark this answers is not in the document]
mistermoe: @diehuxx 

> but DWN B hasn't seen that watermark. Would that produce the same behavior as if no watermark were sent?

nope, it would cause `DWN B` to return nothing

src/event-log/event-log-level.ts
thehenrytsai: Sharing similar sentiment as @dcrousso here.  Since we are dealing with DIDs here, this doesn't feel necessary. Assuming `base32crockford` really "just" encodes, it is only going to make the key (a lot) longer.

src/event-log/event-log-level.ts
mistermoe: right, so putting `const ulid = monotonicFactory();` outside of the loop will make it so the output is monotonically increasing. e.g.

```js
const ulid = monotonicFactory();
let lastUlid = ulid();

for (let i = 0; i < 3_000_000; i++) {
  const nextUlid = ulid();
  console.log(nextUlid, i);
  console.assert(nextUlid > lastUlid);

  lastUlid = nextUlid;
}
```

src/interfaces/records/handlers/records-delete.ts
mistermoe: good catch

src/event-log/event-log.ts
thehenrytsai, line 32: Question: (fine if this is out of scope) have you thought about how we handle large amount events to be returned?

An easy/lazy way to do this would be to wrap `Event[]` and return a subset if exceeding some threshold:

```javascript
{
  events: Event[];
  hasMore: boolean;
}
```

src/event-log/event-log-level.ts
dcrousso: IMO this should be a separate commit and should really be applied to _all_ places we use `tenant` with `LevelWrapper`

> 1. DIDs can be very long and in general can be different lengths (ion short form vs. long form). `base32crockford` produces a fixed-length output, which makes it easier to handle and compare than the variable-length

are you sure that `base32crockford` produces a fixed-length output?  that's not what im seeing :(
```
base32crockford.encode((new TextEncoder).encode('123'))    // 64S36
base32crockford.encode((new TextEncoder).encode('123456')) // 64S36D1N6R
```

> 2. `base32crockford` encoding also produces output that consists of a limited set of alphanumeric characters, making it more suitable for use in contexts where special characters or binary data are not allowed or need to be avoided. Specifically in the case of `level`, you figured out that only ascii characters are allowed when creating sub-levels.

i agree with the sentiment here, but aren't we already "safe" since it's a DID?  i.e. all characters should already be between `37` and `127`

src/event-log/event-log-level.ts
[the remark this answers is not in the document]
diehuxx: Resolving, since I see Devin asked the same question in another comment. Looks like that discussion is still open.

tests/event-log/event-log-level.spec.ts
mistermoe: the lack of loop confuses me 😅 . Happy to use whatever most people find least confusing. @dcrousso @diehuxx thoughts?

dcrousso: perhaps we do a bit of both?
```
const entries = await asyncGeneratorToArray(eventLog.db.iterator());
expect(entries.length).to.equal(1);
expect(entries[0][0]).to.include(watermark);
expect(entries[0][1]).to.equal(messageCid);
```

tests/event-log/event-log-level.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    beforeEach(async () => {
+      await eventLog.clear();
+    });
+
+    after(async () => {
+      await eventLog.close();
+    });
+
+    it('appends a tenant namespaced entry into leveldb', async () => {
```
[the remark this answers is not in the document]
mistermoe: reworded test description and adjusted test to incorporate your suggestion. change can be seen [here](https://github.com/TBD54566975/dwn-sdk-js/blob/66f562395473a4bbc069cc7f9303b176d60157dc/tests/event-log/event-log-level.spec.ts)

src/event-log/event-log.ts
mistermoe: definitely crossed my mind. am going to hold off until we discuss how we want to approach pagination in general

src/event-log/event-log-level.ts
mistermoe: @thehenrytsai any more questions on this front?

tests/event-log/event-log-level.spec.ts
mistermoe: done

src/event-log/event-log.ts
mistermoe: @dcrousso adding a visualization here to illustrate what you're talking about.

image

then thinking out loud now...  🤔 

any time an incoming `RecordsWrite` or `RecordsDelete` targets a pre-existing `recordId`, all intermediate `RecordsWrites` _except_ for the _initial_ `RecordsWrite` _and_ the _latest_ message (which could be a `RecordsWrite` _or_ a `RecordsDelete`) are dropped. 

the response of any `EventsGet` _after_ `t=2` will end up including events with `messageCid`s that no longer exist. 

i suppose the path of least resistance to accommodate this is to delete the events associated to those dropped messages. Instinctively (with no logical reason to back it), deleting events from a log that's supposed to keep track of _all_ persisted messages feels off.. but the DWN design seems to be alright with that, so, it sounds like i need to implement a `deleteEvent` or `deleteEvents`

mistermoe: tried my best to ignore this problem out of existence but it didn't work

src/dwn.ts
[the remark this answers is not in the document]
mistermoe: > Nit: add eventlog to `dump()`

@diehuxx  [done!](https://github.com/TBD54566975/dwn-sdk-js/commit/a267bac68e44548231f8845d000d735c585331bb) thanks

tests/event-log/event-log-level.spec.ts
mistermoe: @thehenrytsai, i'm going to use the fact that `for await (const [key, value] of eventLog.db.iterator()) {` [already exists in other parts of the codebase](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/level-wrapper.ts#L224-L226) (aka the consistency argument) to my benefit

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  async append(tenant: string, messageCid: string): Promise<string> {
+    const tenantEventLog = await this.db.partition(tenant);
+
+    const watermark = this.ulid();
+    await tenantEventLog.put(watermark, messageCid);
+
+    return watermark;
+  }
+  async getEventsAfter(tenant: string, watermark?: string): Promise<Event[]> {
```
dcrousso, line 54: Style: missing newline

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,71 @@
...
+    const tenantEventLog = await this.db.partition(tenant);
+
+    const opts = watermark ? { gt: watermark } : {};
+    const events: Array<Event> = [];
+
+    for await (const [key, value] of tenantEventLog.iterator(opts)) {
+      const event = { watermark: key, messageCid: value };
```
dcrousso, line 61: it feels kinda odd that we'd give a `watermark` for each `event`

should we do some sort of `Math.max` so that we only return the most recent one (as that's likely what the client will want to do anyways)?

src/event-log/event-log-level.ts
```diff
@@ -0,0 +1,71 @@
...
+      events.push(event);
+    }
+
+    return events;
+  }
+
+  async dump(): Promise<void> {
+    await this.db['dump']?.();
```
dcrousso, line 69: NIT: do you mind surrounding this with `console.group('db')` and `console.groupEnd()`?  that way we're consistent with other things (and if we change `EventLogLevel` in the future to have more than just `db` we'd be able to print things in nice clean groupings)

src/store/storage-controller.ts
```diff
@@ -33,9 +35,9 @@ export class StorageController {
     // but NOTE: it is possible that a data stream is not given in such case, for instance,
     // a subsequent RecordsWrite that changes the `published` property, but the data hasn't changed,
     // in this case requiring re-uploading of the data is extremely inefficient so we take care allow omission of data stream
-    if (message.descriptor.dataCid !== undefined) {
-      const messageCid = await Message.getCid(message);
+    const messageCid = await Message.getCid(message);
```
dcrousso, line 38: NIT: this should probably go before the comment, as the comment pertains to the `if` 😅

tests/event-log/event-log-level.spec.ts
thehenrytsai: No worries, not something worth arguing over!

src/event-log/event-log-level.ts
thehenrytsai: Yes, more question(s):

`monotonicFactory();` is instantiated every time the server is restarted, which is more analogous to my code snippet no?

I am just not clear how it can/is ensuring monotonic increments over _multiple_ instances. Granted the factory is invoked at a far quicker succession than restarting a service, so as long as the instances created no less than a second apart can have monotonic increasing behavior I guess it's fine.

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -66,6 +67,9 @@ export class RecordsDeleteHandler implements MethodHandler {
 
       await this.messageStore.put(tenant, incomingMessage, indexes);
 
+      const messageCid = await computeCid(incomingMessage);
```
thehenrytsai, line 70: Nit: consider using `Message.getCid()` for future proofing, even though they end up being the same right now, it hasn't been the case for months when we had `encodedData`, and there is potential that this will be the case again if/when we add `encodedData` (or other properties) back.

src/event-log/event-log-level.ts
mistermoe: @thehenrytsai bear in mind that this is `EventLogLevel`. Multiple instances is out of the question as only a single process can communicate with level at any given time. Multiple instances using `EventLogLevel` would be _net new_ DWNs. 

Restarting a single DWN would re-instantiate the monotonic factory which uses the `Date.now()` value at time of factory creation as its seed. 

Scenarios where multiple instances are communicating with the same stores should not use leveldb period.

src/event-log/event-log.ts
mistermoe: would like to do something like this:

```js
async function deleteEventsByCid(tenant, cids) {
  const cidSet = new Set(cids);
  const tenantEventLog = this.db.partition(tenant);
  const ops = [];

  for await (const [key, value] of tenantEventLog.iterator()) {
    if (cidSet.has(value)) {
      ops.push({ type: 'del', key });
    }
  }

  await this.tenantEventLog.batch(ops);
}
```

the inefficiency here is that we have to iterate through a tenant's _entire_ event log to find all CIDs. this inefficiency is compounded given the way that [deleteAllOlderMessagesButKeepInitialWrite](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/records-interface.ts#L15-L42) is written and more generally how [RecordsWriteHandler](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-write.ts#L24-L103) and [RecordsDeleteHandler](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-delete.ts#L17-L77) are implemented.

imo, there's quite a bit in those 3 areas of code that can be simplified and optimized. probably best to leave this simplify & optimize refactor for a separate PR, so i'll implement `deleteEventsByCid` as shown above and slot it into the `deleteAllOlderMessagesButKeepInitialWrite` loop. 

lmk if anyone else has other ideas

dcrousso: for now could/should we just have it be `delete(tenant: string, messageCid: string): Promise<void>`?  i.e. only do one message at a time?

(i mentioned it offline, but ill put it here for future reference just in case) if we're OK with the storage cost we could also have it include a reverse lookup (i.e. `messageCid` -> `watermark`) to make the deletion much faster (i.e. no need to `iterator` over everything, just one `get` and `delete`)

i definitely appreciate the desire to use `batch`, but unless we include the `watermark` in the `Message` itself (and require that the whole `Message` be provided for `delete`) i dont think we'll be able to easily avoid some sort of lookup somewhere to get the `watermark` in order to delete it :/

dcrousso: oh hmm here's an idea i just had :D

what if instead of doing this when new messages come in, we lazily do this inside `getEventsAfter`?

i.e. instead of `delete` whenever we remove something from the `MessageStore`, instead we wait until someone calls `getEventsAfter` and do a cleanup whenever we notice that an override/replacement/delete/etc. happened?  we already have to `iterator` inside `getEventsAfter` so why not leverage that to get the `watermark` key while we're at it?  and once we've finished the `iterator` we can then do a cleanup of the resulting `events` to find things that we need to delete and submit one final `batch` to remove them all?

src/event-log/event-log.ts
mistermoe: @dcrousso, refactored method signature. [diff](https://github.com/TBD54566975/dwn-sdk-js/commit/902d7babcfcc2cfa3cc34d1bc944e7e3448741a1#diff-1107f46751ce21d33b21d75810446ddf320ec0bc8b7170b3f8ccbefa4194c2e3L54-R59). lmk what you think

src/event-log/event-log.ts
mistermoe: > for now could/should we just have it be `delete(tenant: string, messageCid: string): Promise<void>`? i.e. only do one message at a time?
> 
> (i mentioned it offline, but ill put it here for future reference just in case) if we're OK with the storage cost we could also have it include a reverse lookup (i.e. `messageCid` -> `watermark`) to make the deletion much faster (i.e. no need to `iterator` over everything, just one `get` and `delete`)
> 
> i definitely appreciate the desire to use `batch`, but unless we include the `watermark` in the `Message` itself (and require that the whole `Message` be provided for `delete`) i dont think we'll be able to easily avoid some sort of lookup somewhere to get the `watermark` in order to delete it :/

I think i can still keep `deleteEventsByCid` as it is and just pass it a single cid array for now.

src/event-log/event-log-level.ts
mistermoe: will pass on this for now but i appreciate the suggestion!

src/event-log/event-log.ts
mistermoe: @dcrousso [this commit](https://github.com/TBD54566975/dwn-sdk-js/pull/290/commits/74a91abd8c17a2c60395f87ccd22b5ac71432071) remedies the original point you made

src/interfaces/records/handlers/records-delete.ts
mistermoe: i've exclusively used `computeCid` everywhere in this PR 😅 . do you want me to go ahead and update all the `computeCid`s i added to `Message.getCid()`?

image

src/event-log/event-log.ts
dcrousso: allowing the client to provide any `gt` they want might be a bit vague/permissive, but since this is really only meant to be used internally it's probably fine :)

also it's not like they could do much else with the `gt` since all of the keys (after the `partition` based on the `tenant`) would be a `watermark` anyways

src/interfaces/records/handlers/records-delete.ts
thehenrytsai: I refrained from asking to make the same change in the new tests added, because one can argue that tests should not use product code for validation. But weighing the trade offs, I'd be for changing tests to use `Message.getCid()` also as long as you are okay with it!

src/dwn.ts
```diff
@@ -30,11 +30,15 @@ export class Dwn {
...
-    this.tenantGate = config.tenantGate;
+    this.didResolver = config.didResolver!;
+    this.messageStore = config.messageStore!;
+    this.dataStore = config.dataStore!;
+    this.eventLog = config.eventLog!;
+    this.tenantGate = config.tenantGate!;
+    this.didResolver = config.didResolver!;
+    this.messageStore = config.messageStore!;
+    this.dataStore = config.dataStore!;
+    this.tenantGate = config.tenantGate!;
```
thehenrytsai: Sanity: duplicate code due to merge?

src/event-log/event-log.ts
```diff
@@ -29,5 +34,12 @@ export interface EventLog {
    * @param tenant
    * @param watermark
    */
-  getEventsAfter(tenant: string, watermark?: string): Promise<Array<Event>>
+  getEvents(tenant: string, options?: GetEventsOptions): Promise<Array<Event>>
+
+  /**
+   * deletes any events that have any of the cids provided
```
thehenrytsai, line 40: Add documentation on what the returned `number` is for implementors.

src/event-log/event-log.ts
thehenrytsai: For me, I think it is okay for sync to not able to get a message of a given CID, and we can treat the message being obsolete as the (only?) expected condition of a message fetch to return nothing. There is probably no easy way to prevent this condition to occur: even if we delete the associated events at the time a message is dropped, because a message could be dropped right after `getEvents()` gets called causing immediate message fetch by CID to "fail".

I see deleting the events of deleted messages as an optimization, one dumb place to do it is on demand when fetching message by CID fails. Though like your initial thoughts, I have mixed feelings about deleting "events" from log, but coming around to it because keeping it serves no practical purposes and is just a long-term perf and storage drag.

src/dwn.ts
mistermoe: good catch

src/event-log/event-log.ts
mistermoe: thanks

mistermoe:
@thehenrytsai i went ahead and included functional tests. pls lmk if you'd rather them go elsewhere.

mistermoe:
just realized that we'll need to delete events when a protocol is overwritten as well. Specifically, [here](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/protocols/handlers/protocols-configure.ts#L73-L78). 😭

mistermoe:
> just realized that we'll need to delete events when a protocol is overwritten as well. Specifically, [here](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/protocols/handlers/protocols-configure.ts#L73-L78). 😭

[done](https://github.com/TBD54566975/dwn-sdk-js/pull/290/commits/e1643a0243252fa8845f6d388c70e1eca03956ce)

src/event-log/event-log.ts
mistermoe: i disagree with the notion that deleting events that aren't present at time of delete or overwrite is an "optimization".  

we cannot assume that a CID not being present _always_ implies that the CID has been deleted. this means that sync _must always_ fetch _all_ data associated to messages. a situation where that's not feasible is 4gb blu-ray chonkers.

mistermoe: cc: @csuwildcat

src/interfaces/records/handlers/records-delete.ts
mistermoe: will replace all instances of `computeCid` in a subsequent PR.

mistermoe: will tackle this in a subsequent PR. issue created: #300

src/event-log/event-log.ts
thehenrytsai: Hm... maybe I misread the design, I thought the event log keeps track of _message_ CIDs? If so, isn't data CIDs mostly irrelevant to the discussion? When DWN returns a message CID from even log, but fails to give you the actual message when you try to fetch it by message CID, it seems fine to assume that it is being deleted no? If we don't love the implicitness, we could return an explicit "return code" that says "deleted". But regardless of the explicitness/implicitness, data CID(s) seems to be a layer of abstraction away, not seeing data associated with the message needing to always be downloaded by a DWN. Am I smocking crack...?!

mistermoe: naw you're right @thehenrytsai i'm the one smocking crack. it's message CID based. had something else on my mind and was delirious by the time i wrote this

mistermoe:
does anyone have any reservations left here?

thehenrytsai:
The only reservation for me is that the code does a full scan of the event log for every clobber, but fine if we create an issue to improve event deletion by CIDs as a separate item to unblock the merge of this PR.

mistermoe:
@thehenrytsai i'm now storing events in watermark order _and_ cid order. this prevents us from having to scan a tenant's entire event log when deleting by CID. cost incurred is 2x storage for events. the size of each event is `numDidBytes + numCidBytes`.

requests to change anything else about `EventLogLevel` can be submitted as separate issues that can be tackled later. I'd like to get this PR merged so that i can move to the next sync-related PR.

csuwildcat:
Sounds like a good plan to me, Moe

thehenrytsai (approved):
🥇 🙏 💯 🎸

src/event-log/event-log.ts
mistermoe: see https://github.com/TBD54566975/dwn-sdk-js/pull/290#issuecomment-1500738602

PR #294, opened 2023-04-05
#276 - added SECP256K1 hierarchical key derivation utilities

thehenrytsai:
* added SECP256K1 hierarchical key derivation utilities
* refactored existing SECP256K1 code for better reusability

Note that only 4 public methods are added in this PR (all starts with `derive`), the rest of the methods/code were already there, just refactored.


src/utils/secp256k1.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Parses the given key derivation path.
+   * @returns Path segments if successfully validate the derivation path.
+   * @throws {DwnError} with `DwnErrorCode.HdKeyDerivationPathInvalid` if derivation path fails validation.
+   */
+  private static parseAndValidateKeyDerivationPath(derivationPath: string): string[] {
+    const pathSegments = derivationPath.split('/');
+
+    if (pathSegments.length === 0 || pathSegments.includes('')) {
```
diehuxx, line 201: Do we need to validate that path segments are in ASCII or anything like that?

tests/jose/algorithms/signing/secp256k1.spec.ts
diehuxx, line 1: We should move this test file `tests/utils/secp256k1.spec.ts`

tests/jose/algorithms/signing/secp256k1.spec.ts
```diff
@@ -1,31 +1,93 @@
...
+      const { publicKey, privateKey } = await Secp256k1.generateKeyPairRaw();
+
+      const path = '1/2/3/4/5/6/7/8/9/10';
+
+      const derivedKeyFromPublicKey = await Secp256k1.derivePublicKey(publicKey, path);
+      const derivedKeyFromPrivateKey = await Secp256k1.derivePublicKey(privateKey, path);
+      expect(Comparer.byteArraysEqual(derivedKeyFromPublicKey, derivedKeyFromPrivateKey)).to.be.true;
+    });
+
+    it('should derive the same public key using either the private or public counterpart of the same key pair', async () => {
```
diehuxx, line 84: Related to my other comment, should we also test derivation path validation for paths that include like emojis and empty strings?

src/utils/secp256k1.ts
thehenrytsai: No, in fact, I have a subsequent PR that changed path to be an array of strings so that the path has less limitation (such as delimiter by `/`). I'll see if I can easily move that change into this PR. If not, you'll see it in the upcoming one.

tests/jose/algorithms/signing/secp256k1.spec.ts
diehuxx: resolving based on your reply above. ty

tests/jose/algorithms/signing/secp256k1.spec.ts
thehenrytsai: Good suggestion!

diehuxx (approved):

Issue #289, opened 2023-03-31
Consider separating `processMessage()` into individual handling methods per DWN Interface Method


thehenrytsai:
### Background:
With the introduction and implementation of `RecordsRead` (and other Interface Methods to follow like `RecordsCommit`), it is becoming apparent that the return type cannot be represented by a universal type that covers all need, unless we hack our way by adding a bunch of optional properties or even just use `any`.

### Task Details:
- Maybe the time has come for us to consider evolve `processMessage()` which has served well, and replace/augment it with individual handlers:

`handleRecordsRead(...): RecordsReadResponse { ... }` ✅
`handleRecordsWrite(...): RecordsWriteResponse { ... }`
...

### Objective:
- create a handler for `handleRecordsWrite(...): RecordsWriteResponse { ... }`
> `handleRecordsRead(...): RecordsReadResponse { ... }` was introduced in [this pull request](https://github.com/TBD54566975/dwn-sdk-js/pull/296) if you're interested in an example.

## Picking Up This Issue:
- If you'd like to work on this, please comment "picking this up" below, and I'll assign the issue to you 

## Questions:
- If you need assistance or clarification, feel free to comment below.
- Feeling overwhelmed? Join our [`#hack-together` channel](https://discord.gg/tbd) and collaborate with a buddy.
- New to open-source? Check out our [Hacktober `#getting-started` channel](https://discord.gg/tbd).
- [Leaderboard](https://github.com/TBD54566975/developer.tbd.website/issues/721): Check your ranking after finishing this task.

### Resources:
- **Creating a Pull Request**: If you're new to GitHub and unsure how to create a pull request, follow this [step-by-step guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Remember, communication is key! If you have any questions or face any challenges, we're here to help so please don't hesitate to reach out.

Good Luck! 🍁

shobitb:
@thehenrytsai can I pick this up?

If yes, a few questions before I get started —
1. Do you think the response type for `handleRecordsWrite` can continue to be `GenericMessageReply` (like in the [Handler](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/handlers/records-write.ts#L31))?  Or introduce a new `RecordsWriteReply` (but that would essentially be a `GenericMessageReply` (hence, an alias))?
[...]

PR #297, opened 2023-04-05
return the parsed message if `processMessage` is successful

dcrousso:
this will make it easier for clients (e.g. `web5-js`) to use the actual `class` object instead of having to manipulate JSON

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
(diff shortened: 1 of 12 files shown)


thehenrytsai:
I'd try to tag on the `message` outside of individual handlers, in one place (perhaps somewhere in `Dwn.processMessage()`) so that the handlers don't have to know about it and there is one place that this takes place.

Above said, @frankhinek seems to have ideas already planned for `web5-js` so we should probably hold off until this work is confirmed necessary.

Also, I am aware that `MessageReply` is becoming polluted..., there is a separate discussion to refactor `processMessage()` altogether, see #289.

dcrousso:
> I'd try to tag on the `message` outside of individual handlers, in one place (perhaps somewhere in `Dwn.processMessage()`) so that the handlers don't have to know about it and there is one place that this takes place.

my concern with this is that outside only has the raw JSON payload, not the `class` instance result of `parse`

giving the developer access to the `class` makes it _much_ nicer to access relevant fields (e.g. we could add a `get recordId` to all of `RecordsWrite`, `RecordsRead`, and `RecordsDelete` so that it's a similar API)

diehuxx:
Closing, since there hasn't been any movement or discussion on this PR in over a month. If anyone is still interested in this proposal and wants to track it, feel free to open an issue.

PR #305, opened 2023-04-08
`EventsGet`

mistermoe:
2nd of 3 PRs to implement sync foundations. preceded by #290

Introduces an `EventsGet` message to retrieve events. Still working on tests for `EventsGetHandler` but i wanted to kick off the review process because it can be lengthy. I imagine there will be some discussion around `MessageReply`.

The PR that follows this one will introduce the `MessagesGet` interface method

### DIFF
--- a/src/interfaces/events/handlers/events-get.ts
+++ b/src/interfaces/events/handlers/events-get.ts
@@ -28,7 +28,7 @@ export class EventsGetHandler implements MethodHandler<Event> {
 
       if (eventsGet.author !== tenant) {
         //! FIXME: think of better error message
-        throw new Error('author must be tenant.');
+        throw new Error('message author must be tenant.');
       }
     } catch (e) {
       return MessageReply.fromError(e, 401);
--- a/src/interfaces/events/handlers/events-get.ts
+++ b/src/interfaces/events/handlers/events-get.ts
@@ -4,7 +4,6 @@ import type { GetEventsOptions } from '../../../event-log/event-log.js';
 import type { MethodHandler } from '../../types.js';
 import type { EventsGetMessage, EventsGetReply } from '../types.js';
 
-
 import { authenticate } from '../../../core/auth.js';
 import { EventsGet } from '../messages/events-get.js';
 import { MessageReply } from '../../../core/message-reply.js';
@@ -27,15 +26,14 @@ export class EventsGetHandler implements MethodHandler {
       await authenticate(message.authorization, this.didResolver);
 
       if (eventsGet.author !== tenant) {
-        //! FIXME: think of better error message
         throw new Error('message author must be tenant.');
       }
     } catch (e) {
       return MessageReply.fromError(e, 401);
     }
 
-    // if watermark was provided in message, get all events _after_ the watermark. otherwise,
-    // get all events.
+    // if watermark was provided in message, get all events _after_ the watermark.
+    // Otherwise, get all events.
     let options: GetEventsOptions | undefined;
     if (message.descriptor.watermark) {
       options = { gt: message.descriptor.watermark };
--- a/src/interfaces/events/handlers/events-get.ts
+++ b/src/interfaces/events/handlers/events-get.ts
@@ -4,9 +4,9 @@ import type { GetEventsOptions } from '../../../event-log/event-log.js';
 import type { MethodHandler } from '../../types.js';
 import type { EventsGetMessage, EventsGetReply } from '../types.js';
 
-import { authenticate } from '../../../core/auth.js';
 import { EventsGet } from '../messages/events-get.js';
 import { MessageReply } from '../../../core/message-reply.js';
+import { authenticate, authorize } from '../../../core/auth.js';
 
 type HandleArgs = {tenant: string, message: EventsGetMessage};
 
@@ -24,10 +24,7 @@ export class EventsGetHandler implements MethodHandler {
 
     try {
       await authenticate(message.authorization, this.didResolver);
-
-      if (eventsGet.author !== tenant) {
-        throw new Error('message author must be tenant.');
-      }
+      await authorize(tenant, eventsGet);
     } catch (e) {
       return MessageReply.fromError(e, 401);
     }
--- a/src/interfaces/events/handlers/events-get.ts
+++ b/src/interfaces/events/handlers/events-get.ts
@@ -1,8 +1,8 @@
 import type { DidResolver } from '../../../index.js';
-import type { EventsGetMessage } from '../types.js';
+import type { EventLog } from '../../../event-log/event-log.js';
 import type { GetEventsOptions } from '../../../event-log/event-log.js';
 import type { MethodHandler } from '../../types.js';
-import type { Event, EventLog } from '../../../event-log/event-log.js';
+import type { EventsGetMessage, EventsGetReply } from '../types.js';
 
 
 import { authenticate } from '../../../core/auth.js';
@@ -14,7 +14,7 @@ type HandleArgs = {tenant: string, message: EventsGetMessage};
 export class EventsGetHandler implements MethodHandler {
   constructor(private didResolver: DidResolver, private eventLog: EventLog) {}
 
-  public async handle({ tenant, message }: HandleArgs): Promise<MessageReply> {
+  public async handle({ tenant, message }: HandleArgs): Promise<EventsGetReply> {
     let eventsGet: EventsGet;
 
     try {
@@ -44,8 +44,8 @@ export class EventsGetHandler implements MethodHandler {
     const events = await this.eventLog.getEvents(tenant, options);
 
     return {
-      status  : { code: 200, detail: 'OK' },
-      entries : events
+      status: { code: 200, detail: 'OK' },
+      events
     };
   }
 }
--- a/src/interfaces/events/handlers/events-get.ts
+++ b/src/interfaces/events/handlers/events-get.ts
@@ -1,0 +1,51 @@
+import type { DidResolver } from '../../../index.js';
+import type { EventsGetMessage } from '../types.js';
+import type { GetEventsOptions } from '../../../event-log/event-log.js';
+import type { MethodHandler } from '../../types.js';
+import type { Event, EventLog } from '../../../event-log/event-log.js';
+
+
+import { authenticate } from '../../../core/auth.js';
+import { EventsGet } from '../messages/events-get.js';
+import { MessageReply } from '../../../core/message-reply.js';
+
+type HandleArgs = {tenant: string, message: EventsGetMessage};
+
+export class EventsGetHandler implements MethodHandler<Event> {
+  constructor(private didResolver: DidResolver, private eventLog: EventLog) {}
+
+  public async handle({ tenant, message }: HandleArgs): Promise<MessageReply<Event>> {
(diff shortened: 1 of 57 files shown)


mistermoe:
added handler tests

thehenrytsai (commented):
Looks pretty clean on DWN method implementation front, nothing much to say there! Have comments on `MessageReply` changes as you predicted.

tests/interfaces/events/handlers/events-get.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    message['descriptor']['troll'] = 'hehe';
+
+    const reply = await dwn.processMessage(alice.did, message);
+
+    expect(reply.status.code).to.equal(400);
+    expect(reply.entries).to.not.exist;
+    expect(reply.data).to.not.exist;
+  });
+
+  it('retuns all events for a tenant if watermark is not provided', async () => {
```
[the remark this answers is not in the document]
mistermoe: thanks @thehenrytsai !

src/interfaces/events/handlers/events-get.ts
```diff
@@ -0,0 +1,51 @@
...
+      eventsGet = await EventsGet.parse(message);
+    } catch (e) {
+      return MessageReply.fromError(e, 400);
+    }
+
+    try {
+      await authenticate(message.authorization, this.didResolver);
+
+      if (eventsGet.author !== tenant) {
+        //! FIXME: think of better error message
```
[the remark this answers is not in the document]
mistermoe: haha fair. haven't been able to think of a good error message yet! all ears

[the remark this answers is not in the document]
thehenrytsai: Honestly I thought whatever you got is good enough, author must be tenant, what more do you want! = ]

src/interfaces/types.ts
```diff
@@ -5,13 +5,13 @@ import type { Readable } from 'readable-stream';
...
+export interface MethodHandler<ResultEntryKind> {
   /**
    * Handles the given message and returns a `MessageReply` response.
    */
   handle(input: {
     tenant: string;
     message: BaseMessage;
     dataStream?: Readable
-  }): Promise<MessageReply>;
+  }): Promise<MessageReply<ResultEntryKind>>;
```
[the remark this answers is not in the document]
mistermoe: @thehenrytsai , i def wasn't a fan of the generic approach specifically bc of  [this weirdness](https://github.com/TBD54566975/dwn-sdk-js/blob/events-get/src/interfaces/records/handlers/records-read.ts#L13).  it should be `MessageReply<never>` because `RecordsRead` will  _never_ have `entries`!  

but i can't use `never` without changing [this](https://github.com/TBD54566975/dwn-sdk-js/blob/events-get/src/dwn.ts#L26) from `unknown` to `any`. so end up being stuck between a rock and a hard place.

 using `unknown` implies that _something_ will for sure be there, but exactly _what_ is unknown. In the case of `RecordsRead` this is a straight up lie  because we _know_ `entries` will _never_ be there. 

whereas `any` means: "could be.. or maybe not 🤷🏽 ." neither of which necessarily seem helpful.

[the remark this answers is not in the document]
mistermoe: @thehenrytsai i can't use the approach that you used and suggested i follow from #296 without making some changes. here's why:

the [`MethodHandler`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/types.ts#L16) interface has a return signature of `Promise<MessageReply>`.  This works out fine in the case of [`RecordsReadReply`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/types.ts#L120-L130) because `RecordsReadReply` is an intersection of `BaseMessageReply` _and_ an object with an _optional_ `record` object. the fact that `record` is _optional_ in addition to the fact that `RecordsReadReply` doesn't affect, use, or modify `events` _or_ `data` as defined in [`MessageReply`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/core/message-reply.ts#L19-L49) makes it so that the return signature of [`RecordsReadHandler.handle`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-read.ts#L20) remains adherent to `MethodHandler` interface. 

in the case of `EventsGet`, `EventsGetReply` would look something like this:
```ts
export type EventsGetReply = BaseMessageReply & {
  entries?: Event[];
};
```
_unrelated note: also fine with `entries` being required, but it isn't for `RecordsQuery` replies when an error is present iirc_

changing `EventsGetHandler.handler`'s return signature to `Promise<EventsGetReply>` violates the `MethodHandler` interface specifically because the `entries` types of `MessageReply` and `EventsGetReply` are incompatible. 

unless i'm missing something, which i might be! we'll likely have to rethink the plan mentioned in your comment.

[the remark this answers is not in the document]
dcrousso: maybe call it `events` instead of `entries`?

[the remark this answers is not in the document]
mistermoe: bless you kind sir. did exactly that

src/interfaces/events/handlers/events-get.ts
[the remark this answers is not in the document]
dcrousso: most of the other handlers do something like `eventsGet.authorize(tenant)`

should we add a `EventsGet.prototype.authorize` to do this instead?
```
public async authorize(tenant: string): Promise<void> {
  // if author/requester is the same as the target tenant, we can directly grant access
  if (this.author === tenant) {
    return;
  } else {
    throw new Error('message failed authorization');
  }
}
```

src/interfaces/events/handlers/events-get.ts
```diff
@@ -0,0 +1,51 @@
...
+        throw new Error('message author must be tenant.');
+      }
+    } catch (e) {
+      return MessageReply.fromError(e, 401);
+    }
+
+    // if watermark was provided in message, get all events _after_ the watermark. otherwise,
+    // get all events.
```
dcrousso, line 38: NIT: id put the `otherwise,` on the next line so it reads more like a list instead of a wrapped sentence 😅

src/interfaces/events/messages/events-get.ts
```diff
@@ -0,0 +1,39 @@
...
+    const descriptor: EventsGetDescriptor = {
+      interface : DwnInterfaceName.Events,
+      method    : DwnMethodName.Get,
+    };
+
+    if (options.watermark) {
+      descriptor.watermark = options.watermark;
+    }
+
+    Message.validateJsonSchema({ descriptor, authorization: { } });
```
dcrousso, line 31: both `RecordsWrite` and `RecordsRead` do this after fully constructing the `message`.  can/should we do the same here?

src/interfaces/events/messages/events-get.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const authorization = await Message.signAsAuthorization(descriptor, options.authorizationSignatureInput);
+    const message = { descriptor, authorization };
+
+    const eventsGet = new EventsGet(message);
+    return eventsGet;
```
dcrousso, line 37: NIT: id just directly `return new EventsGet(message);` without creating an intermediary variable

src/interfaces/events/types.ts
```diff
@@ -0,0 +1,18 @@
+import type { BaseMessage } from '../../core/types.js';
+import type { BaseMessageReply } from '../../core/message-reply.js';
+import type { Event } from '../../event-log/event-log.js';
+import type { DwnInterfaceName, DwnMethodName } from '../../core/message.js';
+
+export type EventsGetDescriptor = {
```
dcrousso, line 6: should this `& Descriptor`?

or are you intentionally trying to avoid `dataCid` and `dataSize`?  if so, could we use `Pick` so at least there's some continuity?

src/interfaces/events/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  interface : DwnInterfaceName.Events;
+  method: DwnMethodName.Get;
+  watermark?: string;
+};
+
+export type EventsGetMessage = BaseMessage & {
+  descriptor: EventsGetDescriptor;
+};
+
+export type EventsGetReply = BaseMessageReply & {
```
dcrousso, line 16: rather than create a new subtype, should we just add something like `events?: Event[];` to the `MessageReply`?

otherwise how will a developer reason about "what is the return type of `dwn.processMessage(tenant, message)`"?  are we expecting them to forcibly `as EventsGetReply` in order to get any sort of typing support?

tests/dwn.spec.ts
```diff
@@ -86,6 +86,18 @@ describe('DWN', () => {
...
+    it('should process an EventsGet message', async () => {
+      const alice = await DidKeyResolver.generate();
+      const { message } = await TestDataGenerator.generateEventsGet({ requester: alice });
+
+      const reply: EventsGetReply = await dwn.processMessage(alice.did, message);
+
+      expect(reply.status.code).to.equal(200);
+      expect(reply.events).to.be.empty;
+      expect(reply['data']).to.not.exist;
+
```
dcrousso, line 98: Style: unnecessary newline

src/interfaces/events/handlers/events-get.ts
```diff
@@ -0,0 +1,51 @@
+import type { DidResolver } from '../../../index.js';
+import type { EventLog } from '../../../event-log/event-log.js';
+import type { GetEventsOptions } from '../../../event-log/event-log.js';
+import type { MethodHandler } from '../../types.js';
+import type { EventsGetMessage, EventsGetReply } from '../types.js';
+
+
```
dcrousso, line 7: Style: unnecessary newline

dcrousso (dismissed):

src/interfaces/events/types.ts
mistermoe: currently yeah. that, or remain consistent with what henry's done for `RecordsRead` which is add a specific handler for that in `dwn.ts`

src/interfaces/events/types.ts
mistermoe: intentionally avoiding `dataCid` and `dataSize` because they'll never exist in an `EventsGetDescriptor`

src/interfaces/events/messages/events-get.ts
mistermoe: aye. instinctively wanted to do what you suggested. decided to check what the norm was in the codebase and it seems to be what i currently have. all except for `RecordsQuery`

mistermoe: nvm, `RecordsWrite` does it as well

src/interfaces/events/handlers/events-get.ts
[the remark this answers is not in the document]
mistermoe: switched to using `authorize` instead of repeating the code. it's a wee bit strange imo that the message classes have `authorize` methods because `tenant` is a bit out of context there

diehuxx (dismissed):
lgtm! thx for getting the changes in before Saturday, have a happy n healthy weekend :)

diehuxx (dismissed):

frankhinek (approved):
LGTM based on the most recent changes

Issue #285, opened 2023-03-29
Add `query` to list of supported actions for protocol-based message authorization


thehenrytsai:

csuwildcat:
Should add `read` to this as well.

diehuxx:
I've thought about protocol query a bit more. I think I've figured out what the "difficult" part is and I have an idea of how to approach it.

### Difficult Part
The current `ProtocolAuthorization#authorize` code is meant to look at the protocol definition for a single rule set based on the given actor and protocol path. For query, we potentially need to look down multiple `protocolPaths` to get multiple rule sets. So, calling `authorize` once at the top of `RecordsQuery#handler` won't be sufficient. We need to do some authorization for each record returned from the message store query.

### Approach
If `protocol` appears in the `RecordsQuery` filters: First, do a `messageStore#query` using all the filters given in the RecordsQuery descriptor. This will give us a list of `RecordsWrites`. Then, we iterate over each RecordsWrite to get the associated rule set for `recordsWrite.descriptor.protocolPath` and the actor of the RecordsQuery.

### Open Questions
* What do we return when the `RecordsQuery` filter does not specify `protocol`? Do we omit all protocol records? Or include protocol records, checking each protocol definition to see if the record is auth'd for `query`? If we include all that are allowed, that involves looking through potentially all protocol definitions on the DWN. Not sure if we want that.
  * @csuwildcat This is probably for you to answer.

csuwildcat:
I would require `protocol` to be set if the `protocolPath` is set, and not search through protocols at all if `protocol` isn't set.

thehenrytsai:
Agree that enforcing `protocol` as a filter will help scope the search of a protocol-authorized query.

> Then, we iterate over each RecordsWrite to get the associated rule set for recordsWrite.descriptor.protocolPath and the actor of the RecordsQuery.

This seems super heavy on processing. Is it possible to perhaps "flip" the logic? As in, go through the entire protocol definition hierarchy to find the authorized protocol paths matching the query, then perform a scoped query for each matched protocol path?

Also maybe I misinterpreted it, I vaguely recall @csuwildcat told us a couple days ago that he was okay for a protocol-authorized query to return records of a specific protocol path only? If that's true, then we can just make `protocolPath` also a required property for protocol-authorized query and it will make our lives much easier.

csuwildcat:
I have no issue making protocolPath required for protocol-related queries

diehuxx:
[...]

PR #307, opened 2023-04-10
Support protocol authorized reads

diehuxx:
Addresses part of #285

## Notable Changes
* `json-schemas/protocol-rule-set.json`: Previously `ProtocolRule`s allowed exactly one of (`anyone`, `author`, `recipient`). This PR allows for AT LEAST one of those; i.e. we may have multiple of (`anyone`, `author`, `recipient`).

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -24,23 +24,6 @@ export class ProtocolAuthorization {
     requesterDid: string | undefined,
     messageStore: MessageStore
   ): Promise<void> {
-    const isWrite = incomingMessage.message.descriptor.method === DwnMethodName.Write;
-
-    let recordsWrite: RecordsWrite;
-    if (isWrite) {
-      recordsWrite = incomingMessage as RecordsWrite;
-    } else {
-      const recordsRead = incomingMessage as RecordsRead;
-      const query = {
-        interface : DwnInterfaceName.Records,
-        method    : DwnMethodName.Write,
-        recordId  : recordsRead.message.descriptor.recordId,
-      };
-      const existingMessages = await messageStore.query(tenant, query) as TimestampedMessage[];
-      const recordsWriteMessage = await RecordsWrite.getNewestMessage(existingMessages) as RecordsWriteMessage;
-      recordsWrite = await RecordsWrite.parse(recordsWriteMessage);
-    }
-
     // fetch ancestor message chain
     let ancestorMessageChain: RecordsWriteMessage[] =
       await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, messageStore);
@@ -250,10 +233,12 @@ export class ProtocolAuthorization {
         allowRule.author.of,
         recordSchemaToLabelMap
       );
-      const expectedRequesterDid = Message.getAuthor(messageForAuthorCheck);
+      if (messageForAuthorCheck !== undefined) {
+        const expectedRequesterDid = Message.getAuthor(messageForAuthorCheck);
 
-      if (requesterDid === expectedRequesterDid) {
-        allowRule.author.to.forEach(action => allowedActions.add(action));
+        if (requesterDid === expectedRequesterDid) {
+          allowRule.author.to.forEach(action => allowedActions.add(action));
+        }
       }
     }
 
@@ -263,10 +248,12 @@ export class ProtocolAuthorization {
         allowRule.recipient.of,
         recordSchemaToLabelMap
       );
-      const expectedRequesterDid = messageForRecipientCheck.descriptor.recipient;
+      if (messageForRecipientCheck !== undefined) {
+        const expectedRequesterDid = messageForRecipientCheck.descriptor.recipient;
 
-      if (requesterDid === expectedRequesterDid) {
-        allowRule.recipient.to.forEach(action => allowedActions.add(action));
+        if (requesterDid === expectedRequesterDid) {
+          allowRule.recipient.to.forEach(action => allowedActions.add(action));
+        }
       }
     }
 
@@ -306,6 +293,7 @@ export class ProtocolAuthorization {
 
   /**
    * Gets the message from the message chain based on the path specified.
+   * Returns undefined if matching message does not existing in ancestor chain
    * @param messagePath `/` delimited path starting from the root ancestor.
    *                    Each path segment denotes the expected record type declared in protocol definition.
    *                    e.g. `A/B/C` means that the root ancestor must be of type A, its child must be of type B, followed by a child of type C.
@@ -315,12 +303,12 @@ export class ProtocolAuthorization {
     ancestorMessageChain: RecordsWriteMessage[],
     messagePath: string,
     recordSchemaToLabelMap: Map<string, string>
-  ): RecordsWriteMessage {
+  ): RecordsWriteMessage | undefined {
     const expectedAncestors = messagePath.split('/');
 
     // consider moving this check to ProtocolsConfigure message ingestion
     if (expectedAncestors.length > ancestorMessageChain.length) {
-      throw new Error('specified path to expected recipient is longer than actual length of ancestor message chain');
+      return undefined;
     }
 
     let i = 0;
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -1,13 +1,15 @@
 import type { MessageStore } from '../store/message-store.js';
-import type { RecordsWrite } from '../interfaces/records/messages/records-write.js';
+import type { RecordsRead } from '../interfaces/records/messages/records-read.js';
 import type { RecordsWriteMessage } from '../interfaces/records/types.js';
-import type { BaseMessage, Filter } from './types.js';
+import type { Filter, TimestampedMessage } from './types.js';
 import type { ProtocolDefinition, ProtocolRuleSet, ProtocolsConfigureMessage } from '../interfaces/protocols/types.js';
 
+import { RecordsWrite } from '../interfaces/records/messages/records-write.js';
 import { DwnInterfaceName, DwnMethodName, Message } from './message.js';
 
 const methodToAllowedActionMap: Record<string, string> = {
-  [DwnMethodName.Write]: 'write',
+  [DwnMethodName.Write] : 'write',
+  [DwnMethodName.Read]  : 'read',
 };
 
 export class ProtocolAuthorization {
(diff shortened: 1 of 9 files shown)


src/core/protocol-authorization.ts
```diff
@@ -84,7 +108,12 @@ export class ProtocolAuthorization {
    * Constructs a chain of ancestor messages
    * @returns the ancestor chain of messages where the first element is the root of the chain; returns empty array if no parent is specified.
    */
-  private static async constructAncestorMessageChain(tenant: string, recordsWrite: RecordsWrite, messageStore: MessageStore)
+  private static async constructAncestorMessageChain(
+    tenant: string,
+    incomingMessage: RecordsRead | RecordsWrite,
```
thehenrytsai, line 113: Sanity: this `incomingMessage` being used? If not, it's very odd that linter doesn't pick this up...

src/core/protocol-authorization.ts
```diff
@@ -121,14 +150,10 @@ export class ProtocolAuthorization {
    * Gets the rule set corresponding to the inbound message.
```
thehenrytsai, line 150:
```suggestion
   * Gets the rule set corresponding to the given message chain.
```

src/core/protocol-authorization.ts
```diff
@@ -163,14 +188,13 @@ export class ProtocolAuthorization {
    */
   private static verifyAllowedActions(
     tenant: string,
-    requesterDid: string,
-    incomingMessage: Message<BaseMessage>,
+    requesterDid: string | undefined,
+    incomingMessageMethod: string,
```
thehenrytsai, line 192: Nit:
```suggestion
    incomingMessageMethod: DwnMethodName,
```

src/core/protocol-authorization.ts
diehuxx: Oh you're right. Yea not sure why the linter didn't catch that. Will remove

src/core/protocol-authorization.ts
```diff
@@ -37,24 +57,28 @@ export class ProtocolAuthorization {
...
       protocolDefinition,
-      ancestorMessageChain,
+      [...ancestorMessageChain, recordsWrite.message],
       recordSchemaToLabelMap
     );
 
+    // For non-writes `recordsWrite` is actually part of the existing ancestor chain
+    if (!isWrite) {
+      ancestorMessageChain = [...ancestorMessageChain, recordsWrite.message];
+    }
```
thehenrytsai, line 68: Not liking this, because it means `constructAncestorMessageChain()` is kinda a lie 'coz it actually doesn't construct the ancestor message chain most of the time (assuming reads happen more often than write), so it's confusing.

tests/interfaces/records/handlers/records-read.spec.ts
```diff
@@ -135,6 +138,189 @@ describe('RecordsReadHandler.handle()', () => {
       expect(Comparer.byteArraysEqual(dataFetched, dataBytes!)).to.be.true;
     });
 
+    describe('protocol based reads', () => {
+      it('should allow read with allow-anyone rule', async () => {
+        // scenario: Alice writes an image to Caroline's DWN with Alice herself as recipient,
```
thehenrytsai, line 143: For my understanding: "with Alice herself as recipient": I am not able to understand the scenario, is this common? Is there a more realistic scenario that we can use?

tests/interfaces/records/handlers/records-read.spec.ts
```diff
@@ -135,6 +138,189 @@ describe('RecordsReadHandler.handle()', () => {
...
+          authorizationSignatureInput : Jws.createSignatureInput(dave)
+        });
+        const daveCaptionReadReply = await dwn.processMessage(caroline.did, daveCaptionRecordsRead.message);
+        expect(daveCaptionReadReply.status.code).to.equal(401);
+        expect(daveCaptionReadReply.status.detail).to.include('inbound message action \'read\' not in list of allowed actions');
+      });
+
+      // also test blocking of non-author
+      it('should allow read with author rule', async () => {
+        // scenario: Alice writes an image to Caroline's DWN, then Bob writes a reply to the image.
```
thehenrytsai, line 256: Similar to earlier comment: not sure if Dave not allowed to read Bob's reply/comment is common? Consider an easier-to-understand scenario so a code reader/reviewer doesn't have to inspect the protocol definition to confirm understanding. Maybe simple messaging between alice and bob?

tests/interfaces/records/handlers/records-read.spec.ts
```diff
@@ -135,6 +138,189 @@ describe('RecordsReadHandler.handle()', () => {
...
+      it('should allow read with allow-anyone rule', async () => {
+        // scenario: Alice writes an image to Caroline's DWN with Alice herself as recipient,
+        //           then Bob reads the image because he is "anyone".
+
+        const alice = await DidKeyResolver.generate();
+        const bob = await DidKeyResolver.generate();
+        const caroline = await DidKeyResolver.generate();
+
+        // setting up a stub DID resolver
+        TestStubGenerator.stubDidResolver(didResolver, [alice, bob, caroline]);
```
thehenrytsai, line 151: FYI: DID Key doesn't require stub resolver because it is self resolving.

tests/vectors/protocol-definitions/social-media.json
```diff
@@ -24,20 +24,28 @@
...
       "records": {
         "caption": {
           "allow": {
             "author": {
-                "of": "image",
-                "to": ["write"]
+              "of": "image",
+              "to": ["write"]
+            },
+            "recipient": {
```
thehenrytsai, line 37: Nit: like earlier comment, `recipient` here (and line 49) feels out of place, `anyone` seems to be more appropriate. If you need to test `recipient` rule, consider using/extending `dex.json` or `email.json` (coming in the other PR) for a more realistic (thus easier-to-understand) scenario.

src/core/protocol-authorization.ts

src/core/protocol-authorization.ts

src/core/protocol-authorization.ts
diehuxx: Added a commit refactoring to make an "honest" `constructAncestorMessageChain()`. Took your suggestion to move the query logic into `constructAncestorMessageChain()`. Hopefully it reads cleaner now

tests/interfaces/records/handlers/records-read.spec.ts
diehuxx: Refactored tests to use the email protocol when testing author and recipient reads. For the anyone test, I'm still using the socialMedia protocol in a simplified scenario since it doesn't make sense to have allow anyone to read emails.

tests/interfaces/records/handlers/records-read.spec.ts
diehuxx: Removed

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -1122,56 +1122,6 @@ describe('RecordsWriteHandler.handle()', () => {
         expect(reply.status.detail).to.contain(`no allow rule defined for Write`);
       });
 
-      it('should fail authorization if path to expected recipient in definition is longer than actual length of ancestor message chain', async () => {
```
diehuxx, line 1125: Removing this test because
1. Taking the example of `emailProtocol`, we allow `anyone: { "to": ["write"] }` and allow `author: { "of": "email", "to": ["read"] }`. When we write an email, we scan through (`anyone`, `author`, `recipient`). For the initial `write`, the `author` rule will have a longer expected path (`email`) than it has length of ancestor chain (length == 0). In this case, we shouldn't throw an error because this is a reasonable case. Instead, we should simply not add the `read` to allowed actions.
2. This test feels like its testing implementation more than expected behavior. I find it kinda confusing. If I were a developer using `dwn-sdk-js`, I don't think I would find it more helpful than an error saying `read is not an allowed action`.

src/core/protocol-authorization.ts
```diff
@@ -84,10 +102,31 @@ export class ProtocolAuthorization {
    * Constructs a chain of ancestor messages
    * @returns the ancestor chain of messages where the first element is the root of the chain; returns empty array if no parent is specified.
    */
-  private static async constructAncestorMessageChain(tenant: string, recordsWrite: RecordsWrite, messageStore: MessageStore)
+  private static async constructAncestorMessageChain(
```
thehenrytsai, line 105: Have been thinking, it's a bit odd to call the message chain "ancestor" chain in the context of a query. What if we rename to:

```suggestion
  private static async constructExistingMessageChain(
```

src/core/protocol-authorization.ts
```diff
@@ -84,10 +102,31 @@ export class ProtocolAuthorization {
...
+    if (incomingMessage.message.descriptor.method === DwnMethodName.Write) {
+      recordsWrite = incomingMessage as RecordsWrite;
+    } else {
+      const recordsRead = incomingMessage as RecordsRead;
+      const query = {
+        interface : DwnInterfaceName.Records,
+        method    : DwnMethodName.Write,
+        recordId  : recordsRead.message.descriptor.recordId,
+      };
+      const existingMessages = await messageStore.query(tenant, query) as TimestampedMessage[];
```
thehenrytsai, line 124: Seeing this, I can't help to think that this is a faster and simpler universal approach to do the chain construction regardless of method type? ie. one query that fetches all messages of the given recordId then construct the chain. What do you think?

tests/interfaces/records/handlers/records-read.spec.ts
```diff
@@ -135,6 +140,146 @@ describe('RecordsReadHandler.handle()', () => {
...
+        // Install social-media protocol on Alice's DWN
+        const protocolsConfig = await TestDataGenerator.generateProtocolsConfigure({
+          requester: alice,
+          protocol,
+          protocolDefinition
+        });
+        const protocolWriteReply = await dwn.processMessage(alice.did, protocolsConfig.message, protocolsConfig.dataStream);
+        expect(protocolWriteReply.status.code).to.equal(202);
+
+        // Alice writes image to Caroline's DWN with Alice herself as recipient
```
thehenrytsai, line 162:
```suggestion
        // Alice writes image to her DWN
```

src/core/protocol-authorization.ts
diehuxx: I like "ancestor" since I think of the records in protocol space as tree-like, where we are either writing to or reading from a node and we want to trace back up a list of parents. "Existing" message chain doesn't invoke the same idea of traversing upward to me.

src/core/protocol-authorization.ts
diehuxx: To get all the ancestor messages, we have to query for different records with different recordIds, since each ancestor is a separate `RecordsWrite`. We need to successively query by a different `parentId` for each message in the chain.

It's possible I'm misunderstanding your suggestion, but I don't think we can reduce down to one query by recordId.

mistermoe: only way i can think of to use a single query would be to query using `contextId` _only_. then you'd have to search the result-set for the record that has a `recordId === parentId` of incoming message and start building the ancestor chain from there. and i think you'd still have to query for the newest message for any given `recordId` i believe.

there may be more optimal ways that i'm sure we can get into later. for now i feel like this is very easy to read/follow which really helps think through whether it's correct

mistermoe: @diehuxx i'm wondering if we need to check whether a given record in the ancestral chain has been deleted?

csuwildcat: Brings up some good questions, namely: what happens to records that descend from deletes? I think there are a couple ways to look at it:

1. It should cause all descendants to be deleted.
2. Because of tombstoning, the chain is able to be retained. I'd personally favor #1, because app use cases generally expect it, and it's hard to reason about things if we don't keep it simple.

diehuxx: @csuwildcat Deleting all descendants seems too restrictive. For example, in a chat app with threads, if someone deletes on message in the thread, I wouldn't expect all subsequent messages to disappear. I know that Slack works this way: if one chat message disappears, the deleted message disappears or is replaced with "Message Deleted" and the rest of the thread is untouched. I've definitely seen other chat apps operate similarly.

Now I'm gonna devils advocate myself. The "Message Deleted" UI might be supportable by sending a `RecordsWrite` that overwrites the data to just say "Message Deleted" or something. In that case, the app developer could prohibit `RecordsDelete` for such records, so they wouldn't need to care about whether descendants get deleted.

So, it's possible that clever app developers use workarounds like this to "delete" single messages in the chain. But it still feels too restrictive, and IMO hard to reason about if descendants are always deleted.

Just thinking out loud: Protocol authz'd deletes aren't implemented yet. What if we had both `delete` and `delete-descendants` actions in protocols. Give developers both options.

mistermoe: @diehuxx and i chatted and we both feel like addressing this issue is out of scope for this PR. @diehuxx is going to create an issue for the topic brought up in this comment and we'll address it separately

mistermoe (approved):
💯

src/core/protocol-authorization.ts
diehuxx: Moving discussion to a separate issue as record deletion is out of scope for this PR
#311

PR #308, opened 2023-04-11
#276 - added encryption support to RecordsWrite

thehenrytsai:
There is A LOT of quality-of-life stuff that can be done to make the SDK much easier to use as follow ups, but this is nevertheless the first complete milestone that lights up the encryption feature.

Some features to be added in follow up PR(s) include:

1. Add `protocolPath` to every protocol-based `RecordsWrite`
1. Extend `protocol-context` derivation scheme further
1. Add more key derivation schemes

### DIFF
--- a/src/interfaces/records/handlers/records-query.ts
+++ b/src/interfaces/records/handlers/records-query.ts
@@ -1,11 +1,11 @@
 import type { MethodHandler } from '../../types.js';
-import type { BaseMessage, QueryResultEntry } from '../../../core/types.js';
 import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
-import type { RecordsQueryMessage, RecordsWriteMessage } from '../types.js';
+import type { RecordsQueryMessage, RecordsQueryReplyEntry, RecordsWriteMessage } from '../types.js';
 
 import { authenticate } from '../../../core/auth.js';
 import { lexicographicalCompare } from '../../../utils/string.js';
 import { MessageReply } from '../../../core/message-reply.js';
+import type { RecordsWriteMessageWithOptionalEncodedData } from '../../../store/storage-controller.js';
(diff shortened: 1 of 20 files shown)


src/interfaces/records/handlers/records-query.ts
```diff
@@ -152,12 +156,12 @@ export class RecordsQueryHandler implements MethodHandler {
...
-async function sortRecords(
-  entries: BaseMessage[],
+async function sortRecords<T extends RecordsWriteMessage>(
+  entries: T[],
   dateSort: DateSort
-): Promise<BaseMessage[]> {
+): Promise<T[]> {
 
-  const messages = entries as RecordsWriteMessage[];
+  const messages = entries;
```
diehuxx, line 164: nit: this line seems superfluous now

src/interfaces/records/handlers/records-query.ts
```diff
@@ -152,12 +156,12 @@ export class RecordsQueryHandler implements MethodHandler {
  * @param dateSort - Sorting scheme
  * @returns Sorted Messages
  */
-async function sortRecords(
-  entries: BaseMessage[],
+async function sortRecords<T extends RecordsWriteMessage>(
```
diehuxx, line 159: I don't necessarily have an argument against this, but it feels maybe overkill since at the moment we only use `sortRecords` in one place.

src/interfaces/records/handlers/records-query.ts
thehenrytsai: Good call!

src/interfaces/records/handlers/records-query.ts
thehenrytsai: Good call.

diehuxx (approved):
LGTM. Thanks for adding those utils, makes the tests easier to read

PR #313, opened 2023-04-16
`MessagesGet`

mistermoe:
Final PR to round out sync foundations. preceded by #305 

Introduces an `MessagesGet` message to retrieve messages.  Will include more details in this description tmrw. brain is total mush right now.


src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    const messages = await Promise.all(promises);
+
+    // for every message, include associated data as `encodedData` IF:
+    //  * its a RecordsWrite
+    //  * the data size is equal or smaller than the size threshold
+    //! NOTE: this is somewhat duplicate code that also exists in `StorageController.query`.
+    //! reviewers: consider whether we oughta abstract the common aspects.
+    for (const entry of messages) {
```
mistermoe, line 58: will refactor to use `Promise.all`

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+    const promises: Promise<MessagesGetReplyEntry>[] = [];
+
+    for (const messageCid of message.descriptor.messageCids) {
+      const promise = this.messageStore.get(tenant, messageCid)
+        .then(message => {
+          return { messageCid, message };
+        })
+        .catch(_ => {
+          // FIXIT: log error? throw error?
+          return { messageCid, message: undefined };
```
diehuxx, line 45: Do we have any error logging set up for DWNS? I'm not aware of it, but seems like a reasonable action to take for this error.

I don't mind leaving `message` empty, but also could dig adding an `error` field to `MessagesGetReplyEntry`. Throwing seems like like overkill.

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+      promises.push(promise);
+    }
+
+    const messages = await Promise.all(promises);
+
+    // for every message, include associated data as `encodedData` IF:
+    //  * its a RecordsWrite
+    //  * the data size is equal or smaller than the size threshold
+    //! NOTE: this is somewhat duplicate code that also exists in `StorageController.query`.
+    //! reviewers: consider whether we oughta abstract the common aspects.
```
diehuxx, line 57: I don't think this duplication is egregious that it _needs_ to be abstracted, but personally I would make a `StorageController#addEncodedData(message: BaseMessage): BaseMessage)` (or maybe call it `conditionallyAddEncodedData` idk) method that contains lines 65-81. Checks if should add `encodedData`, then adds it.

src/interfaces/messages/messages/messages-get.ts
```diff
@@ -0,0 +1,53 @@
...
+    const message = { descriptor, authorization };
+
+    Message.validateJsonSchema(message);
+    this.validateMessageCids(options.messageCids);
+
+    return new MessagesGet(message);
+  }
+
+  /**
+   * Q for reviewers: worth it?
```
diehuxx, line 40: Eh, might as well. Seems harmless. I might even go so far as to add a custom `format` to the JSON schema.

src/interfaces/messages/types.ts
```diff
@@ -0,0 +1,24 @@
+import type { BaseMessageReply } from '../../core/message-reply.js';
+import type { BaseMessage, Descriptor } from '../../core/types.js';
+import type { DwnInterfaceName, DwnMethodName } from '../../core/message.js';
+
+export type MessagesGetDescriptor = Pick<Descriptor, 'interface' | 'method' > & {
```
diehuxx, line 5: Just thinking out loud: I don't like `Pick`s very much. Would prefer to reduce the fields in `Descriptor` and move non-universal fields into message-specific descriptor types. But that is outside the scope of this PR.

mistermoe: I agree! Gave it a shot because it was mentioned in the last PR.

src/interfaces/messages/handlers/messages-get.ts
mistermoe: Re: adding `error` field

great idea @diehuxx! Will add today

src/interfaces/messages/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  method: DwnMethodName.Get;
+  messageCids: string[];
+};
+
+export type MessagesGetMessage = BaseMessage & {
+  descriptor: MessagesGetDescriptor;
+};
+
+// oof horrible name
+export type MessagesGetReplyEntry = {
```
diehuxx, line 16: Another idea for a refactor: I've never used [typescript namespaces](https://www.typescriptlang.org/docs/handbook/namespaces.html#namespaced-validators) before, but it might make these type names less ridiculous.

Like so:
```
namespace MessagesGet {
  export type ReplyEntry = {
     ...
  };

  export type Message {
     ...
  }
}
```

Again this idea is outside the scope of this PR, so I'm fine with this goofy variable name for now.

tests/interfaces/messages/handlers/messages-get.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    const alice = await DidKeyResolver.generate();
+    const { recordsWrite } = await TestDataGenerator.generateRecordsWrite({ requester: alice });
+    const recordsWriteMessageCid = await Message.getCid(recordsWrite.message);
+
+    const { message } = await TestDataGenerator.generateMessagesGet({
+      requester   : alice,
+      messageCids : [recordsWriteMessageCid]
+    });
+
+    // 0 messages expected because the RecordsWrite created above was never stored
```
diehuxx, line 173: wait, where does bob enter into this?

tests/interfaces/messages/handlers/messages-get.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    const reply = await dwn.processMessage(alice.did, recordsWrite.toJSON(), dataStream);
+    expect(reply.status.code).to.equal(202);
+
+    const recordsWriteMessageCid = await Message.getCid(recordsWrite.message);
+    const { message } = await TestDataGenerator.generateMessagesGet({
+      requester   : alice,
+      messageCids : [recordsWriteMessageCid]
+    });
+
+    // 1 messages expected because the RecordsWrite created above is not bob's
```
diehuxx, line 200: wait, where does bob enter into this?

diehuxx: oh I see a similar comment in the test below which does have a bob persona. I might just omit this comment

tests/interfaces/messages/messages/messages-get.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      try {
+        await MessagesGet.create({
+          authorizationSignatureInput : await Jws.createSignatureInput(alice),
+          messageCids                 : []
+        });
+
+        expect.fail();
+      } catch (e: any) {
+        expect(e.message).to.include('messageCids');
```
diehuxx, line 38: nit: include more of the expected error message here. Makes it easier to CMD+f what I broke if I ever break this test one day.

tests/interfaces/messages/messages/messages-get.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      try {
+        await MessagesGet.create({
+          authorizationSignatureInput : await Jws.createSignatureInput(alice),
+          messageCids                 : ['abcd']
+        });
+
+        expect.fail();
+      } catch (e: any) {
+        expect(e.message).to.include('valid CID');
```
diehuxx, line 53: Same as above. More of error message pls.

tests/interfaces/messages/messages/messages-get.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+    });
+  });
+
+  describe('parse', () => {
+    it('parses a message into an MessagesGet instance', async () => {
```
diehuxx, line 59:
```suggestion
    it('parses a message into a MessagesGet instance', async () => {
```

diehuxx (commented):
looking solid! no major issues, just an assortment of small things.

tests/interfaces/messages/messages/messages-get.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      const message = messagesGet.toJSON() as MessagesGetMessage;
+      message.descriptor.messageCids = ['abcd'];
+
+      try {
+        await MessagesGet.parse(message);
+
+        expect.fail();
+      } catch (e: any) {
+        expect(e.message).to.include('valid CID');
```
diehuxx, line 94: Same as above. More of error message pls.

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    const promises: Promise<MessagesGetReplyEntry>[] = [];
+
+    for (const messageCid of message.descriptor.messageCids) {
+      const promise = this.messageStore.get(tenant, messageCid)
+        .then(message => {
+          return { messageCid, message };
+        })
+        .catch(_ => {
```
dcrousso, line 43: i dont think `MessageStore` will ever throw (unless the operation is aborted).  you probably should check for `message === undefined` instead

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+          // FIXIT: log error? throw error?
+          return { messageCid, message: undefined };
+        });
+
+      promises.push(promise);
+    }
+
+    const messages = await Promise.all(promises);
```
dcrousso, line 51: NIT: instead of creating a `promises` array you could do this with `Array.prototype.map`

and also at the same time you could merge the `encodedData` part so that we dont iterate over all the `messages` twice

```
const messages = await Promise.all(message.descriptor.messageCids.map(async messageCid => {
  const message = await this.messageStore.get(tenant, messageCid);
  if (!message) {
    return { messageCid, error }; // FIXME
  }

  if (message.descriptor.interface === DwnInterfaceName.Records && message.descriptor.method === DwnMethodName.Write) {
    const dataCid = message.descriptor.dataCid;
    const dataSize = message.descriptor.dataSize;
    if (dataCid !== undefined && dataSize! <= DwnConstant.maxDataSizeAllowedToBeEncoded) {
      const { dataStream } = await dataStore.get(tenant, messageCid, dataCid) ?? { };
      if (dataStream) {
        const dataBytes = await DataStream.toBytes(dataStream);
        message.encodedData = Encoder.bytesToBase64Url(dataBytes);
      }
    }
  }

  return { messageCid, message };
});
```

src/interfaces/messages/messages/messages-get.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type MessagesGetOptions = {
+  messageCids: string[];
+  authorizationSignatureInput: SignatureInput;
+};
+
+export class MessagesGet extends Message<MessagesGetMessage> {
+  public static async parse(message: MessagesGetMessage): Promise<MessagesGet> {
+    Message.validateJsonSchema(message);
+    this.validateMessageCids(message.descriptor.messageCids);
```
dcrousso, line 16: NIT: i personally avoid using `this` in a `static` context as it can be confusing at a glance (i.e. only use `this` to represent an instance, not the `class`).  id suggest using `MessagesGet.validateMessageCids` instead

src/interfaces/messages/messages/messages-get.ts
dcrousso: i dont think we do this kind of validation anywhere else (e.g. `dataCid`) so i think it's probably fine to not do it

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -47,6 +47,7 @@ export class RecordsDeleteHandler implements MethodHandler {
     const newestExistingMessage = await RecordsWrite.getNewestMessage(existingMessages);
     let incomingMessageIsNewest = false;
     let newestMessage;
+
```
dcrousso, line 50: oops?

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,89 @@
...
+    try {
+      await authenticate(message.authorization, this.didResolver);
+      await authorize(tenant, messagesGet);
+    } catch (e) {
+      return MessageReply.fromError(e, 401);
+    }
+
+    const promises: Promise<MessagesGetReplyEntry>[] = [];
+
+    for (const messageCid of message.descriptor.messageCids) {
```
dcrousso, line 38: should we attempt to do any deduplication here in case the caller provided the same `messageCid` more than once?

src/interfaces/messages/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  messageCid: string;
```
dcrousso, line 17: i wonder if we should return a `Map`/`Object` where the keys are the `messageCid` and the values are the `{ message, encodedData }` so that the caller can:

- ensure they will never encounter the same `messageCid` more than once (since keys are unique)
- quickly grab the `message`/`encodedData` for a given `messageCid` instead of having to iterate the entire array to find the entry matching the `messageCid`
- possibly even avoid storing the `messageCid` (at least without having to manipulate the data) if they wanted to do something like `savedMessages.push(Object.values(result))`

dcrousso (dismissed):

src/interfaces/messages/handlers/messages-get.ts
michaelneale: a promise to use a promise

src/interfaces/messages/handlers/messages-get.ts
diehuxx: `message-store-level` should not throw, but it's possible a custom message store will. In that case, it's useful to `catch`. Also, don't think we should throw/log error on `message === undefined`. That's a non-exceptional case IMO, and the client should be okay with receiving `message: undefined` in the reply entry.

src/interfaces/messages/messages/messages-get.ts
mistermoe: reason behind doing so was to prevent unnecessary i/o for bunk CIDs

src/interfaces/messages/handlers/messages-get.ts
mistermoe: good idea. done

src/interfaces/records/handlers/records-delete.ts
mistermoe: yeah instinctive fat pinkied. fixed

src/interfaces/messages/handlers/messages-get.ts
mistermoe: > `message-store-level` should not throw, but it's possible a custom message store will. In that case, it's useful to `catch`.  Also, don't think we should throw/log error on `message === undefined`. That's a non-exceptional case IMO, and the client should be okay with receiving `message: undefined` in the reply entry.

💯

tests/interfaces/messages/messages/messages-get.spec.ts
mistermoe: good call. this one specifically is auto-generated by AJV, our json schema parser/validator. still went ahead and included fuller error messages for all tests. In this case, hopefully the fuller error message will maaaybe help indicate that it's coming from json schema. went ahead and called that out in a comment

src/interfaces/messages/types.ts
diehuxx: > possibly even avoid storing the `messageCid` (at least without having to manipulate the data) if they wanted to do something like `savedMessages.push(Object.values(result))`

Not sure what you mean by this

> quickly grab the message/encodedData for a given messageCid instead of having to iterate the entire array to find the entry matching the messageCid

In practice, I think clients will generally be iterating over the array so they can process messages in order, but you're right that it may benefit the odd client. If we use an object with `messageCid`s as keys, we don't return an ordered list. [Looks like some JSON serializers don't preserve key order](https://filemakerhacks.com/2020/04/29/ordering-elements-within-json-objects/#:~:text=The%20JSON%20Data%20Interchange%20Standard,%2C%20and%20should%20not%2C%20matter.) That might not be a big deal; the client could remember the list of `messageCids` they sent in the request and use that to determine the order of messages to process. Not arguing against your approach, just something to think about as we put sync in production.

src/interfaces/messages/handlers/messages-get.ts
mistermoe: @diehuxx  for error message: `Failed to get message`?

src/interfaces/messages/messages/messages-get.ts
mistermoe: @dcrousso imo `dataCid` is validated when its integrity checked [here](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/storage-controller.ts#L70-L81).

mistermoe: @diehuxx definitely considered adding a custom `format` to JSON schema but decided against it. Trying to keep use of json schema as language agnostic as possible in hopes that they can be used if anyone takes a stab at implementing any part of DWNs in another language

src/interfaces/messages/handlers/messages-get.ts
diehuxx: @mistermoe worx 4 me

diehuxx (dismissed):

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -62,7 +63,9 @@ export class MessagesGetHandler implements MethodHandler {
       }
 
       const { interface: messageInterface, method } = message.descriptor;
-      if (messageInterface !== DwnInterfaceName.Records || method !== DwnMethodName.Write) {
+      const interfaceMethod = `${messageInterface}${method}`;
+
+      if (interfaceMethod !== RecordsWrite) {
```
dcrousso: im confused.  what was wrong with the old logic?

mistermoe: just feel like it's easier to read

diehuxx: @mistermoe ngl disagree, but not gonna stop you.

having a var named `RecordsWrite` is confusing cuz dwn-sdk-js has a class named `RecordsWrite` already.
if we really wanna make this more readable, the ruby-ish thing that I would do is define a bunch of methods like `isRecordsWrite(message)`. might be overkill, but i tend to like that specific type of overkill.

dcrousso: ah.  personally i find this less easy to read, especially since there's already a `class RecordsWrite` elsewhere in this repo (tho we could have a `RecordsWrite.Identifier` or something?) 😕

perhaps we could flip this around and have it be `if (interface === DwnInterfaceName.Records && method === DwnMethodName.Write)` (or put a `if (!(...))` if you wanna keep it with `continue`)?

mistermoe: the crowd has spoken! reverting

mistermoe: what are y'all talking about? i didn't change anything 🥸

src/dwn.ts
```diff
@@ -40,6 +41,7 @@ export class Dwn {
 
     this.methodHandlers = {
       [DwnInterfaceName.Events + DwnMethodName.Get]          : new EventsGetHandler(this.didResolver, this.eventLog),
+      [DwnInterfaceName.Messages + DwnMethodName.Get]        : new MessagesGetHandler(this.didResolver, this.messageStore, this.dataStore),
```
thehenrytsai, line 44: Sanity: no explicit handler method? Was there more discussions on this while I was away?

src/interfaces/messages/types.ts
thehenrytsai, line 5: For my understanding: what is the intent of Picking `interface` and `method` when line 6 and 7 specifies it again?

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -0,0 +1,88 @@
...
+    const promises: Promise<MessagesGetReplyEntry>[] = [];
+    const messageCids = new Set(message.descriptor.messageCids);
+
+    for (const messageCid of messageCids) {
+      const promise = this.messageStore.get(tenant, messageCid)
+        .then(message => {
+          return { messageCid, message };
+        })
+        .catch(_ => {
+          return { messageCid, message: undefined, error: `Failed to get message ${messageCid}` };
```
thehenrytsai, line 45: No test for this line.

tests/interfaces/messages/handlers/messages-get.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+      expect(messageReply.messageCid).to.not.be.undefined;
+      expect(messageReply.message).to.not.be.undefined;
+      expect(messageCids).to.include(messageReply.messageCid);
+
+      const cid = await Message.getCid(messageReply.message!);
+      expect(messageReply.messageCid).to.equal(cid);
+    }
+  });
+
+  it('returns message as undefined in reply entry when a messageCid is not found', async () => {
```
thehenrytsai, line 163: Sanity: is this meant to be the test that tests the line above that is never hit?

src/interfaces/messages/types.ts
mistermoe: removed `Pick` usage

src/dwn.ts
mistermoe: nothing to the contrary was discussed while you were gone. went ahead and added the explicit `handleMessagesGet` method to `dwn`

tests/interfaces/messages/handlers/messages-get.spec.ts
mistermoe: na. this test is not supposed to hit the error code path

src/interfaces/messages/handlers/messages-get.ts
mistermoe: added

thehenrytsai (approved):
Good to go! 🙇

Issue #265, opened 2023-03-16
normalize `schema` and `protocol` properties in `descriptor`


thehenrytsai:
toLower the string to prevent segregation when filtering by those properties

thehenrytsai:
I believe this is done with #317 and #325.

PR #314, opened 2023-04-18
Normalize protocol URIs

diehuxx:
Resolves part of #265. PR to normalize schemas will come next. Thanks @dcrousso for pointing me in the right direction!

The general approach is: Anywhere we call `MessageStore#put` or `MessageStore#query` normalize the protocol string in the indexes/filters respective.

### DIFF
--- a/src/interfaces/protocols/handlers/protocols-configure.ts
+++ b/src/interfaces/protocols/handlers/protocols-configure.ts
@@ -57,8 +57,8 @@ export class ProtocolsConfigureHandler implements MethodHandler {
     if (incomingMessageIsNewest) {
       const { author } = protocolsConfigure;
       const indexes = {
-        author,
         ... message.descriptor,
+        author,
         protocol: normalizeProtocolUrl(message.descriptor.protocol),
       };
 
--- a/src/interfaces/protocols/handlers/protocols-configure.ts
+++ b/src/interfaces/protocols/handlers/protocols-configure.ts
@@ -1,10 +1,12 @@
 import type { EventLog } from '../../../event-log/event-log.js';
+import type { Filter } from '../../../core/types.js';
 import type { MethodHandler } from '../../types.js';
 import type { ProtocolsConfigureMessage } from '../types.js';
 import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
 
 import { canonicalAuth } from '../../../core/auth.js';
 import { MessageReply } from '../../../core/message-reply.js';
+import { normalizeProtocolUrl } from '../../../utils/url.js';
 import { ProtocolsConfigure } from '../messages/protocols-configure.js';
 import { StorageController } from '../../../store/storage-controller.js';
 
@@ -35,10 +37,10 @@ export class ProtocolsConfigureHandler implements MethodHandler {
     }
 
     // attempt to get existing protocol
-    const query = {
+    const query: Filter = {
       interface : DwnInterfaceName.Protocols,
       method    : DwnMethodName.Configure,
-      protocol  : message.descriptor.protocol
+      protocol  : normalizeProtocolUrl(message.descriptor.protocol)
     };
     const existingMessages = await this.messageStore.query(tenant, query) as ProtocolsConfigureMessage[];
 
@@ -56,7 +58,8 @@ export class ProtocolsConfigureHandler implements MethodHandler {
       const { author } = protocolsConfigure;
       const indexes = {
         author,
-        ... message.descriptor
+        ... message.descriptor,
+        protocol: normalizeProtocolUrl(message.descriptor.protocol),
       };
 
       // FIXME: indexes, Property 'dataSize' is incompatible with index signature.
--- a/src/interfaces/protocols/handlers/protocols-query.ts
+++ b/src/interfaces/protocols/handlers/protocols-query.ts
@@ -1,13 +1,13 @@
 import type { MethodHandler } from '../../types.js';
 import type { ProtocolsQueryMessage } from '../types.js';
-import type { QueryResultEntry } from '../../../core/types.js';
 import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
+import type { Filter, QueryResultEntry } from '../../../core/types.js';
 
 import { canonicalAuth } from '../../../core/auth.js';
 import { MessageReply } from '../../../core/message-reply.js';
+import { normalizeProtocolUrl } from '../../../utils/url.js';
 import { ProtocolsQuery } from '../messages/protocols-query.js';
 import { removeUndefinedProperties } from '../../../utils/object.js';
-
 import { DwnInterfaceName, DwnMethodName } from '../../../core/message.js';
 
 export class ProtocolsQueryHandler implements MethodHandler {
@@ -32,11 +32,14 @@ export class ProtocolsQueryHandler implements MethodHandler {
       return MessageReply.fromError(e, 401);
     }
 
-    const query = {
+    const query: Filter = {
       interface : DwnInterfaceName.Protocols,
       method    : DwnMethodName.Configure,
-      ...message.descriptor.filter
     };
+    const protocol = message.descriptor.filter?.protocol;
+    if (protocol !== undefined) {
+      query.protocol = normalizeProtocolUrl(protocol);
+    }
     removeUndefinedProperties(query);
 
     const messages = await this.messageStore.query(tenant, query);
--- a/src/interfaces/protocols/handlers/protocols-query.ts
+++ b/src/interfaces/protocols/handlers/protocols-query.ts
@@ -35,10 +35,10 @@ export class ProtocolsQueryHandler implements MethodHandler {
     const query: Filter = {
       interface : DwnInterfaceName.Protocols,
       method    : DwnMethodName.Configure,
+      ...message.descriptor.filter
     };
-    const protocol = message.descriptor.filter?.protocol;
-    if (protocol !== undefined) {
-      query.protocol = normalizeProtocolUrl(protocol);
+    if (query.protocol !== undefined) {
+      query.protocol = normalizeProtocolUrl(query.protocol as string);
     }
     removeUndefinedProperties(query);
 
--- a/src/interfaces/records/handlers/records-write.ts
+++ b/src/interfaces/records/handlers/records-write.ts
@@ -121,8 +121,8 @@ export async function constructRecordsWriteIndexes(
     recordId  : message.recordId,
     entryId   : await RecordsWrite.getEntryId(recordsWrite.author, recordsWrite.message.descriptor)
   };
-  if (descriptor.protocol !== undefined) {
-    indexes.protocol = normalizeProtocolUrl(descriptor.protocol);
+  if (indexes.protocol !== undefined) {
+    indexes.protocol = normalizeProtocolUrl(indexes.protocol);
   }
 
   // add additional indexes to optional values if given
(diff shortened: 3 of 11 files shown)


src/interfaces/protocols/handlers/protocols-configure.ts
```diff
@@ -56,7 +58,8 @@ export class ProtocolsConfigureHandler implements MethodHandler {
       const { author } = protocolsConfigure;
       const indexes = {
         author,
-        ... message.descriptor
+        ... message.descriptor,
```
dcrousso, line 61: NIT: we probably could/should move this to be before `author` just in case it's inside `message.descriptor` so that it's not overridden by the `...`

src/interfaces/protocols/handlers/protocols-query.ts
```diff
@@ -32,11 +32,14 @@ export class ProtocolsQueryHandler implements MethodHandler {
       return MessageReply.fromError(e, 401);
     }
 
-    const query = {
+    const query: Filter = {
       interface : DwnInterfaceName.Protocols,
       method    : DwnMethodName.Configure,
-      ...message.descriptor.filter
```
dcrousso, line 38: NIT: i recognize that there's no need for this anymore since the only value in `filter` is `protocol`, but some part of me feels like we should still leave this here just for future proofing sake 🤔

src/interfaces/records/handlers/records-write.ts
```diff
@@ -120,6 +121,9 @@ export async function constructRecordsWriteIndexes(
     recordId  : message.recordId,
     entryId   : await RecordsWrite.getEntryId(recordsWrite.author, recordsWrite.message.descriptor)
   };
+  if (descriptor.protocol !== undefined) {
```
dcrousso, line 124: NIT: rather than check against the `descriptor.protocol`, should we check `indexes.protocol` since ultimately we really only care about normalizing the `protocol` value in the `IndexLevel`?  some part of me feels like it's odd to have it be added to `indexes` only to then check some other place before overriding it

src/interfaces/records/messages/records-query.ts
```diff
@@ -92,6 +93,10 @@ export class RecordsQuery extends Message<RecordsQueryMessage> {
       (filterCopy as Filter).dateCreated = rangeFilter;
     }
 
+    if (filter.protocol !== undefined) {
```
dcrousso, line 96: ditto

src/utils/url.ts
```diff
@@ -0,0 +1,13 @@
+export function normalizeProtocolUrl(url: string): string {
+  const fullUrl = url.includes('://') ? url : `http://${url}`;
```
dcrousso, line 2: what if the URL contains `://` in the path/query/etc.?  perhaps we should `/^[^:]+:\/\/./.test(url)` instead?

src/utils/url.ts
```diff
@@ -0,0 +1,13 @@
...
+  const { hostname, pathname } = new URL(fullUrl);
```
dcrousso, line 3: why not include the `search` as well?  i feel like that's a pretty commonly used way of distinguishing resources on many webservers

NIT: we may want to `try { ... } catch { }` this just in case

tests/interfaces/protocols/handlers/protocols-query.spec.ts
```diff
@@ -93,6 +100,88 @@ describe('ProtocolsQueryHandler.handle()', () => {
...
+      // Sort messages into lexicographic order. We only allow protocol overwrites if the new message CID
+      // has higher lexicographic value.
+      const equivalentProtocols = [protocol1, protocolTrailingSlash, protocolParams, protocolCapitalized];
+      let messageDataWithCid: (GenerateProtocolsConfigureOutput & { cid: string })[] = [];
+      for (const messageData of equivalentProtocols) {
+        const cid = await Message.getCid(messageData.message);
+        messageDataWithCid.push({ cid, ...messageData });
+      }
+
+      messageDataWithCid = messageDataWithCid.sort((messageDataA, messageDataB) => {
```
dcrousso, line 141: `Array.prototype.sort` is in-place, so no need to re-assign it :)

tests/interfaces/protocols/handlers/protocols-query.spec.ts
```diff
@@ -93,6 +100,88 @@ describe('ProtocolsQueryHandler.handle()', () => {
...
+
+      // Configure protocols. Each one overwrites the previous
+      for (const protocol of messageDataWithCid) {
+        const configureReply = await dwn.processMessage(alice.did, protocol.message, protocol.dataStream);
+        expect(configureReply.status.code).to.equal(202);
+      }
+
+      // configure several more protocols, none of which match `example.com`
+      const protocolSlashFoo = await TestDataGenerator.generateProtocolsConfigure({ requester: alice, protocol: 'example.com/foo' });
+      const protocolSubdomainFoo = await TestDataGenerator.generateProtocolsConfigure({ requester: alice, protocol: 'foo.example.Com' });
```
dcrousso, line 153: is the `Com` a typo or intentional?  if the latter, can you explain why?

tests/interfaces/protocols/handlers/protocols-query.spec.ts
```diff
@@ -93,6 +100,88 @@ describe('ProtocolsQueryHandler.handle()', () => {
...
+        protocolDefinition
+      });
+      const protocolCapitalized = await TestDataGenerator.generateProtocolsConfigure({
+        requester : alice,
+        protocol  : 'EXAMPLE.Com',
+        protocolDefinition
+      });
+
+      // Sort messages into lexicographic order. We only allow protocol overwrites if the new message CID
+      // has higher lexicographic value.
```
dcrousso, line 133: wow really?  that seems like an odd way of restricting things.

we dont sort based on the `dateModified` or anything like that?

tests/interfaces/protocols/handlers/protocols-query.spec.ts
```diff
@@ -93,6 +100,88 @@ describe('ProtocolsQueryHandler.handle()', () => {
...
+      expect(reply.status.code).to.equal(200);
+      expect(reply.entries?.length).to.equal(1);
+
+      const resultProtocols = reply.entries!.map(entry =>
+        (entry.descriptor as ProtocolsConfigureDescriptor).protocol
+      );
+
+      // Expect equivalent protocol with highest lexicographic value to remain
+      expect(resultProtocols).to.contain(
+        messageDataWithCid[messageDataWithCid.length-1].message.descriptor.protocol
```
dcrousso, line 177: Style: i think there should be spaces around the `-`

tests/interfaces/records/handlers/records-query.spec.ts
```diff
@@ -112,6 +112,65 @@ describe('RecordsQueryHandler.handle()', () => {
...
+        });
+        const writeReply = await dwn.processMessage(alice.did, write.message, write.dataStream);
+        expect(writeReply.status.code).to.equal(202);
+      }
+
+      // Query for matching protocol
+      const messageData = await TestDataGenerator.generateRecordsQuery({
+        requester : alice,
+        filter    : {
+          protocol: protocol1
```
dcrousso, line 153: should we also do a query for `protocol2` to make sure that it also matches the same entries?

dcrousso (approved):

tests/interfaces/protocols/handlers/protocols-query.spec.ts
diehuxx: Yea I don't understanding why. @csuwildcat Why do we do this? I know I pinged in Slack, but asking here too so we can get an opened source'd explanation.

src/utils/url.ts
diehuxx: Good call on try/catch. will add that today.

Excluding `search` because there is some conversation lately around how to configure different "roles" of a protocol. For example, for TBDex a DWN may want to set up as an "issuer" or as a "PFI" or a "wallet". Each of those roles would use the TBDex protocol a bit differently. The current idea that has been tossed around DWN office hours is to have that "role" be specified in a query param. So they should be normalized out.

This is a very very WIP idea, so we may end up changing it in the coming months (or even weeks), so if keeping `search` ends up being the right approach, I'm open to changing it later.

dcrousso: hmmm if that's the case then how would someone query for a specific "role" of the same protocol?

since we're always normalizing the `protocol` anywhere we send it to `MessageStore` (and therefore `IndexLevel`), would it be up to the caller to do their own filtering?

dcrousso: ah sorry i shouldve been clearer.  i was specifically referring to using `RecordsQuery` to something like

> get all messages with `{ protocol: 'example.com?role=issuer' }`

AFAICT what you currently have in this PR would return everything matching `{ protocol: 'example.com' }`, which could include other things (e.g. `{ protocol: 'example.com?role=other' }`) that the caller might not expect

we may want to make the normalization even more minimal than what's there now, such as only

- always having a `protocol` of `http://`
- removing trailing `'/'` from the `pathname`
- removing `port`
- removing `password`
- sorting the `search`
- removing `hash`

IMO this is part of why i dont think it's a good idea to do normalization, as we're effectively constraining the possibilities of URLs without any sort of specification as to what those constraints are (e.g. it's totally valid for a different port to return a different result).  i understand that we're trying to make the developer experience better by trying to avoid (possibly) common mistakes (e.g. adding a trailing `'/'`), but i think that could just result in developers being more confused by seemingly magical stuff happening instead of exactly what they asked for

---

as far as what you said, i think so long as `example.com?role=issuer` and `example.com?role=pfi` resolve to different things then that's fine :)

tests/interfaces/protocols/handlers/protocols-query.spec.ts
diehuxx: typo. fixed

tests/interfaces/records/handlers/records-query.spec.ts
diehuxx: good idea. added

src/utils/url.ts
diehuxx: I'm starting to agree with you as I jam on edge cases. Lemme think about this a bit

tests/interfaces/protocols/handlers/protocols-query.spec.ts
thehenrytsai: @dcrousso and @diehuxx, the comment is not entirely correct. We only resort to lexicographical order if all messages for the same ID are written with the exact same timestamp, which is the case here in the context of testing, but not a commonly expected scenario in the real-world.

thehenrytsai (commented):
Seeing all the places that had `normalizeProtocolUrl()` inserted, I am wondering if it is better/worse to use prevention as a strategy to keep the internals clean? ie. do the normalization in `create()` methods and perform message validation upon request...

diehuxx:
> do the normalization in create() methods and perform message validation upon request...

Not sure what you mean by perform message validation upon request?

tests/interfaces/protocols/handlers/protocols-query.spec.ts
diehuxx: Ahh okay that makes more sense. Was very confused. Thanks!

src/interfaces/protocols/handlers/protocols-query.ts
diehuxx: added back in

PR #317, opened 2023-04-24
Reject messages with nonnormalize protocol URIs

diehuxx:
Addresses part of https://github.com/TBD54566975/dwn-sdk-js/issues/265
Revises a previous attempt #314 

## Changes
To ensure that protocols URIs are bucketed intuitively,
1. `#create` methods auto-normalize protocols for an ergonomic developer experience
2. non-normalized protocols are rejected from DWNs

## TODO
Going to add a little more test coverage in the morning to bring relevant files back up to 100%

### DIFF
--- a/src/interfaces/records/messages/records-write.ts
+++ b/src/interfaces/records/messages/records-write.ts
@@ -24,11 +24,11 @@ import { ProtocolAuthorization } from '../../../core/protocol-authorization.js';
 import { Records } from '../../../utils/records.js';
 import { removeUndefinedProperties } from '../../../utils/object.js';
 import { Secp256k1 } from '../../../utils/secp256k1.js';
-
 import { authorize, validateAuthorizationIntegrity } from '../../../core/auth.js';
 import { Cid, computeCid } from '../../../utils/cid.js';
 import { DwnError, DwnErrorCode } from '../../../core/dwn-error.js';
 import { DwnInterfaceName, DwnMethodName } from '../../../core/message.js';
+import { normalizeProtocolUri, validateProtocolUriNormalized } from '../../../utils/url.js';
 
 export type RecordsWriteOptions = {
   recipient?: string;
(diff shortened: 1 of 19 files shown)


src/utils/url.ts
```diff
@@ -0,0 +1,42 @@
+import { DwnError, DwnErrorCode } from '../core/dwn-error.js';
+
+const URL_PROTOCOL_REGEX = /^[^:]+:\/\/./;
```
thehenrytsai, line 3: For my understanding: why this particular regex? Is this something you created or is it from some well-known regex for URIs? It just looks too lenient. e.g. this string `??://asdf.` would pass as valid...

thehenrytsai: Oh, I see. you are only checking the "scheme" portion later in code. I thought "protocol" here meant DWN protocol for some reason. I'd suggest to encapsulate the regex within `normalizeProtocolUri(...)` as reusability is probably unlikely.

src/utils/url.ts
```diff
@@ -0,0 +1,42 @@
...
+    fullUrl = `http://${url}`;
+  }
+
+  try {
+    const result = new URL(fullUrl);
+    result.search = '';
+    result.hash = '';
+    return removeTrailingSlash(result.href);
+  } catch (e) {
+    throw new Error('Could not normalize protocol URI');
```
thehenrytsai, line 32: Throw `DwnError` for all new error thrown going forward.

src/interfaces/records/messages/records-write.ts
```diff
@@ -119,6 +119,10 @@ export class RecordsWrite extends Message<RecordsWriteMessage> {
     await validateAuthorizationIntegrity(message, { allowedProperties: new Set(['recordId', 'contextId', 'attestationCid', 'encryptionCid']) });
     await RecordsWrite.validateAttestationIntegrity(message);
 
+    if (message.descriptor.protocol !== undefined) {
+      validateProtocolUriNormalized(message.descriptor.protocol);
+    }
```
thehenrytsai, line 124: Try putting the check within `validateIntegrity()`, that's the intent of that method. Not picking on other methods since the checks you introduced are the "first check", if we have more than 1 check in the future we'd follow the same pattern also.

thehenrytsai (dismissed):

thehenrytsai (approved):
🙏

Issue #310, opened 2023-04-11
Add `protocolPath` to protocol-based RecordsWrite


thehenrytsai:
This will:
1. Enable efficient implementation of key derivation scheme based on protocol path for DWN encryption
2. Improve dev experience because the protocol path will be part of the message returned
3. Potentially simplify existing implementation of protocol-based auth

PR #318, opened 2023-04-24
#310 - added protocolPath validation

thehenrytsai:


src/core/protocol-authorization.ts
```diff
@@ -199,6 +199,43 @@ export class ProtocolAuthorization {
...
+    inboundMessage: RecordsRead | RecordsWrite,
+    ancestorMessageChain: RecordsWriteMessage[],
+    recordSchemaToLabelMap: Map<string, string>
+  ): void {
+    // skip verification if this is not a RecordsWrite
+    if (inboundMessage.message.descriptor.method !== DwnMethodName.Write) {
+      return;
+    }
+
+    const currentRecordSchema = inboundMessage.message.descriptor.schema!;
```
diehuxx, line 216: Nit-ish: Not sure it's safe to say that `schema` is definitely defined. I just tested sending a RecordsWrite with undefined schema and was able to reach this line of code. `recordSchemaToLabelMap.get(currentRecordSchema)` handles undefined input without throwing, so this line runs, but if we want honest typescript, maybe we should explicitly check for undefined.

diehuxx (approved):
one nit, other lgtm

tests/validation/json-schemas/records/records-write.spec.ts
```diff
@@ -123,6 +123,34 @@ describe('RecordsWrite schema definition', () => {
     }).throws('must NOT have additional properties');
   });
 
+  it('should pass if `protocol` exists and its related properties are all present', () => {
+    const invalidMessage = {
```
diehuxx, line 127: naming this `invalidMessage` is a bit confusing because the test name and inline comments make it seems like this should be a valid message. Do you mean to test that this message will throw if one of the protocol-related fields is missing?

src/core/protocol-authorization.ts
thehenrytsai: Super keen eyes! Was hoping to slip this through under the radar. You are right, there was no enforcement for schema to be there for protocol-based records but there ought to be. I've updated the schema and tests to enforce so.

Was unsure if Daniel wants to allow a non-schema records to be able to written using DWN protocol and what the implication is. I am obviously not in favor of it simply because it invalidates a lot of our protocol authorization code due to its assumptions and it's likely a longer discussion than we want right now. So I will try to let this fly under the radar again by just go ahead and add the check for `schema` (notice I didn't actually tag Daniel). = ]

diehuxx: > (notice I didn't actually tag Daniel). = ]

absolute anarchy. corporate espionage in action. i'd have it no other way

diehuxx (approved):

Issue #306, opened 2023-04-10
Add `dataFormat` awareness/filtering to Protocols


csuwildcat:
Currently we aren't/limiting bucketing using the `dataFormat` property in protocol object definitions, and should be. Let's add that filter/awareness where needed.

thehenrytsai:
#320 addressed this.

PR #320, opened 2023-04-25
#306 Validate data format on protocol authd writes

diehuxx:
Fully addresses #306

* Add `dataFormats` to protocol definition JSON schema
* Validate that incoming RecordsWrites are within the allowed `dataFormats`

### DIFF
--- a/src/interfaces/protocols/types.ts
+++ b/src/interfaces/protocols/types.ts
@@ -11,7 +11,10 @@ export type ProtocolsConfigureDescriptor = {
 
 export type ProtocolDefinition = {
   labels: {
-    [key: string]: { schema: string };
+    [key: string]: {
+      schema: string,
+      dataFormats?: string[],
+    };
   };
   records: {
     [key: string]: ProtocolRuleSet;
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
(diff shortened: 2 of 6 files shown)


csuwildcat:
Niiice!

src/interfaces/protocols/types.ts
```diff
@@ -11,7 +11,10 @@ export type ProtocolsConfigureDescriptor = {
 
 export type ProtocolDefinition = {
   labels: {
```
thehenrytsai, line 13: This is non-blocking, just discussion: calling this `labels` now feels like an over-simplification, maybe `constraints` is a better name.

thehenrytsai (approved):
🙏 
Looks good, just a non-blocking comment.

src/interfaces/protocols/types.ts
diehuxx: Agreed, especially as we add more to `labels`. I'm open to changing this field (though not in the PR preferably).

`constraints` doesn't feel right either. `constraints` implies to me that we only need to list a label/type/name/whatever here if we have specific restrictions to put on it like `dataFormat`. But in fact we need to include all labels/types/names/whatevers in this property if we want them to be valid in the `record` section.

Spitballing some more ideas:
* `recordTypes`
* `types`
* `models`
* `tables`

csuwildcat:
Labels could be definitions or constrains, and if so, we'd need to change it right away.

PR #150, opened 2022-11-11
Added date regex

rajakash-dev:

Issue #147, opened 2022-11-08
Add regex in JSON schema to check date properties in ISO 8601 UTC format 


thehenrytsai:
See date format requirements in #113.

The regex should be added for `dateCreated` and `datePublished` properties.

rajakash-dev:
Can you please elaborate on how to use the regex , i have built and tested a regex exp but not sure how to use it in this use case.

thehenrytsai:
You can see regex examples in https://github.com/TBD54566975/dwn-sdk-js/blob/ecd72c95e4ab31f68dbd74d36bc3502967d163da/json-schemas/definitions.json used for validations.

rajakash-dev:
https://github.com/TBD54566975/dwn-sdk-js/pull/150

duncanmak:
We can close this now!

thehenrytsai:
Yes we can!

PR #322, opened 2023-04-25
Specify format: 'date-time' for date properties

duncanmak:
Addresses #147.

According to https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times, there's already a predefined format called `date-time` for the ISO-8601 format in JSON Schema.

The concrete definition is [here](https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L104):

```javascript
"date-time": fmtDef(
    /^\d\d\d\d-[0-1]\d-[0-3]\dt(?:[0-2]\d:[0-5]\d:[0-5]\d|23:59:60)(?:\.\d+)?(?:z|[+-]\d\d(?::?\d\d)?)$/i,
    compareDateTime
  ),
```

These formats are part of the `ajv-formats` package, so I've added it via npm. If we want to avoid installing an entire package, we could also add a single format using the `addFormat` method in our own `compile-validators.js`.

~~While` I'm at it, I also bumped the ajv package version, I could back that out if we don't want to bundle it in.~~

Having reviewed #150, I hope that by using a regex defined elsewhere, that this is a better approach that what was in #150.

If this PR is indeed correct, I can work out some test cases and add it to the test suite.

cc @thehenrytsai @diehuxx


duncanmak:
Looks like we might have hit this issue https://github.com/ajv-validator/ajv-formats/pull/73

thehenrytsai (changes requested):
I am not a regex guru by any means and I understand if you don't want to write a full test suite before confirming approach is valid, but a couple of sanity test cases to begin would help.

Even the happiest path `2022-10-14T10:20:30.405060Z` doesn't appear to pass the regex.

I have no problem at all (and actually probably prefer it) for us to come up with our own tailor-made regex since we have specific requirements on the format.

dcrousso:
> Even the happiest path `2022-10-14T10:20:30.405060Z` doesn't appear to pass the regex.

where are you seeing this?  im pretty sure it works just fine
```js
/^\d\d\d\d-[0-1]\d-[0-3]\dt(?:[0-2]\d:[0-5]\d:[0-5]\d|23:59:60)(?:\.\d+)?(?:z|[+-]\d\d(?::?\d\d)?)$/i.test('2022-10-14T10:20:30.405060Z') // true
```

the only issue i potentially could see is that we have extra precision compared to browser environments, so if they changed the `\.\d+` to be `\.\d{,3}` then we may suddenly start having problems

thehenrytsai:
@dcrousso, you are right, I sanity tested a few tests and saw the wrong result, it's `2022-10-14T10:20:30.405Z` that is unexpectedly passing.

dcrousso:
> @dcrousso, you are right, I sanity tested a few tests and saw the wrong result, it's `2022-10-14T10:20:30.405Z` that is unexpectedly passing.

can you elaborate why that is unexpected?  is the issue that there's not enough precision?

duncanmak:
I'm happy that @thehenrytsai gave me permission to just add our own regex for the `date-time` format, so lets me sidestep the whole issue with using `ajv-formats` and dealing with ESM vs CJS. What a relief!

Once I made the change, I can see that some tests started to fail immediately, and they were easy to patch up. 

We started with 12 errors; I got it down to 2 errors with 59cc6ba9163eb617ccebf362898a22efe112865e.

The remaining errors I get from running `npm run test:node` are:

```
  2 failing

  1) RecordsQueryHandler.handle()
       functional tests
         should be able to range query by `dateCreated`:
     Error: /descriptor/dateCreated: must match format "date-time"
   
  2) RecordsQueryHandler.handle()
       functional tests
         should be able use range and exact match queries at the same time:
     Error: /descriptor/dateCreated: must match format "date-time"
  ```

Let me know if I'm going in the right direction, thanks!

**Comment by duncanmak** [comment]
In https://github.com/TBD54566975/dwn-sdk-js/pull/322/commits/edea94967ea29202f40382dd2e6641675259d183 I just decided to inline the date strings and that made all the tests pass. Let me know if we want to keep that.

**Comment by duncanmak** [comment]
I figured out how to construct `Temporal.Instant`s, so I took out the hardcoded strings.

Now the code looks more like how it was before, but we're not using `Temporal.PlainDateTime` anymore, and we now have the `Z` marker to designate UTC at the end of the string.

I hope this patch can now be called completed.

If we want to ensure that there are always 6 digits before the Z, the following regex can be used:

```
> const r = /^\d\d\d\d-[0-1]\d-[0-3]\dt(?:[0-2]\d:[0-5]\d:[0-5]\d|23:59:60)(?:\.\d{6})?(?:z|[+-]\d\d(?::?\d\d)?)$/i;
>
> r.test('2022-10-14T10:20:30.405060Z')
true
> r.test('2022-10-14T10:20:30.405Z')
false
```

**Review by dcrousso — APPROVED** [review]

File: build/compile-validators.js [hunk 1]
```diff
@@ -60,6 +60,7 @@ const schemas = {
 };
 
 const ajv = new Ajv({ code: { source: true, esm: true } });
+ajv.addFormat('date-time', /^\d\d\d\d-[0-1]\d-[0-3]\dt(?:[0-2]\d:[0-5]\d:[0-5]\d|23:59:60)(?:\.\d+)?(?:z|[+-]\d\d(?::?\d\d)?)$/i);
```
thehenrytsai, line 63: Can we define the regex in `definitions.json` so we only need to add these regex declaratively without needing awareness to js file.

tests/interfaces/records/handlers/records-query.spec.ts
```diff
@@ -20,14 +20,21 @@ import { Message } from '../../../../src/core/message.js';
...
 import { constructRecordsWriteIndexes } from '../../../../src/interfaces/records/handlers/records-write.js';
 import { DataStream, DidResolver, Dwn, HdKey, KeyDerivationScheme, Records } from '../../../../src/index.js';
 import { DateSort, RecordsQuery } from '../../../../src/interfaces/records/messages/records-query.js';
 
+declare global {
+  interface Date {
+    toTemporalInstant: typeof toTemporalInstant;
+  }
+}
+Date.prototype.toTemporalInstant = toTemporalInstant;
```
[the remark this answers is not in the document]
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/322/commits/9c1ef77f063e033217a5f8998bf1a2d226cf34fc

thehenrytsai:
> If we want to ensure that there are always 6 digits before the Z, the following regex can be used:

Yes, please enforce 6 digits, and please add tests to sanity verify the 3 cases outlined in #113 as you referenced.

build/compile-validators.js
duncanmak: I'm gonna try to find a way to do it, it's a good idea.

I had no idea that round-tripping a RegExp to string and back in JavaScript is this involved: https://stackoverflow.com/questions/39154255/converting-regexp-to-string-then-back-to-regexp

thehenrytsai, line 63: Took a look at the regex more closely and see that it has a few invalid/unnecessary requirements:

1. We don't want to allow time zone segment. Require UTC.
2. We don't want to allow case-insensitive check. Require upper case.
3. We want 6 digit decimal (as you are aware now)
4. Maybe this is needed by the standard, but I don't understand the `23:59:60` part.

dcrousso: > 2. We don't want to allow case-insensitive check. Require upper case.

is this because of the sorting/comparison logic we're using?

is uppercase `T` and `Z` required by ISO-8601?  if not i think we may be getting a bit too customized (and therefore might have interop problems with other libraries/platforms/etc. for generating date-time values) for this to still be called an "ISO date-time" :/

> 3. We want 6 digit decimal (as you are aware now)

this will cause issues in browser environments right now since `Date` will only give you millisecond precision

```
(new Date).toISOString() // "2023-04-26T21:01:02.638Z"
```

are we expecting that in the browser environment callers will manually parse and adjust the string to add an extra `000` before the `Z`?

> 4. Maybe this is needed by the standard, but I don't understand the `23:59:60` part.

im think this is for [leap seconds](https://en.wikipedia.org/wiki/Leap_second)

thehenrytsai: > is this because of the sorting/comparison logic we're using?

Yes, mainly.

> is uppercase T and Z required by ISO-8601?

I don't know, probably not. My view is that we already require a more restricted format e.g. UTC (while still being an IOS-8601 string), requiring upper case doesn't make it more evil than what it is. Also having a more restricted format means we can always relax it when we want to.

> are we expecting that in the browser environment callers will manually parse and adjust the string to add an extra 000 before the Z?

Depends on who you talk to. I am imagining them using our SDK to craft DWN messages, so that it's taken care for them regardless of environment.

> im think this is for [leap seconds](https://en.wikipedia.org/wiki/Leap_second)

Thanks for the info! Then I guess we should keep it as it's arguably better than specifying `[0-6]`, although I recognize that we still run through an actual parser to make sure the time is valid regardless (#324).

thehenrytsai: @duncanmak, I have gone ahead a update the regex. Let me know if you have any questions!

thehenrytsai: @duncanmak, I have gone ahead and moved it into `definitions.json`. Let me know if you have any questions.

json-schemas/definitions.json
```diff
@@ -14,6 +14,10 @@
     "did": {
       "type": "string",
       "pattern": "^did:([a-z0-9]+):((?:(?:[a-zA-Z0-9._-]|(?:%[0-9a-fA-F]{2}))*:)*((?:[a-zA-Z0-9._-]|(?:%[0-9a-fA-F]{2}))+))((;[a-zA-Z0-9_.:%-]+=[a-zA-Z0-9_.:%-]*)*)(\/[^#?]*)?([?][^#]*)?(#.*)?$"
+    },
+    "date-time": {
+      "type": "string",
+      "pattern": "^[0-9]{4}-[0-1]\\d-[0-3]\\dT([0-2]\\d:[0-5]\\d:[0-5]\\d|23:59:60)\\.[0-9]{6}Z$"
```
dcrousso, line 20: why not `\\d{4}` and `\\d{6}`?

also i we probably should put `?:` at the start of every capture group just so that we aren't actually capturing anything.  `(?:[0-2]\\d:[0-5]\\d:[0-5]\\d|23:59:60)`

thehenrytsai: Good suggestion on `?:`! On it!

build/compile-validators.js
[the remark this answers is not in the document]
thehenrytsai: Great suggestion, added #324 to track.

thehenrytsai (approved):
Thanks @duncanmak for the contribution! 🙏 🚀 🥇

Issue #319, opened 2023-04-25
Implement `protocols` key derivation scheme for encryption


thehenrytsai:
Implement `protocol` key derivation scheme with the following hierarchy order:

```JavaScript
const derivationPath = [
    'protocols',
    PROTOCOL_URI,
    CONTEXT_ID,
    ...[PROTOCOL_PATH_SEGMENTS],
    DATA_FORMAT,
    RECORD_ID
]
```

thehenrytsai:
Completed.

PR #323, opened 2023-04-25
#319 - implemented 'protocols' key derivation scheme for encryption

thehenrytsai:


src/core/dwn-error.ts
```diff
@@ -27,9 +27,10 @@ export enum DwnErrorCode {
   RecordsDeriveLeafPrivateKeyUnSupportedCurve = 'RecordsDeriveLeafPrivateKeyUnSupportedCurve',
   RecordsDeriveLeafPublicKeyUnSupportedCurve = 'RecordsDeriveLeafPublicKeyUnSupportedCurve',
   RecordsInvalidAncestorKeyDerivationSegment = 'RecordsInvalidAncestorKeyDerivationSegment',
+  RecordsProtocolsDerivationSchemeMissingProtocol = 'RecordsProtocolsDerivationSchemeMissingProtocol',
   RecordsWriteGetEntryIdUndefinedAuthor = 'RecordsWriteGetEntryIdUndefinedAuthor',
   RecordsWriteValidateIntegrityEncryptionCidMismatch = 'RecordsWriteValidateIntegrityEncryptionCidMismatch',
   Secp256k1KeyNotValid = 'Secp256k1KeyNotValid',
   UrlProtocolNotNormalized = 'UrlProtocolNotNormalized',
-  UrlPrococolNotNormalizable = 'UriPrococolNotNormalizable'
```
diehuxx, line 34: ty

src/utils/records.ts
```diff
@@ -58,6 +56,51 @@ export class Records {
...
+        DwnErrorCode.RecordsProtocolsDerivationSchemeMissingProtocol,
+        'Unable to construct key derivation path using `protocols` scheme because `protocol` is missing.'
+      );
+    }
+
+    const protocolPathSegments = descriptor.protocolPath!.split('/');
+    const fullDerivationPath = [
+      KeyDerivationScheme.Protocols,
+      descriptor.protocol,
+      contextId!,
```
diehuxx, line 95: Nit: why bother making allowing `contextId` to be undefined if we're just gonna `!`? I didn't see at a glance the spot between here and the top of `RecordsWrite#handler` that we ensure `contextId` is in fact defined.

diehuxx: Ah wait I see it. In `RecordsWrite#handler`, we do 
```typescript
if (descriptor.protocol !== undefined) {
  contextId = await RecordsWrite.getEntryId(author, descriptor);
}
```
Do you mind making `contextId: string` inside `constructKeyDerivationPathUsingProtocolsScheme` and doing the `context!` in `constructKeyDerivationPath` on line 70 instead?

thehenrytsai: @diehuxx, but `context` is only guaranteed defined if `protocol` is defined, and in `constructKeyDerivationPath` we don't know if `protocol` is defined or not, unless we move the `if (descriptor.protocol === undefined)` into `constructKeyDerivationPath` also, but then I intentionally wanted to hide scheme specific checks within scheme specific construction methods.

diehuxx: Ah that's fair. Agreed we def should keep the `contextId!` assertion after we check whether `protocol` is defined.

diehuxx (approved):

PR #327, opened 2023-04-26
Quality-of-life improvements for devs using the SDK for encryption

thehenrytsai:
Added a number of useful quality-of-life improvements for devs using the SDK for encryption:
1. When using root private key, derivation path of `[]` is no longer required to be specified
2. Encryption input used to create an encrypted message now only accepts the root public key


diehuxx (approved):
noice!

Issue #326, opened 2023-04-26
Revisit protocol definition schema style and consider adopting a style with predefined property names


thehenrytsai:
At minimum we should change `allow` into an array so that the same actor/role with different paths can be defined:

e.g. (just for consideration)
```json
"allow": [{
    "actor/role": "author",
    "of": "album",
    "to": ["create", "read"]
  },
  {
    "actor/role": "anyone",
    "to": ["read"]
  },
  {
    "actor/role": "author",
    "of": "photo",
    "to": ["update", "delete"]
  }
]
```

Also consider using this opportunity to reexamine choice of property names throughout the entire schema. Perhaps to `of` -> `protocolPath`, `to` -> `action/operation`, `labels` -> `constraints`,

diehuxx:
I figure we should rip this off like a bandaid, and I'm in favor of having all property names be predefined. I personally don't find the current declarative style more readable, and it leads to restrictions that I find hard to reason about.

Here's my proposal:
1. Rename certain fields
  a. `of` -> `protocolPath`
  b. `to` -> `action`
  c. `records` at the top level of the definition -> `recordsHierarchy` or just `hierarchy`
  d. `records` nested -> `descendants`
  e. `labels` -> `recordTypes`
2. Make all property names predefined
  a. Turn `labels` into an object array named `recordTypes`. Use `labels` as a property within each of the contained objects.
  b. Turn `recordsHierarchy` into an array of `ProtocolRuleSets`. Use `labels` as a property name within each `ProtocolRuleSet`


Here's Typescript types illustrating the above:

```typescript
type ProtocolDefinition = {
  recordTypes: ProtocolRecordType[],
  recordHierarchy: ProtocolRuleSet[],
};

type ProtocolRecordType = {
  label: string,
  schema?: string,
  dataFormats?: string[]
};

type ProtocolRuleSet = {
  label: string,
  allow?: {
    actor: ProtocolActor,
    protocolPath: string,
    actions: ProtocolAction[]
  },
  descendants?: ProtocolRuleSet[],
};

enum ProtocolActor {
  ANYONE = 'anyone',
  AUTHOR = 'author',
  RECIPIENT = 'recipient',
};

enum ProtocolAction {
  READ = 'read',
  WRITE = 'write',
}
```

And a short example protocol
```json
{
  "recordTypes": [
    {
      "label": "image",
      "schema": "http://example.com/image",
      "dataFormats": ["image/jpeg"]
    },
    {
[...]

PR #329, opened 2023-04-28
Refactor protocol allow rules

diehuxx:
Addresses one aspect of #326

A shortcoming of the existing protocol allow rule structure is that we can only specify one rule per actor. For example, this means cannot create both a rule allowing `author` of `image` to `write` and a rule `author` of `image/reply` to `read`. By refactoring `allow` to be an array, protocol allow rules become more flexible.

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -5,9 +5,9 @@ import type { ProtocolDefinition, ProtocolRuleSet, ProtocolsConfigureMessage } f
 import type { RecordsReadMessage, RecordsWriteMessage } from '../interfaces/records/types.js';
 
 import { RecordsWrite } from '../interfaces/records/messages/records-write.js';
-import { ProtocolAction, ProtocolActor } from '../interfaces/protocols/types.js';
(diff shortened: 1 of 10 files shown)


csuwildcat:
This is great, thank you for the quick turnaround, @diehuxx!

src/core/protocol-authorization.ts
```diff
@@ -1,16 +1,16 @@
...
 
 import { RecordsWrite } from '../interfaces/records/messages/records-write.js';
 import { DwnError, DwnErrorCode } from './dwn-error.js';
 import { DwnInterfaceName, DwnMethodName, Message } from './message.js';
 
 const methodToAllowedActionMap: Record<string, string> = {
-  [DwnMethodName.Write] : 'write',
-  [DwnMethodName.Read]  : 'read',
+  [DwnMethodName.Write] : ProtocolAction.WRITE,
+  [DwnMethodName.Read]  : ProtocolAction.READ,
```
thehenrytsai, line 13: We should have consistent naming style for enums.

thehenrytsai (approved):
Thanks for the fast turnaround!

PR #334, opened 2023-04-29
move published === true check from handlers/records-read.ts into RecordsRead.authorize(), resolve #321

### LINKED ISSUE #321: move `published === true` check from `handlers/records-read.ts` into `RecordsRead.authorize()`

vlad-timofeev:


diehuxx (approved):
Thanks

PR #337, opened 2023-05-02
#333 #326 Refactor protocol recordTypes

diehuxx:
Another step in the refactor outlined in #326. Also, fully addresses feature #333.

## Changes
#326
* Change property `label` to `recordDefinitions`. Make `recordDefinitions` an array, where each element has `id`, `schema`, and `dataFormats`

#333
* Makes `schema` optional in protocol auth'd `RecordsWrites`. (`schema` was previously required for protocol `RecordsWrites`.)

Other
* Remove duplicate tests and flesh out tests for `ProtocolAuthrorization.getRuleSet`, particularly testing for new error code `DwnErrorCode.ProtocolAuthorizationMissingRuleSet`.

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -4,11 +4,11 @@ import type { Filter, TimestampedMessage } from './types.js';
 import type { ProtocolDefinition, ProtocolRecordDefinition, ProtocolRuleSet, ProtocolsConfigureMessage } from '../interfaces/protocols/types.js';
 import type { RecordsReadMessage, RecordsWriteMessage } from '../interfaces/records/types.js';
 
+import { Protocols } from '../utils/protocols.js';
 import { RecordsWrite } from '../interfaces/records/messages/records-write.js';
 import { DwnError, DwnErrorCode } from './dwn-error.js';
 import { DwnInterfaceName, DwnMethodName, Message } from './message.js';
 import { ProtocolAction, ProtocolActor } from '../interfaces/protocols/types.js';
-import { Protocols } from '../utils/protocols.js';
 
 const methodToAllowedActionMap: Record<string, string> = {
   [DwnMethodName.Write] : ProtocolAction.Write,
@@ -215,7 +215,7 @@ export class ProtocolAuthorization {
 
     const recordDefinitionIds = recordDefinitions.map((recordDefinition) => recordDefinition.id);
     const declaredProtocolPath = (inboundMessage as RecordsWrite).message.descriptor.protocolPath!;
-    const declaredRecordDefinitionId = ProtocolAuthorization.getRecordDefinitionFromPath(declaredProtocolPath);
+    const declaredRecordDefinitionId = ProtocolAuthorization.getRecordDefinitionId(declaredProtocolPath);
     if (!recordDefinitionIds.includes(declaredRecordDefinitionId)) {
       throw new DwnError(DwnErrorCode.ProtocolAuthorizationInvalidRecordDefinition,
         `record with recordDefinition ${declaredRecordDefinitionId} not allowed in protocol`);
@@ -224,7 +224,7 @@ export class ProtocolAuthorization {
     let ancestorProtocolPath: string = '';
     for (const ancestor of ancestorMessageChain) {
       const protocolPath = ancestor.descriptor.protocolPath!;
-      const ancestorRecordDefinitionId = ProtocolAuthorization.getRecordDefinitionFromPath(protocolPath);
+      const ancestorRecordDefinitionId = ProtocolAuthorization.getRecordDefinitionId(protocolPath);
       ancestorProtocolPath += `${ancestorRecordDefinitionId}/`; // e.g. `foo/bar/`, notice the trailing slash
     }
 
@@ -254,7 +254,7 @@ export class ProtocolAuthorization {
     const recordsWriteMessage = inboundMessage as RecordsWriteMessage;
 
     const protocolPath = recordsWriteMessage.descriptor.protocolPath!;
-    const recordDefinitionId = ProtocolAuthorization.getRecordDefinitionFromPath(protocolPath);
+    const recordDefinitionId = ProtocolAuthorization.getRecordDefinitionId(protocolPath);
     // existence of recordDefinition has already been verified
     const recordDefinition: ProtocolRecordDefinition = Protocols.getRecordDefinition(protocolDefinition, recordDefinitionId)!;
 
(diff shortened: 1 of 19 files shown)


src/core/protocol-authorization.ts
```diff
@@ -174,36 +169,40 @@ export class ProtocolAuthorization {
...
-
-      // if we are looking at the inbound message itself (the last message in the chain),
-      // then we have found the access control object we need to evaluate against
-      if (currentMessageIndex === messageChain.length - 1) {
-        return allowedRecordsAtCurrentLevel[currentRecordType];
+    // get top level rule set
+    const topLevelRuleSet: ProtocolRuleSet | undefined = protocolDefinition.records[protocolPathArray[0]];
+    if (topLevelRuleSet === undefined) {
+      throw new DwnError(DwnErrorCode.ProtocolAuthorizationMissingRuleSet,
+        `No rule set defined for protocolPath ${protocolPathArray[0]}`);
```
diehuxx, line 185: Question for reviewers:
Previously this error message was:
```
throw new Error(`record with schema: '${currentRecordSchema}' not allowed in structure level ${currentMessageIndex}\`);
```

Do we prefer language similar to that or "No rule set defined" as I have it here?

csuwildcat (approved):

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -1480,7 +1520,7 @@ describe('RecordsWriteHandler.handle()', () => {
           data         : new TextEncoder().encode('data1'),
           protocol     : 'example.com/',
           protocolPath : 'email', // from email protocol
-          schema       : emailProtocolDefinition.labels.email.schema
+          schema       : emailProtocolDefinition.recordTypes.find(({ id }) => id === 'email')!.schema
```
thehenrytsai, line 1523: Consider creating a util method perhaps `getSchemaFromRecordTypeId(definition, recordTypeId)`

src/interfaces/protocols/types.ts
```diff
@@ -10,17 +10,18 @@ export type ProtocolsConfigureDescriptor = {
...
-      dataFormats?: string[],
-    };
-  };
+  recordTypes: ProtocolRecordType[];
   records: {
     [key: string]: ProtocolRuleSet;
   };
 };
 
+export type ProtocolRecordType = {
```
thehenrytsai, line 19: Discussion: frankly I am still not 100% with "Record Type", because "schema" and "type" feels interchangeable in common programming lingo, it is not the case here, hence the hesitation.

I feel "constraint" is the best word I can find that says what this data is for, but also not 100% happy if that means we end up calling the referencing property name under `records` hierarchy `recordConstraint` instead of `recordType`.

What if we call `recordTypes` -> `recordDefinitions`? And in `records` hierarchy, we just call the referencing property `record`?

tests/vectors/protocol-definitions/dex.json
```diff
@@ -1,15 +1,18 @@
...
     },
-    "fulfillment": {
+    {
+      "id": "fulfillment",
       "schema": "https://tbd/website/tbdex/fulfillment"
     }
-  },
+  ],
   "records": {
     "ask": {
```
thehenrytsai, line 17: Sanity: will there be a follow up PR that removes the dynamic properties under `records` (e.g. `ask` in this case)?

diehuxx: Yes, that's the plan. I didn't want to put it all in one PR

src/core/protocol-authorization.ts
thehenrytsai: We definitely should avoid reusing error code in principle. I am more interested in knowing why we can't include this check in the loop. (I have not studied the code yet)

src/interfaces/protocols/types.ts
diehuxx: This works for me. I like `recordDefinitions` better than `recordTypes` now that you mention it. `record` property in the hierarchy also works for me.

src/core/protocol-authorization.ts
```diff
@@ -244,35 +245,42 @@ export class ProtocolAuthorization {
...
     }
     const recordsWriteMessage = inboundMessage as RecordsWriteMessage;
 
-    const currentRecordSchema = recordsWriteMessage.descriptor.schema!;
-    const currentRecordSchemaLabel = recordSchemaToLabelMap.get(currentRecordSchema)!;
-    const expectedDataFormats = protocolDefinition.labels[currentRecordSchemaLabel].dataFormats;
+    const protocolPath = recordsWriteMessage.descriptor.protocolPath!;
+    const recordDefinitionId = ProtocolAuthorization.getRecordDefinitionFromPath(protocolPath);
+    // existence of recordDefinition has already been verified
+    const recordDefinition: ProtocolRecordDefinition = protocolDefinition.recordDefinitions.find((recordDefinition) => recordDefinition.id === recordDefinitionId)!;
```
thehenrytsai, line 265: Maybe use `Protocols.getRecordDefinition()`

src/core/protocol-authorization.ts
```diff
@@ -413,4 +417,7 @@ export class ProtocolAuthorization {
     }
   }
 
+  private static getRecordDefinitionFromPath(protocolPath: string): string {
```
thehenrytsai, line 420: I kept thinking it's the definition that's returned.

```suggestion
  private static getRecordDefinitionIdFromPath(protocolPath: string): string {
```
Could also be more concise if you prefer:
```suggestion
  private static getRecordDefinitionId(protocolPath: string): string {
```

src/core/protocol-authorization.ts
```diff
@@ -399,7 +403,7 @@ export class ProtocolAuthorization {
       const expectedAncestorType = expectedAncestors[i];
       const ancestorMessage = ancestorMessageChain[i];
 
-      const actualAncestorType = recordSchemaToLabelMap.get(ancestorMessage.descriptor.schema!);
+      const actualAncestorType = ProtocolAuthorization.getRecordDefinitionFromPath(ancestorMessage.descriptor.protocolPath!);
```
thehenrytsai, line 406: Nit:
```suggestion
      const actualAncestorDefinitionId = ProtocolAuthorization.getRecordDefinitionFromPath(ancestorMessage.descriptor.protocolPath!);
```

src/core/protocol-authorization.ts
diehuxx: That seems fine for now, but only because I think I can make it pretty again once I refactor `records` into an array of `ProtocolRuleSet`s.

My gripe is that I don't like having a variable called `currentRuleSet` where the type is not just `ProtocolRuleSet`. And having two different error codes for what is IMO the same error also feels wrong.

src/core/protocol-authorization.ts
diehuxx: fixed

src/core/protocol-authorization.ts
diehuxx: fixed

src/core/protocol-authorization.ts
diehuxx: fixed

thehenrytsai (approved):

thehenrytsai (approved):

src/core/protocol-authorization.ts
diehuxx: > the desire to figure out the exact line that causes an issue when debugging

great point, I didn't think of that. You've got me convinced :)

Issue #288, opened 2023-03-31
consider changing `StorageController` to a class that is instantiated with all of the stores


mistermoe:
I see that all interface methods, in their constructors, take `messageStore`, `dataStore`, and now `eventLog`. e.g.
* [records-delete](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-delete.ts#L15)
* [records-write](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-write.ts#L16)
* [records-query](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-query.ts#L16)
* [records-read](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-read.ts#L15)

but it seems like they're almost always passed to [StorageController](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/storage-controller.ts)  to do stuff like queries, puts etc. i do see some areas that use _just_ `messageStore.query` and not `StorageController.query`, presumably because `StorageController` also returns the respective data if its under a certain size?

it seems like, we could instantiate a `StorageController` instance in `dwn.create` and pass that to each individual handler. `StorageController` could simply have getters for `messageStore` when its needed individually. 

curious to hear ppls thoughts here. basically, since we have `StorageController` now, we might as well leverage it to reduce the # of arguments that are passed to each interface method handler

duncanmak:
I started working on this here: https://github.com/TBD54566975/dwn-sdk-js/pull/338

I just re-read @mistermoe's message above, I think the tests are failing in my PR because I've confused 'messageStore.query` with `storageController.query`, even though their argument lists are compatible.

thehenrytsai:
`StorageController` is refactored and is on "less" critical path than before, for instance, `put()` is no longer going through `StorageController`. The debate in PR #338 also seems to suggest that it is not all bad that dependencies are injected explicitly, so closing this and we can reopen again if the code is become too unwieldy that requires a second look.

PR #338, opened 2023-05-03
Turn StorageController into a class

duncanmak:
Refactor the `(messageStore, dataStore, eventlog)` tuple into a single instance of `StorageController`, which addresses #288 

### Pending questions:

~~* Should StorageController use `BaseMessage` instead of the more specific `RecordsWriteMessageWithOptionalEncodedData`?~~
~~* Settle on method names for new methods in `StorageController`, i.e. how best to differentiate between operations (`get`, `put`) and the different stores (`messageStore` vs `dataStore`)~~

cc. @thehenrytsai @mistermoe

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -352,7 +352,10 @@ export class ProtocolAuthorization {
    * @throws {Error} if fails verification
    */
   private static async verifyActionCondition(
-    tenant: string, incomingMessage: RecordsRead | RecordsWrite, storageController: StorageController): Promise<void> {
+    tenant: string,
+    incomingMessage: RecordsRead | RecordsWrite,
+    storageController: StorageController
+  ): Promise<void> {
     if (incomingMessage.message.descriptor.method === DwnMethodName.Read) {
       // Currently no conditions for reads
     } else if (incomingMessage.message.descriptor.method === DwnMethodName.Write) {
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -1,5 +1,5 @@
-import type { MessageStore } from '../store/message-store.js';
 import type { RecordsRead } from '../interfaces/records/messages/records-read.js';
+import type { StorageController } from '../store/storage-controller.js';
 import type { Filter, TimestampedMessage } from './types.js';
 import type { ProtocolDefinition, ProtocolRuleSet, ProtocolsConfigureMessage } from '../interfaces/protocols/types.js';
 import type { RecordsReadMessage, RecordsWriteMessage } from '../interfaces/records/types.js';
@@ -24,18 +24,18 @@ export class ProtocolAuthorization {
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
     requesterDid: string | undefined,
-    messageStore: MessageStore
+    storageController: StorageController
   ): Promise<void> {
     // fetch ancestor message chain
     const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, messageStore);
+      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, storageController);
 
     // fetch the protocol definition
     const protocolDefinition = await ProtocolAuthorization.fetchProtocolDefinition(
       tenant,
       incomingMessage,
       ancestorMessageChain,
-      messageStore
+      storageController
     );
 
     // record schema -> schema label map
@@ -73,7 +73,7 @@ export class ProtocolAuthorization {
     );
 
     // verify allowed condition of incoming message
-    await ProtocolAuthorization.verifyActionCondition(tenant, incomingMessage, messageStore);
+    await ProtocolAuthorization.verifyActionCondition(tenant, incomingMessage, storageController);
   }
 
   /**
@@ -83,7 +83,7 @@ export class ProtocolAuthorization {
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
     ancestorMessageChain: RecordsWriteMessage[],
-    messageStore: MessageStore
+    storageController: StorageController
   ): Promise<ProtocolDefinition> {
     // get the protocol URI
     let protocolUri: string;
@@ -99,7 +99,7 @@ export class ProtocolAuthorization {
       method    : DwnMethodName.Configure,
       protocol  : protocolUri
     };
-    const protocols = await messageStore.query(tenant, query) as ProtocolsConfigureMessage[];
+    const protocols = await storageController.queryMessages(tenant, query) as ProtocolsConfigureMessage[];
 
     if (protocols.length === 0) {
       throw new Error(`unable to find protocol definition for ${protocolUri}`);
@@ -116,7 +116,7 @@ export class ProtocolAuthorization {
   private static async constructAncestorMessageChain(
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
-    messageStore: MessageStore
+    storageController: StorageController
   )
     : Promise<RecordsWriteMessage[]> {
     const ancestorMessageChain: RecordsWriteMessage[] = [];
@@ -132,7 +132,7 @@ export class ProtocolAuthorization {
         method    : DwnMethodName.Write,
         recordId  : recordsRead.message.descriptor.recordId,
       };
-      const existingMessages = await messageStore.query(tenant, query) as TimestampedMessage[];
+      const existingMessages = await storageController.queryMessages(tenant, query) as TimestampedMessage[];
       const recordsWriteMessage = await RecordsWrite.getNewestMessage(existingMessages) as RecordsWriteMessage;
       recordsWrite = await RecordsWrite.parse(recordsWriteMessage);
       ancestorMessageChain.push(recordsWrite.message);
@@ -152,7 +152,7 @@ export class ProtocolAuthorization {
         contextId,
         recordId  : currentParentId
       };
-      const parentMessages = await messageStore.query(tenant, query) as RecordsWriteMessage[];
+      const parentMessages = await storageController.queryMessages(tenant, query) as RecordsWriteMessage[];
 
       if (parentMessages.length === 0) {
         throw new Error(`no parent found with ID ${currentParentId}`);
@@ -351,7 +351,8 @@ export class ProtocolAuthorization {
    * Currently the only check is: if the write is not the initial write, the author must be the same as the initial write
    * @throws {Error} if fails verification
    */
-  private static async verifyActionCondition(tenant: string, incomingMessage: RecordsRead | RecordsWrite, messageStore: MessageStore): Promise<void> {
+  private static async verifyActionCondition(
+    tenant: string, incomingMessage: RecordsRead | RecordsWrite, storageController: StorageController): Promise<void> {
     if (incomingMessage.message.descriptor.method === DwnMethodName.Read) {
       // Currently no conditions for reads
     } else if (incomingMessage.message.descriptor.method === DwnMethodName.Write) {
@@ -362,7 +363,7 @@ export class ProtocolAuthorization {
         const query = {
           entryId: recordsWrite.message.recordId
         };
-        const result = await messageStore.query(tenant, query) as RecordsWriteMessage[];
+        const result = await storageController.queryMessages(tenant, query) as RecordsWriteMessage[];
 
         // check the author of the initial write matches the author of the incoming message
         const initialWrite = result[0];
--- a/src/interfaces/messages/handlers/messages-get.ts
+++ b/src/interfaces/messages/handlers/messages-get.ts
@@ -36,7 +36,7 @@ export class MessagesGetHandler implements MethodHandler {
     const messageCids = new Set(message.descriptor.messageCids);
 
     for (const messageCid of messageCids) {
-      const promise = this.storageController.MessageStore.get(tenant, messageCid)
+      const promise = this.storageController.getMessage(tenant, messageCid)
         .then(message => {
           return { messageCid, message };
         })
@@ -70,7 +70,7 @@ export class MessagesGetHandler implements MethodHandler {
 
       if (dataCid !== undefined && dataSize! <= DwnConstant.maxDataSizeAllowedToBeEncoded) {
         const messageCid = await Message.getCid(message);
-        const result = await this.storageController.get(tenant, messageCid, dataCid);
+        const result = await this.storageController.getData(tenant, messageCid, dataCid);
 
         if (result) {
           const dataBytes = await DataStream.toBytes(result.dataStream);
--- a/src/interfaces/messages/handlers/messages-get.ts
+++ b/src/interfaces/messages/handlers/messages-get.ts
@@ -1,7 +1,6 @@
-import type { DataStore } from '../../../store/data-store.js';
 import type { DidResolver } from '../../../did/did-resolver.js';
-import type { MessageStore } from '../../../store/message-store.js';
 import type { MethodHandler } from '../../types.js';
+import type { StorageController } from '../../../store/storage-controller.js';
 import type { MessagesGetMessage, MessagesGetReply, MessagesGetReplyEntry } from '../types.js';
 
 import { DataStream } from '../../../utils/data-stream.js';
@@ -15,7 +14,7 @@ import { DwnInterfaceName, DwnMethodName, Message } from '../../../core/message.
 type HandleArgs = { tenant: string, message: MessagesGetMessage };
 
 export class MessagesGetHandler implements MethodHandler {
-  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private dataStore: DataStore) {}
+  constructor(private didResolver: DidResolver, private storageController: StorageController) {}
 
   public async handle({ tenant, message }: HandleArgs): Promise<MessagesGetReply> {
     let messagesGet: MessagesGet;
@@ -37,7 +36,7 @@ export class MessagesGetHandler implements MethodHandler {
     const messageCids = new Set(message.descriptor.messageCids);
 
     for (const messageCid of messageCids) {
-      const promise = this.messageStore.get(tenant, messageCid)
+      const promise = this.storageController.MessageStore.get(tenant, messageCid)
         .then(message => {
           return { messageCid, message };
         })
@@ -71,7 +70,7 @@ export class MessagesGetHandler implements MethodHandler {
 
       if (dataCid !== undefined && dataSize! <= DwnConstant.maxDataSizeAllowedToBeEncoded) {
         const messageCid = await Message.getCid(message);
-        const result = await this.dataStore.get(tenant, messageCid, dataCid);
+        const result = await this.storageController.get(tenant, messageCid, dataCid);
 
         if (result) {
           const dataBytes = await DataStream.toBytes(result.dataStream);
--- a/src/interfaces/permissions/handlers/permissions-request.ts
+++ b/src/interfaces/permissions/handlers/permissions-request.ts
@@ -35,7 +35,7 @@ export class PermissionsRequestHandler implements MethodHandler {
       author,
       ... message.descriptor
     };
-    await this.storageController.put(tenant, message, index as any); // FIXME
+    await this.storageController.putMessageStore(tenant, message, index as any); // FIXME
 
     return new MessageReply({
       status: { code: 202, detail: 'Accepted' }
--- a/src/interfaces/permissions/handlers/permissions-request.ts
+++ b/src/interfaces/permissions/handlers/permissions-request.ts
@@ -35,7 +35,7 @@ export class PermissionsRequestHandler implements MethodHandler {
       author,
       ... message.descriptor
     };
-    await this.storageController.putMessage(tenant, message, index as any); // FIXME
+    await this.storageController.putMessageWithoutData(tenant, message, index as any); // FIXME
 
     return new MessageReply({
       status: { code: 202, detail: 'Accepted' }
--- a/src/interfaces/permissions/handlers/permissions-request.ts
+++ b/src/interfaces/permissions/handlers/permissions-request.ts
@@ -35,7 +35,7 @@ export class PermissionsRequestHandler implements MethodHandler {
       author,
       ... message.descriptor
     };
-    await this.storageController.putMessageStore(tenant, message, index as any); // FIXME
+    await this.storageController.putMessage(tenant, message, index as any); // FIXME
 
     return new MessageReply({
       status: { code: 202, detail: 'Accepted' }
--- a/src/interfaces/permissions/handlers/permissions-request.ts
+++ b/src/interfaces/permissions/handlers/permissions-request.ts
@@ -1,6 +1,8 @@
+import type { DidResolver } from '../../../index.js';
 import type { MethodHandler } from '../../types.js';
 import type { PermissionsRequestMessage } from '../types.js';
-import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
+import type { StorageController } from '../../../store/storage-controller.js';
+
 
 import { canonicalAuth } from '../../../core/auth.js';
 import { MessageReply } from '../../../core/message-reply.js';
@@ -8,7 +10,7 @@ import { PermissionsRequest } from '../messages/permissions-request.js';
 
 export class PermissionsRequestHandler implements MethodHandler {
 
-  constructor(private didResolver: DidResolver, private messageStore: MessageStore,private dataStore: DataStore) { }
+  constructor(private didResolver: DidResolver, private storageController: StorageController) { }
 
   public async handle({
     tenant,
@@ -33,7 +35,7 @@ export class PermissionsRequestHandler implements MethodHandler {
       author,
       ... message.descriptor
     };
-    await this.messageStore.put(tenant, message, index as any); // FIXME
+    await this.storageController.put(tenant, message, index as any); // FIXME
 
     return new MessageReply({
       status: { code: 202, detail: 'Accepted' }
--- a/src/interfaces/protocols/handlers/protocols-configure.ts
+++ b/src/interfaces/protocols/handlers/protocols-configure.ts
@@ -60,7 +60,7 @@ export class ProtocolsConfigureHandler implements MethodHandler {
 
       // FIXME: indexes, Property 'dataSize' is incompatible with index signature.
       // Type 'number' is not assignable to type 'string'.
-      await this.storageController.putWithData(tenant, message, indexes as any, dataStream);
+      await this.storageController.putMessageWithData(tenant, message, indexes as any, dataStream);
 
       messageReply = new MessageReply({
         status: { code: 202, detail: 'Accepted' }
--- a/src/interfaces/protocols/handlers/protocols-configure.ts
+++ b/src/interfaces/protocols/handlers/protocols-configure.ts
@@ -1,18 +1,17 @@
-import type { EventLog } from '../../../event-log/event-log.js';
+import type { DidResolver } from '../../../index.js';
 import type { MethodHandler } from '../../types.js';
 import type { ProtocolsConfigureMessage } from '../types.js';
-import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
+import type { StorageController } from '../../../store/storage-controller.js';
 
 import { canonicalAuth } from '../../../core/auth.js';
 import { MessageReply } from '../../../core/message-reply.js';
 import { ProtocolsConfigure } from '../messages/protocols-configure.js';
-import { StorageController } from '../../../store/storage-controller.js';
 
 import { DwnInterfaceName, DwnMethodName, Message } from '../../../core/message.js';
 
 export class ProtocolsConfigureHandler implements MethodHandler {
 
-  constructor(private didResolver: DidResolver, private messageStore: MessageStore, private dataStore: DataStore, private eventLog: EventLog) { }
+  constructor(private didResolver: DidResolver, private storageController: StorageController) { }
 
   public async handle({
     tenant,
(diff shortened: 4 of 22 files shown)


dcrousso:
we probably should just make `StorageController` use `BaseMessage` instead of `RecordsWriteMessageWithOptionalEncodedData` since the base `Descriptor` also has `dataCid` and `dataSize` (though this may ripple out into a bunch of the callers having to `as RecordsWriteMessageWithOptionalEncodedData[]` instead, but IMO that's maybe better?)

alternatively we could rename `query` to `queryRecordsWrites` and add a new `query` that just calls into `MessageStore`?

duncanmak:
@dcrousso Thanks for your comments. I had a similar idea and introduced `queryMessageStore` in a new commit.

Personally, I'm not a fan of types like `RecordsWriteMessageWithOptionalEncodedData`, they seem overly specific - so the suggestion to use `BaseMessage` resonates with me. However, I'll defer to those who actually understand the protocol and the actual details to make the right design decision.

Some tests are still failing, so I'll take another look and see if I can spot anything out of the ordinary. I welcome more reviews!

src/interfaces/messages/handlers/messages-get.ts
```diff
@@ -37,7 +36,7 @@ export class MessagesGetHandler implements MethodHandler {
     const messageCids = new Set(message.descriptor.messageCids);
 
     for (const messageCid of messageCids) {
-      const promise = this.messageStore.get(tenant, messageCid)
+      const promise = this.storageController.MessageStore.get(tenant, messageCid)
```
dcrousso, line 39: instead of exposing the `MessageStore` should we have a `getMessage`?
```
getMessage(tenant: string, messageCid: string) {
  return this.messageStore.get(tenant, messageCid);
}
```

src/interfaces/protocols/handlers/protocols-configure.ts
```diff
@@ -40,7 +39,7 @@ export class ProtocolsConfigureHandler implements MethodHandler {
       method    : DwnMethodName.Configure,
       protocol  : message.descriptor.protocol
     };
-    const existingMessages = await this.messageStore.query(tenant, query) as ProtocolsConfigureMessage[];
+    const existingMessages = await this.storageController.queryMessageStore(tenant, query) as ProtocolsConfigureMessage[];
```
dcrousso, line 42: similar to [src/interfaces/messages/handlers/messages-get.ts:39](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-5d5c842d6507a8df7c2f87057043b86e17d4fc15a5735e010e2c3c3d79c765a0R39), perhaps we should name this `queryMessages`?

src/interfaces/records/handlers/records-read.ts
```diff
@@ -53,13 +53,13 @@ export class RecordsReadHandler implements MethodHandler {
 
     const newestRecordsWrite = newestExistingMessage as RecordsWriteMessage;
     try {
-      await recordsRead.authorize(tenant, await RecordsWrite.parse(newestRecordsWrite), this.messageStore);
+      await recordsRead.authorize(tenant, await RecordsWrite.parse(newestRecordsWrite), this.storageController.MessageStore);
```
dcrousso, line 56: hmmm moving everything into `StorageController` to then still have to expose `MessageStore` is unfortunate :(

perhaps we should do something like this?
```
authorizeRecordsRead(tenant: string, recordsRead: RecordsRead, newsetRecordsWrite: RecordsWrite) {
  return recordsRead.authorize(tenant, newestRecordsWrite, this.messageStore);
}
```
this way we don't expose `this.messageStore` and keep it kinda internal :)

src/interfaces/records/handlers/records-write.ts
```diff
@@ -32,7 +31,7 @@ export class RecordsWriteHandler implements MethodHandler {
     // authentication & authorization
     try {
       await authenticate(message.authorization, this.didResolver);
-      await recordsWrite.authorize(tenant, this.messageStore);
+      await recordsWrite.authorize(tenant, this.storageController.MessageStore);
```
dcrousso, line 34: ditto [src/interfaces/records/handlers/records-read.ts:56](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-72cc0e5d5cf69521228f5e6253cb4d3be3b0888330231880addb51e836c24f6fR56) but as `authorizeRecordsWrite`

src/store/storage-controller.ts
```diff
@@ -111,19 +109,35 @@ export class StorageController {
...
+    await this.messageStore.delete(tenant, messageCid);
+  }
+
+  public get MessageStore(): MessageStore { return this.messageStore; }
+
+  public queryMessageStore(tenant: string, filter: Filter, options?: MessageStoreOptions): Promise<BaseMessage[]> {
+    return this.messageStore.query(tenant, filter, options);
+  }
+
+  public get(tenant: string, messageCid: string, dataCid: string): Promise<GetResult|undefined> {
```
dcrousso, line 131: i think `getData` might be more clear since this is just returning the data, not the message (also matches nicely with [src/interfaces/messages/handlers/messages-get.ts:39](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-5d5c842d6507a8df7c2f87057043b86e17d4fc15a5735e010e2c3c3d79c765a0R39))

src/store/storage-controller.ts
```diff
@@ -111,19 +109,35 @@ export class StorageController {
...
+    return this.dataStore.get(tenant, messageCid, dataCid);
+  }
+
+  public deleteEventsByCid(tenant: string, deletedMessageCids: any): void {
```
dcrousso, line 135: NIT: I think we can just call this `deleteEvents` since the `ByCid` part is kinda obvious based on the `deletedMessageCids` parameter

src/store/storage-controller.ts
```diff
@@ -111,19 +109,35 @@ export class StorageController {
...
+    if (this.eventLog) { this.eventLog.deleteEventsByCid(tenant, deletedMessageCids); }
```
dcrousso, line 136: when does `this.eventLog` not exist?

also, should we `return` the result of this?

src/store/storage-controller.ts
```diff
@@ -111,19 +109,35 @@ export class StorageController {
...
+  }
+
+  public appendEventLog(tenant: string, messageCid: string): Promise<string> | undefined {
```
dcrousso, line 139: NIT: how about just `appendEvent`?

src/store/storage-controller.ts
```diff
@@ -111,19 +109,35 @@ export class StorageController {
...
+    if (this.eventLog) { return this.eventLog.append(tenant, messageCid); }
```
dcrousso, line 140: ditto ([:139](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-cbe02e4d9564ce57729575f2fce1680efec7c8f72fe777276201b785fc0b3e8dR136))

src/store/storage-controller.ts
duncanmak: I made the event log an optional argument of the StorageController constructor.

dcrousso: is there a real usecase where it's not expected/required?

if the answer is "tests" IMO that's not a very good reason and we should just create it where needed (or not even use `StorageController` in the first place)

src/interfaces/messages/handlers/messages-get.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/interfaces/protocols/handlers/protocols-configure.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/8a39de45aaa23185e406373a304c4ee9e4d68730

src/store/storage-controller.ts
```diff
@@ -23,10 +26,7 @@ export class StorageController {
    * @throws {DwnError} with `DwnErrorCode.MessageStoreDataSizeMismatch`
    *                    if `dataSize` in `descriptor` given mismatches the actual data size
    */
-  public static async put(
-    messageStore: MessageStore,
-    dataStore: DataStore,
-    eventLog: EventLog,
+  public async put(
```
dcrousso, line 29: should we rename this `putWithData` to better differentiate it from `putMessage`(/`putMessageWithoutData`)?

src/store/storage-controller.ts
```diff
@@ -80,26 +80,24 @@ export class StorageController {
...
-    await messageStore.put(tenant, message, indexes);
-    await eventLog.append(tenant, messageCid);
+    await this.messageStore.put(tenant, message, indexes);
+    if (this.eventLog) { await this.eventLog.append(tenant, messageCid); }
   }
 
-  public static async query(
-    messageStore: MessageStore,
-    dataStore: DataStore,
+  public async query(
```
dcrousso, line 87: should we rename this `queryRecordsWrites` to better differentiate it from `queryMessages`?

src/store/storage-controller.ts
```diff
@@ -111,19 +109,43 @@ export class StorageController {
...
+    await this.messageStore.delete(tenant, messageCid);
+  }
+
+  public get MessageStore(): MessageStore { return this.messageStore; }
+
+  public queryMessages(tenant: string, filter: Filter, options?: MessageStoreOptions): Promise<BaseMessage[]> {
+    return this.messageStore.query(tenant, filter, options);
+  }
+
+  public putMessage(tenant: string, messageJson: BaseMessage, indexes, options?): Promise<void> {
```
dcrousso, line 131: maybe `putMessageWithoutData` so that it's super clear "we are adding to the `MessageStore` without adding to the `DataStore`, which probably should only be done in super specific circumstances"?

also can we move these to be more nearby the similarly named functions (e.g. `putMessageWithoutData` would go right above/below `putMessageWithData`, `queryMessages` would go right above/below `queryRecordsWrites`, etc.)?

src/store/storage-controller.ts
```diff
@@ -111,19 +109,43 @@ export class StorageController {
     return messages;
   }
 
-  public static async delete(
-    messageStore: MessageStore,
-    dataStore: DataStore,
+  public async delete(
```
dcrousso, line 112: maybe `deleteMessage` to differentiate from `deleteEvents`?

src/store/storage-controller.ts
```diff
@@ -111,19 +109,43 @@ export class StorageController {
...
+
+  public getMessage(tenant: string, messageCid: string): Promise<BaseMessage|undefined> {
+    return this.messageStore.get(tenant, messageCid);
+  }
+
+  public getData(tenant: string, messageCid: string, dataCid: string): Promise<GetResult|undefined> {
+    return this.dataStore.get(tenant, messageCid, dataCid);
+  }
+
+  public appendEvents(tenant: string, messageCid: string): Promise<string>|undefined {
```
dcrousso, line 143: i think this should just be `appendEvent` since it's only dealing with a single event at a time

src/store/storage-controller.ts
```diff
@@ -80,26 +80,24 @@ export class StorageController {
       }
     }
 
-    await messageStore.put(tenant, message, indexes);
-    await eventLog.append(tenant, messageCid);
+    await this.messageStore.put(tenant, message, indexes);
+    if (this.eventLog) { await this.eventLog.append(tenant, messageCid); }
```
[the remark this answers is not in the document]
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/b06fca146c088cc95556799286f149ea8394406a

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/ecb7e0da3886d0d2023341165c753d09504f9ea2

src/interfaces/records/handlers/records-write.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/10146b17e43411da36e2e7a615a26c726441d9b7

src/interfaces/records/handlers/records-read.ts
```diff
@@ -53,13 +53,13 @@ export class RecordsReadHandler implements MethodHandler {
 
     const newestRecordsWrite = newestExistingMessage as RecordsWriteMessage;
     try {
-      await recordsRead.authorize(tenant, await RecordsWrite.parse(newestRecordsWrite), this.messageStore);
+      await this.storageController.authorizeRecordsRead(tenant, recordsRead, await RecordsWrite.parse(newestRecordsWrite));
```
dcrousso, line 56: so i thought about this a bit more, and i think it might actually be nicer to pass the `StorageController` through `RecordsRead.prototype.authorize` instead of `MessageStore`, as that way we can basically isolate `MessageStore` vs` DataStore` vs etc. inside `StorageController` by having everyone else use only the `StorageController` API

i think ultimately it's only used for `MessageStore.prototype.query` inside a few of the `ProtocolAuthorization` methods, so we should be able to replace `messageStore.query` with `storageController.queryMessages` and be good

apologies for sending you down a different path.  lmk if you think what we have now is better and we can discuss further :)

src/store/storage-controller.ts
```diff
@@ -111,19 +109,53 @@ export class StorageController {
...
+
+  public getData(tenant: string, messageCid: string, dataCid: string): Promise<GetResult|undefined> {
+    return this.dataStore.get(tenant, messageCid, dataCid);
+  }
+
+  public appendEvent(tenant: string, messageCid: string): Promise<string>|undefined {
+    return this.eventLog.append(tenant, messageCid);
+  }
+
+  public deleteEvents(tenant: string, deletedMessageCids: any): Promise<number>|undefined {
```
dcrousso, line 149: should this be `Array<string>`?

also, i wonder if TypeScript has a way to do a "passthrough" for scenarios like this where we dont have to declare the types 🤔

src/store/storage-controller.ts
```diff
@@ -111,19 +109,53 @@ export class StorageController {
...
+
+  public getMessage(tenant: string, messageCid: string): Promise<BaseMessage|undefined> {
+    return this.messageStore.get(tenant, messageCid);
+  }
+
+  public putMessage(
+    tenant: string,
+    messageJson: BaseMessage,
+    indexes: Record<string, string>,
+    options?: MessageStoreOptions): Promise<void> {
```
dcrousso, line 133: i think we normally put the `): Promise<void> {` onto a separate line indented to match the line with the `(`

src/interfaces/records/handlers/records-read.ts
diehuxx: +1. You beat me to it. I made a similar suggestions above.

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/931ab6bd4e5079229f6b9663de5a442eedfe4983

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/931ab6bd4e5079229f6b9663de5a442eedfe4983

src/interfaces/records/handlers/records-read.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/931ab6bd4e5079229f6b9663de5a442eedfe4983

src/interfaces/records/handlers/records-read.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/931ab6bd4e5079229f6b9663de5a442eedfe4983

src/core/protocol-authorization.ts
```diff
@@ -351,7 +351,8 @@ export class ProtocolAuthorization {
    * Currently the only check is: if the write is not the initial write, the author must be the same as the initial write
    * @throws {Error} if fails verification
    */
-  private static async verifyActionCondition(tenant: string, incomingMessage: RecordsRead | RecordsWrite, messageStore: MessageStore): Promise<void> {
+  private static async verifyActionCondition(
+    tenant: string, incomingMessage: RecordsRead | RecordsWrite, storageController: StorageController): Promise<void> {
```
dcrousso, line 355: pretty sure that if you have to have a separate line we want all the parameters to be on their own line
``` suggestion
  private static async verifyActionCondition(
    tenant: string,
    storageController: StorageController
  ): Promise<void> {
```

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -1,8 +1,9 @@
-import type { EventLog } from '../../../event-log/event-log.js';
+import type { DidResolver } from '../../../index.js';
 import type { MethodHandler } from '../../types.js';
 import type { RecordsDeleteMessage } from '../types.js';
+import type { StorageController } from '../../../store/storage-controller.js';
 import type { TimestampedMessage } from '../../../core/types.js';
-import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
+
```
dcrousso, line 6: unnecessary newline

src/interfaces/permissions/handlers/permissions-request.ts
```diff
@@ -1,14 +1,16 @@
+import type { DidResolver } from '../../../index.js';
 import type { MethodHandler } from '../../types.js';
 import type { PermissionsRequestMessage } from '../types.js';
-import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
+import type { StorageController } from '../../../store/storage-controller.js';
+
```
dcrousso, line 5: unnecessary newline

src/interfaces/records/handlers/records-query.ts
```diff
@@ -1,19 +1,19 @@
...
 import type { MethodHandler } from '../../types.js';
-import type { DataStore, DidResolver, MessageStore } from '../../../index.js';
 import type { RecordsQueryMessage, RecordsQueryReplyEntry, RecordsWriteMessage } from '../types.js';
 
 import { authenticate } from '../../../core/auth.js';
 import { lexicographicalCompare } from '../../../utils/string.js';
 import { MessageReply } from '../../../core/message-reply.js';
 import type { RecordsWriteMessageWithOptionalEncodedData } from '../../../store/storage-controller.js';
-import { StorageController } from '../../../store/storage-controller.js';
+import type { StorageController } from '../../../store/storage-controller.js';
```
dcrousso, line 9: i think this should be moved above with all the other `import type`

src/interfaces/records/handlers/records-read.ts
```diff
@@ -9,10 +10,9 @@ import { MessageReply } from '../../../core/message-reply.js';
 import { RecordsRead } from '../messages/records-read.js';
 import { RecordsWrite } from '../messages/records-write.js';
 import { DwnInterfaceName, DwnMethodName } from '../../../core/message.js';
-
```
dcrousso, line 12: i think we still want a newline here

src/interfaces/records/records-interface.ts
```diff
@@ -17,9 +14,7 @@ export async function deleteAllOlderMessagesButKeepInitialWrite(
   tenant: string,
   existingMessages: TimestampedMessage[],
   comparedToMessage: TimestampedMessage,
-  messageStore: MessageStore,
-  dataStore: DataStore,
-  eventLog: EventLog
+  storageController: StorageController,
```
dcrousso, line 17: i dont think we normally have a trailing `,` after the last parameter

src/store/message-store.ts
```diff
@@ -22,7 +22,7 @@ export interface MessageStore {
   put(
     tenant: string,
     messageJson: BaseMessage,
-    indexes: { [key: string]: string },
+    indexes: Record<string, string>,
```
dcrousso, line 25: neat!  there's probably a bunch of other places where we could/should do this.  @duncanmak can you file an issue for updating all those too?  :)

src/store/storage-controller.ts
```diff
@@ -111,19 +109,46 @@ export class StorageController {
...
     if (message.descriptor.dataCid !== undefined) {
-      await dataStore.delete(tenant, messageCid, message.descriptor.dataCid);
+      await this.dataStore.delete(tenant, messageCid, message.descriptor.dataCid);
     }
 
-    await messageStore.delete(tenant, messageCid);
+    await this.messageStore.delete(tenant, messageCid);
+  }
+
+  public getMessage(tenant: string, messageCid: string): Promise<BaseMessage|undefined> {
```
dcrousso, line 125: i think we normally put spaces around the ` | `
```suggestion
  public getMessage(tenant: string, messageCid: string): Promise<BaseMessage | undefined> {
```

src/store/storage-controller.ts
```diff
@@ -111,19 +109,46 @@ export class StorageController {
...
+    options?: MessageStoreOptions
+  ): Promise<void> {
+    return this.messageStore.put(tenant, messageJson, indexes, options);
+  }
+
+  public queryMessages(tenant: string, filter: Filter, options?: MessageStoreOptions): Promise<BaseMessage[]> {
+    return this.messageStore.query(tenant, filter, options);
+  }
+
+  public getData(tenant: string, messageCid: string, dataCid: string): Promise<GetResult|undefined> {
```
dcrousso, line 142: ditto ([:125](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-cbe02e4d9564ce57729575f2fce1680efec7c8f72fe777276201b785fc0b3e8dR125))
```suggestion
  public getData(tenant: string, messageCid: string, dataCid: string): Promise<GetResult | undefined> {
```

src/store/storage-controller.ts
```diff
@@ -111,19 +109,46 @@ export class StorageController {
...
+    return this.dataStore.get(tenant, messageCid, dataCid);
+  }
+
+  public appendEvent(tenant: string, messageCid: string): Promise<string>|undefined {
```
dcrousso, line 146: i dont think this needs ` | undefined` based on the signature of `EventLog`
```suggestion
  public appendEvent(tenant: string, messageCid: string): Promise<string> {
```

src/store/storage-controller.ts
```diff
@@ -111,19 +109,46 @@ export class StorageController {
...
+    return this.eventLog.append(tenant, messageCid);
+  }
+
+  public deleteEvents(tenant: string, deletedMessageCids: string[]): Promise<number>|undefined {
```
dcrousso, line 150: ditto ([:146](https://github.com/TBD54566975/dwn-sdk-js/pull/338/files#diff-cbe02e4d9564ce57729575f2fce1680efec7c8f72fe777276201b785fc0b3e8dR146))
```suggestion
  public deleteEvents(tenant: string, deletedMessageCids: string[]): Promise<number> {
```

duncanmak: Obsoleted by https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/f872016b3691c4b2f4465ca59aaa33507847a820

src/store/storage-controller.ts
duncanmak: Obsoleted by https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/f872016b3691c4b2f4465ca59aaa33507847a820

src/store/storage-controller.ts
duncanmak: Obsoleted by https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/f872016b3691c4b2f4465ca59aaa33507847a820

src/store/storage-controller.ts
duncanmak: Obsoleted by https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/f872016b3691c4b2f4465ca59aaa33507847a820

src/interfaces/records/handlers/records-read.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/interfaces/records/handlers/records-query.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/interfaces/permissions/handlers/permissions-request.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/interfaces/records/handlers/records-delete.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/store/storage-controller.ts
```diff
@@ -14,6 +14,9 @@ import { DwnError, DwnErrorCode } from '../core/dwn-error.js';
  * A class that provides an abstraction for the usage of BlockStore and DataStore.
  */
 export class StorageController {
+
+  constructor(private messageStore: MessageStore, private dataStore: DataStore, private eventLog: EventLog) {}
```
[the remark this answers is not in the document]
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/store/storage-controller.ts
duncanmak: Addressed in https://github.com/TBD54566975/dwn-sdk-js/pull/338/commits/42f36fb0385704079ff7f0a11a32f953d80a25c7

src/interfaces/records/records-interface.ts
duncanmak: Addressed in 7bc13ef

src/store/message-store.ts
duncanmak: https://github.com/TBD54566975/dwn-sdk-js/issues/339

dcrousso (dismissed):
awesome work!  thanks for iterating with me :)

src/core/protocol-authorization.ts
```diff
@@ -24,18 +24,18 @@ export class ProtocolAuthorization {
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
     requesterDid: string | undefined,
-    messageStore: MessageStore
+    storageController: StorageController
```
thehenrytsai, line 27: This interesting to me. My interpretation of the original issue is that `StorageController` is the "better" abstraction for various handlers which made sense to me because instead of passing multiple interfaces all handlers need, define an abstraction that encapsulate those interfaces and simply pass one abstraction to the handlers.

However this PR seems to have taken one step further and elevated `StorageController` as THE abstraction that all code should interface with when it comes to all things store related.

Have not made up my mind yet on if this new model is better or worse, but it is clearly that in the case of protocol authorization here, it only needed messages to do its job, but is now given greater access to all stores unnecessarily, which goes against the generally accepted good programming principle of interface segregation.

diehuxx: I agree with all of @thehenrytsai's points, except that I'd go a step further to say I definitely do not like turning `StorageController` into THE abstraction. `StorageController` shines best IMO as a util for common operations that involve multiple stores. In all other cases, it is better (more readable, explicit, modular) to pass the underlying store.

United storage into a single API does not come off as easier to understand or more wieldy than having distinct, separately-concerned storage interfaces.

decentralgabe: don't understand well enough to give a coherent response but happy to jump on a call to better understand

diehuxx:
@duncanmak Sorry to keep this PR in limbo! @thehenrytsai and I are a bit swamped this week and @dcrousso is out of office, but we promise to come to a decision on this PR as soon as we are can. We haven't forgotten you!

diehuxx:
@duncanmak Closing this since Henry, myself, and other folks won't have a time to have a full discussion til after next week. By that point, `main` might look pretty different and it'll be easier to start a fresh PR. We'll add discussion in #288. Thanks for bearing with us :)

PR #341, opened 2023-05-04
add myself as a code owner

dcrousso:


thehenrytsai:
Hey @dcrousso, I think by adding you as a code owner, it gives you the permission to merge pull requests into this repo right? If that's the case, I think we should restrict that power to the DRIs of the repo (@diehuxx and I).

More than happy to (and we should) add you to the list of contributors though!

diehuxx:
> More than happy to (and we should) add you to the list of contributors though!

+1! @dcrousso We want to make sure you're getting credit where it's due :)

PR #342, opened 2023-05-05
Protocol definition semantic fields: protocolPath -> of, actions -> can

diehuxx:
@csuwildcat :)

Final decision, we like this naming convention better. Some reasons (other than vibes):
* We anticipate other properties that take a protocol path -- e.g. `under` -- so `protocolPath` is too generic of a property name.
* `can` is better than `to` because `to` has two potential meanings in English. First, preceding an infinite, e.g. "to be or not to be". Second, as a preposition, e.g. "from the moon to the stars". This overloading is not my favorite.

### DIFF
--- a/json-schemas/protocol-rule-set.json
+++ b/json-schemas/protocol-rule-set.json
@@ -11,7 +11,7 @@
         "type": "object",
         "required": [
           "actor",
-          "actions"
+          "can"
         ],
         "properties": {
           "actor": {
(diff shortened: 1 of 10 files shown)


thehenrytsai (approved):

diehuxx:
Revisiting after some discussion with other DWN folks. I'll add a full update in #326.

PR #343, opened 2023-05-06
Require protocol uri in protocol definition

diehuxx:
Require URL to protocol inside the definition itself. We do NOT normalize this URL because we are not indexing on it. For indexing, we still use the `protocol` field pass in the descriptor of messages such as `RecordsWrite`, `ProtocolsConfigure`, and `ProtocolsQuery`.


diehuxx:
Made a slight design oops. Gimme one second to fix.

Issue #139, opened 2022-11-04
move protocol definition out of the `descriptor`


thehenrytsai:

thehenrytsai:
This has been postponed indefinitely.

PR #348, opened 2023-05-08
Protocols: `recordDefinitions` array -> `types` object

diehuxx:
Next step in protocol DSL overhaul.
* Rename `recordDefinitions` -> `types`
* Restructure `types` as an object instead of an array
* Make `dataFormat` required in each `type`

In the next PR, I will update the protocol definition to rename the `records` property at the top level, and remove `records` from nested properties.


README.md
```diff
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-94.22%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-93.94%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-92.02%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-94.22%25-brightgreen.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-94.21%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-93.91%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-92%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-94.21%25-brightgreen.svg?style=flat)
```
diehuxx, line 6: I think the main driver of test coverage drop is that `utils/protocols.ts` had 100% coverage and I've now removed it. Going to look at it again today to see if there's another culprit.

diehuxx:
> 3\. This would be officially be the first precedence of `descriptor` having objects as properties, we'd likely need to add logic to descriptor indexing logic as right now it assumes `descriptor` properties are all strings.

Confused by this point. Isn't `definition` already part of the `descriptor`? This essentially splits `definition` into two properties (`definition` and `types`) which are both still in the `descriptor.

thehenrytsai:
> Confused by this point. Isn't definition already part of the descriptor? This essentially splits definition into two properties (definition and types) which are both still in the `descriptor.

#349 probably explains it. Basically, because of issue #139, the code on indexing is super wishy-washy, with a bunch of `as any` and as long as there is nothing blowing up immediately, we can address it as part of #139, but if #139 is going away, we'd need to address it.

src/interfaces/protocols/types.ts
```diff
@@ -10,16 +10,19 @@ export type ProtocolsConfigureDescriptor = {
...
-export type ProtocolRecordDefinition = {
-  id: string,
+export type ProtocolType = {
   schema?: string,
-  dataFormats?: string[],
+  dataFormats: string[],
+};
+
+export type ProtocolTypes = {
+  [key: string]: ProtocolType;
```
thehenrytsai, line 25: Discussion: Given the merge of https://github.com/TBD54566975/dwn-sdk-js/commit/6bc148c72eaee2e8fd35b63753a764bad98661e5, do you think we should use `Record<T, U>` consistently?

Personally I much prefer the `[key:string]: T` style because `Record` is a first class concept in DWN and can easily be confused with `Record<T, U>` for people not familiar with TypeScript, but I never got around to review the PR.

If you agree we can revert the `Record<T, U>` changes in separate PR.

src/core/protocol-authorization.ts
```diff
@@ -254,26 +253,26 @@ export class ProtocolAuthorization {
     const recordsWriteMessage = inboundMessage as RecordsWriteMessage;
 
     const protocolPath = recordsWriteMessage.descriptor.protocolPath!;
-    const recordDefinitionId = ProtocolAuthorization.getRecordDefinitionId(protocolPath);
-    // existence of recordDefinition has already been verified
-    const recordDefinition: ProtocolRecordDefinition = Protocols.getRecordDefinition(protocolDefinition, recordDefinitionId)!;
+    const typeName = ProtocolAuthorization.getTypeId(protocolPath);
+    // existence of type has already been verified
```
thehenrytsai, line 257: Nit:

```suggestion
    // existence of `protocolType` has already been verified
```

src/core/protocol-authorization.ts
```diff
@@ -254,26 +253,26 @@ export class ProtocolAuthorization {
     const recordsWriteMessage = inboundMessage as RecordsWriteMessage;
 
     const protocolPath = recordsWriteMessage.descriptor.protocolPath!;
-    const recordDefinitionId = ProtocolAuthorization.getRecordDefinitionId(protocolPath);
-    // existence of recordDefinition has already been verified
-    const recordDefinition: ProtocolRecordDefinition = Protocols.getRecordDefinition(protocolDefinition, recordDefinitionId)!;
+    const typeName = ProtocolAuthorization.getTypeId(protocolPath);
```
thehenrytsai, line 256: I guess concept of ID no longer applies?
Consider:

```suggestion
    const typeName = ProtocolAuthorization.getTypeName(protocolPath);
```
or
```suggestion
    const typeName = ProtocolAuthorization.getType(protocolPath);
```

src/core/protocol-authorization.ts
```diff
@@ -53,10 +52,10 @@ export class ProtocolAuthorization {
       ancestorMessageChain
     );
 
-    // Verify `dataFormat` and `schema` for the given `recordDefinition`
-    ProtocolAuthorization.verifyRecordDefinition(
+    // Verify `dataFormat` and `schema` for the given `type`
```
thehenrytsai, line 55: Nit:
```suggestion
    // Verify `dataFormat` and `schema` for the given `types`
```
or
```suggestion
    // Verify the type of the given message
```
or just remove the comment

json-schemas/protocol-definition.json
```diff
@@ -4,29 +4,26 @@
...
+        ".*": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "schema": {
               "type": "string"
+            },
+            "dataFormats": {
+              "type": "array",
+              "minItems": 1,
```
thehenrytsai, line 23: Sanity: was it discussed that `dataFormats` is a required constraint?

json-schemas/protocol-definition.json
```diff
@@ -4,29 +4,26 @@
   "type": "object",
   "additionalProperties": false,
   "required": [
-    "recordDefinitions",
+    "types",
     "records"
```
frankhinek, line 8: change to `structure` or is that the next PR?

diehuxx: That's in the next one

src/interfaces/protocols/types.ts
diehuxx: Let's keep `[key: string]`. I like your point about `Record` already having meaning in dwn-sdk. In other contexts I usually prefer `Record` as I find it more readable.

src/core/protocol-authorization.ts
diehuxx: update to `getTypeName`

json-schemas/protocol-definition.json
diehuxx: Update from discussion earlier today: we were originally planning on making it required, and now we've changed our minds to leave it optional

diehuxx: Added a commit making `dataFormats` optional again and adding a test verifying it's optional

src/core/protocol-authorization.ts
diehuxx: removed comment

src/core/protocol-authorization.ts
```diff
@@ -214,7 +213,7 @@ export class ProtocolAuthorization {
 
     const typeIds = Object.keys(types);
     const declaredProtocolPath = (inboundMessage as RecordsWrite).message.descriptor.protocolPath!;
-    const declaredTypeId = ProtocolAuthorization.getTypeId(declaredProtocolPath);
+    const declaredTypeId = ProtocolAuthorization.getTypeName(declaredProtocolPath);
```
thehenrytsai, line 216: Nit:
Do a search on `typeId/TypeId` and rename to `typeName/TypeName`

thehenrytsai (approved):
🙏

PR #350, opened 2023-05-08
Move type declarations

mistermoe:
* currently type declarations are generated in `cjs/` and `esm/`. This causes issues for downstream libraries attempting to package up their own type declarations. this PR moves type declarations to `dist/types.
* exports `RecordsWriteDescriptor` so that it can be used by other libs.


thehenrytsai (approved):
Fact: if build related changes are good for you, it must be good for me!
🙏 🙏 🙏

thehenrytsai (approved):

PR #355, opened 2023-05-10
Remove records keyword from protocol definitions

diehuxx:
Last change we have planned for the protocol definition DSL. Remove `records` from `ProtocolRuleSets`.

There are some tradeoffs about different ways to structure `type ProtocolRuleSet` so it is ergonomic for consumers of `dwn-sdk-js`. The requirements are:
1. The `$actions` field must be optional and will have type `ProtocolActionRule` when it is present.
2. Any other string property will have type `ProtocolRuleSet`.

### Option 1: `any`
In my opinion, this is the least gross option. All other options entail writing unsightly typescript.

Benefit: Doesn't require lots of `(...as ProtocolActionRule)` or `(...as ProtocolRuleSet), and doesn't involve weird `Symbol` conversion logic.
Drawback: Least strict typing. Not very different from vanilla JS experience.

```typescript
export type ProtocolRuleSet = {
  $actions?: ProtocolActionRule[];
  // JSON Schema verifies that properties other than `$actions` will actually have type ProtocolRuleSet
  [key: string]: any;
}

// The following all transpile, but results may have type `any`
const ruleSet: ProtocolRuleSet = {
  $actions: [{ who: 'anyone', can: 'read' }],
  ask: {
    $actions: [{ who: 'anyone', can: 'read' }],
    offer: {
      $actions: [{ who: 'anyone', can: 'read' }],
      fulfillment: {}
    }
  }
};

ruleSet.$actions;
ruleSet.$actions![0].who;
ruleSet.ask.foo.fulfillment; // still transpiles because `ask` is `any`
(ruleSet.ask as ProtocolRuleSet).$actions?.[0].who;
```

### Option 2: Symbols
Benefit: Allows good typing of each nested `ProtocolRuleSet`
Drawback: Every time we ingest or send a protocol definition, we need to parse through it and replace string `$actions` with the symbol. I find this kinda gross and error prone to need to call `toSymbol(protocol)` and `stringToSymbol(protocol)` throughout the code and in tests. More importantly app developers would need to make this conversion all over the place too.

```typescript
const $ACTIONS = Symbol('$actions');

export type ProtocolRuleSet = {
  [$ACTIONS]?: ProtocolActionRule[];
  [key: string]: ProtocolRuleSet;
}

const ruleSet: ProtocolRuleSet = {
  [$ACTIONS]: [{ who: 'anyone', can: 'read' }],
  ask: {
    [$ACTIONS]: [{ who: 'anyone', can: 'read' }],
    offer: {
      [$ACTIONS]: [{ who: 'anyone', can: 'read' }],
      fulfillment: {}
    }
  }
};

ruleSet[$ACTIONS];
ruleSet[$ACTIONS]?.[0].who;
ruleSet.ask.offer.fulfillment;
ruleSet.ask[$ACTIONS]?.[0].who;
```

### Option 3: Too many `as`
```typescript
export type ProtocolRuleSet = {
  $actions?: ProtocolActionRule[];
  [key: string]: ProtocolRuleSet | ProtocolActionRule[] | undefined;
}

const ruleSet: ProtocolRuleSet = {
  $actions: [{ who: 'anyone', can: 'read' }],
  ask: {
    $actions: [{ who: 'anyone', can: 'read' }],
    offer: {
      actions: [{ who: 'anyone', can: 'read' }],
      fulfillment: {}
    }
  }
};

ruleSet.$actions;
ruleSet.$actions?.[0].who;
// ruleSet.ask.offer.fulfillment; // Does not transpile
((ruleSet.ask as ProtocolRuleSet)?.offer as ProtocolRuleSet)?.fulfillment;
((ruleSet.ask as ProtocolRuleSet).$actions as (ProtocolActionRule | undefined))?.[0].who;
```

### DIFF
--- a/json-schemas/protocol-rule-set.json
+++ b/json-schemas/protocol-rule-set.json
@@ -61,14 +61,11 @@
           }
         ]
       }
-    },
-    "records": {
-      "type": "object",
-      "patternProperties": {
-        ".*": {
-          "$ref": "https://identity.foundation/dwn/json-schemas/protocol-rule-set.json"
-        }
-      }
+    }
+  },
+  "patternProperties": {
+    "^[^$].*": {
+      "$ref": "https://identity.foundation/dwn/json-schemas/protocol-rule-set.json"
     }
   }
 }
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -178,11 +178,11 @@ export class ProtocolAuthorization {
     const protocolPathArray = protocolPath.split('/');
 
     // traverse rule sets using protocolPath
-    let currentRuleSet: ProtocolRuleSet = { records: protocolDefinition.structure };
+    let currentRuleSet: ProtocolRuleSet = protocolDefinition.structure;
     let i = 0;
     while (i < protocolPathArray.length) {
       const currentTypeName = protocolPathArray[i];
-      const nextRuleSet = currentRuleSet.records?.[currentTypeName];
+      const nextRuleSet: ProtocolRuleSet | undefined = currentRuleSet[currentTypeName];
 
       if (nextRuleSet === undefined) {
         const partialProtocolPath = protocolPathArray.slice(0, i + 1).join('/');
--- a/src/interfaces/protocols/types.ts
+++ b/src/interfaces/protocols/types.ts
@@ -36,15 +36,16 @@ export enum ProtocolAction {
   Write = 'write'
 }
 
+export type ProtocolActionRule = {
+  who: string,
+  of?: string,
+  can: string
+};
+
 export type ProtocolRuleSet = {
-  $actions?: {
-    who: string,
-    of?: string,
-    can: string
-  }[];
-  records?: {
-    [key: string]: ProtocolRuleSet;
-  }
+  $actions?: ProtocolActionRule[];
+  // JSON Schema verifies that properties other than `$actions` will actually have type ProtocolRuleSet
+  [key: string]: any;
 };
 
 export type ProtocolsConfigureMessage = BaseMessage & {
--- a/tests/interfaces/records/handlers/records-write.spec.ts
+++ b/tests/interfaces/records/handlers/records-write.spec.ts
@@ -1266,9 +1266,9 @@ describe('RecordsWriteHandler.handle()', () => {
         // create an invalid ancestor path that is longer than possible
         const invalidProtocolDefinition = { ...credentialIssuanceProtocolDefinition };
         const actionRuleIndex =
-          invalidProtocolDefinition.structure.credentialApplication.records.credentialResponse.$actions
+          invalidProtocolDefinition.structure.credentialApplication.credentialResponse.$actions
             .findIndex((actionRule) => actionRule.who === ProtocolActor.Recipient);
-        invalidProtocolDefinition.structure.credentialApplication.records.credentialResponse
+        invalidProtocolDefinition.structure.credentialApplication.credentialResponse
           .$actions[actionRuleIndex].of
             = 'credentialResponse';
         // this is invalid as the root ancestor can only be `credentialApplication` based on record structure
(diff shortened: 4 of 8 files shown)


thehenrytsai:
Sorry, for beating a dead horse... Let me know if I read the above wrong: we chose option 1, so TypeScript will not be able to assist developers at code time when crafting the protocol definition, and error will be thrown at run time.

@csuwildcat and @frankhinek, just so you are aware, we are removing the `records` layering in the name of "cleaner/better dev experience" by removing type safety guardrails when creating the protocol definition that developers would arguably find more useful in actual dev experience. It can be counter-argued that they won't always be using the SDK to craft the definitions and pure JS developers will not be affected because they don't get type safety to begin with.

Regardless, this tradeoff was not foreseen when the removing the `records` layers was discussed. As long as you are aware and okay, I will sign off.

diehuxx:
> Regardless, this tradeoff was not foreseen when the removing the `records` layers was discussed. As long as you are aware and okay, I will sign off.

We did to some extent discuss the tradeoffs of Options 2 and 3 on Tuesday. I wasn't happy with either of those approaches. I went with Option 1 after giving it some more thought by myself. That said, I'm still open to either Options 2 or 3 (or some else entirely) if we feel those are better.

thehenrytsai:
> We did to some extent discuss the tradeoffs of Options 2 and 3 on Tuesday. I wasn't happy with either of those approaches. I went with Option 1 after giving it some more thought by myself. That said, I'm still open to either Options 2 or 3 (or some else entirely) if we feel those are better.

I am with you on what option 2 and 3 not being better. But you also know what my preferred option 4 is at this point. I am at peace, as long as @csuwildcat and @frankhinek are explicitly aware of the final tradeoffs, because I wasn't, I was under the impression that option 2 would give pure TS dev experience at the expense of ultra ugly backend implementation, but that seems not the case.

mistermoe:
i'd only use option 2 if we can tuck away the extra step needed to serialize and deserialize an object with symbols. 

Generally speaking, i think to some extent we've conflated two different discussions:
1. the general structure of protocol definitions absent of the _implementation_ of them in a specific language
2. _our_ SDKs implementation of protocol definitions.

re: 1
i believe that protocols will be referred to by developers as a form of "documentation" in plain json far more than they're created in actual code. I also feel like protocols and their understanding are critical to the success and adoption of DWNs, Given this, i would opt in favor of _not_ resorting back to the recursive `structure` approach. i think we all agree that the recusing structures is ugmugz and less human readable:
```javascript
{
  $actions: [],
  structure: {
    $actions: [],
    structure: {
      $actions: [],
      structure: {
        $actions: [],
        structure: {
          $actions: [],
          structure: {}
        }
      }
    }
  }
}
```
re: 2
i think there's a 4th option that we haven't yet explored which is the idea of creating a new class that is dedicated to streamlining the construction of protocol definitions. i think we can achieve the developer experience and provide the guardrails we want if we go down the path of a dedicated class with method chaining. 

Also, have we looked into how `ajv` achieves type safety? don't json schemas in general run into the same problem?

thehenrytsai:
Thanks @mistermoe for the insights!

> I believe that protocols will be referred to by developers as a form of "documentation" in plain json far more than they're created in actual code.

Agreed, given this, which optionswe choose becomes more of an academic exercise not seen by external devs for the most part, and @diehuxx and I are fine with option 1, so imma sign off.

We can revisit option 2, 4, `ajv` what not for a nicer impl, though I doubt we'll feel the urge to do so anytime soon.

csuwildcat:
No poopy wrapper recursion 💩

thehenrytsai (approved):

KendallWeihe:
> i believe that protocols will be referred to by developers as a form of "documentation" in plain json far more than they're created in actual code

I'm naive w/ these concepts, but yeah... that would be quite strange to programmatically, which is to say dynamically, create new protocols. I mean... I suppose you could have user-generated protocols... but that's what programming is for, not so much a normie GUI-experience. Perhaps there is an end user experience which my imagination is lacking.

PR #356, opened 2023-05-11
deps: update all ipfs-* deps to the latest versions

achingbrain:
A few updates and changes to support the APIs of the latest ipfs-* deps.


package.json
```diff
@@ -61,15 +61,16 @@
...
     "flat": "^5.0.2",
-    "interface-blockstore": "4.0.1",
-    "ipfs-unixfs": "6.0.9",
-    "ipfs-unixfs-exporter": "7.0.11",
-    "ipfs-unixfs-importer": "14.0.1",
+    "interface-blockstore": "5.2.0",
+    "ipfs-unixfs": "11.0.0",
+    "ipfs-unixfs-exporter": "13.1.4",
+    "ipfs-unixfs-importer": "15.1.4",
+    "it-to-browser-readablestream": "^2.0.4",
```
diehuxx, line 73: @achingbrain Did you add this yourself? Looks like it wasn't here before. I tried removing it and running tests. Tests seem to pass just fine without it.

thehenrytsai: @diehuxx, I think @achingbrain is helping us complete #245, which is a much needed help!!!

achingbrain: Ah, looks like I added it during some experimentation - I've removed it along with `ipfs-unixfs` which is unused here.

Looks like there are a few other unused production deps too but that seems outside of the scope here:

* @scure/base
* @swc/helpers
* ajv
* date-fns
* ms
* randombytes

diehuxx: Thanks! Agreed, we can remove those in another PR, thanks for flagging.

diehuxx (approved):
ty!

thehenrytsai (approved):
@achingbrain, 🙏 🙏 🙏

PR #361, opened 2023-05-12
Refactor Dwn#processMessage and other top level handlers

diehuxx:
I think a few of us have talked about how we don't love the current pattern for DWN method-specific handling methods like `Dwn#handleRecordsRead` and `Dwn#handlerMessagesGet`. Here's an attempt at a different pattern that doesn't involve just casting the result with `as`.

Benefits are:
* We keep `Dwn#processMessage` as loosely-typed, convenient "do-it-all" API
* We have different `Dwn#handle<Interface><Method>` for stricter typing
* Pretty DRY

This PR is only moving the existing `Dwn#handleMessagesGet` and `Dwn#handleRecordsRead` to the new pattern. We can add the other methods in a later PR if we like this pattern.


thehenrytsai (approved):
Love it! Updated `synchronizePrunedInitialRecordsWrite()` accordingly.

Issue #359, opened 2023-05-11
Unauthorized access of data through referencing of `dataCid` in `RecordsWrite`


thehenrytsai:
Below is a test repro of Bob being able to access Alice's private data by referencing `dataCid`.

```typescript
it('should not allow `RecordsWrite` without data stream unless it is a modify and the `dataCid` is unchanged', async () => {
    const alice = await DidKeyResolver.generate();
    const bob = await DidKeyResolver.generate();

    const dataForCid = await dataStore.blockstore.partition('data');

    // alice writes a private record
    const dataString = 'private data';
    const dataSize = dataString.length;
    const data = Encoder.stringToBytes(dataString);
    const dataCid = await Cid.computeDagPbCidFromBytes(data);

    const { message, dataStream } = await TestDataGenerator.generateRecordsWrite({
      requester: alice,
      data,
    });

    const reply = await dwn.processMessage(alice.did, message, dataStream);
    expect(reply.status.code).to.equal(202);

    expect(asyncGeneratorToArray(dataForCid.db.keys())).to.eventually.eql([ dataCid ]);

    const protocol = 'https://tbd.website/decentralized-web-node/protocols/social-media';
    const protocolDefinition = socialMediaProtocolDefinition;

    // alice has a social media protocol that allows anyone to write and read images
    const protocolsConfig = await TestDataGenerator.generateProtocolsConfigure({
      requester: alice,
      protocol,
      protocolDefinition
    });
    const protocolWriteReply = await dwn.processMessage(alice.did, protocolsConfig.message, protocolsConfig.dataStream);
    expect(protocolWriteReply.status.code).to.equal(202);

    // bob learns of metadata (ie. dataCid) of alice's secret data,
    // attempts to gain unauthorized access by write to alice's DWN through open protocol referencing the dataCid without supplying the data
    const imageRecordsWrite = await TestDataGenerator.generateRecordsWrite({
      requester    : bob,
      protocol,
      protocolPath : 'image',
      schema       : protocolDefinition.types.image.schema,
      dataFormat   : 'image/jpeg',
      dataCid, // bob learns of, and references alice's secrete data's CID
      dataSize,
      recipientDid : alice.did
    });
    const imageReply = await dwn.processMessage(alice.did, imageRecordsWrite.message, imageRecordsWrite.dataStream);
    expect(imageReply.status.code).to.equal(202); // should be disallowed

    // proof of unauthorized access
[...]

PR #363, opened 2023-05-13
#359 - fixed unauthorized access of data through dataCid reference in RecordsWrite

thehenrytsai:
1. Fixed unauthorized access of data through dataCid reference in RecordsWrite
2. Moved some misplaces tests from `records-delete.spec.ts` to `records-write.spec.ts`
3. Grouped and renamed a couple of array utility methods.


diehuxx (approved):
Looks generally good, thanks in particular for moving those tests!

I remain apprehensive about the object-oriented approach of `PrunedInitialRecordsWriteHandler`, and this PR is adding a bit more to that approach. For the sake of keeping you unblocked, I'm comfortable with this being merged. However, I took a stab at a refactor #364, which we can either merge into this PR or review as a new subsequent PR. Let's discuss this week.

PR #364, opened 2023-05-15
Refactor data storage for RecordsWrite

diehuxx:
An attempt at a refactor for pruned initial records writes that establishes a new pattern for message handlers.

First, seeking feedback on the general idea of `RecordsWriteHandlerOptions` and of migrating `StorageController.put` to `RecordsWrite#putData`. Then, I'll fix test coverage and code clean up.

## Changes
1. Delete class `PrunedInitialRecordsWritesHandler`. Replace that with an `options.skipDataStorage` parameter in `RecordsWriteHandler`. This establishes the potential for `options` in other handlers too. But this PR does not add them.
  a. I'm open to pushing this pattern even further: expose `options` as a parameter in `Dwn#processMessage`. This PR does not implement that, and if it is too big a can of worms for now, we can table it.
2. Move `StorageController.put` to `RecordsWrite#putData`. IMO, `StorageController.put` was becoming too complex and specific to be a useful/intuitive abstraction.
  a. Remove call to `StorageController.put` that did not involve `RecordsWrite`.
  a. Stricten checks in `RecordsWrite#putData`. Add a new error code and types for the `result` variable.

## Open Questions
* ~Why do we have test `storageController.put() throws unknown error`? Should I update or remove?~


src/interfaces/records/handlers/records-write.ts
```diff
@@ -103,41 +110,69 @@ export class RecordsWriteHandler implements MethodHandler {
...
         throw new DwnError(
           DwnErrorCode.RecordsWriteMissingDataStream,
           'Data stream is not provided.'
         );
       }
+
+      const associateResult = await this.dataStore.associate(tenant, messageCid, message.descriptor.dataCid);
+      if (associateResult === undefined) {
+        // Question for reviewers: I think if this throws, it is indicative of an underlying bug in the data store.
+        //   If previous RecordsWrites have gone correctly, this shouldn't happen, right?
```
thehenrytsai, line 145: I agree this condition should not happen under normal operation. Though I think it can also happen beyond a bug inside data store, under a contrived circumstance now that sync is at play:

1. a sync wrote a pruned initial `RecordsWrite` without data stream
1. A normal (not via sync) `RecordWrite` for the same record immediately follows referencing the same `dataCid` but also without data stream

So we could either keep the code you have to handle this case or alternatively we make `associate()` throw instead of returning `undefined`. Either case we should probably add a test to cover the scenario. The code you've got now is arguably still better.

thehenrytsai:
Love the objectively better approach, so totally onboard! A couple of comments related to design (without commenting on details since this is a WIP):

1. Just context: originally `StorageController` was created in anticipation that it would be used by other interface(s): ie. the `ProtocolsConfigure`, if protocol definition was external to the message and was referenced like record's data. Since we have postpone the move until needed, and as you said, it is becoming more specialized, it makes sense to bring the code back into `RecordsWrite`.
1. Have you considered passing `RecordsWriteHandlerOptions` to constructor instead? What are your thoughts on it?

thehenrytsai:
>  Is there a particular reason you might prefer it? If you just prefer it based on vibes, that's good enough for me.

No strong opinion, just curious what you think, though my immediate instinct when reading option being introduced was thinking you added to the constructor. It's a trade-off that we probably don't need to geek out over too much, but here I go... LOL

Pros if given at construction time
1. More locked down thus more safe, you need to know what kind of handler you want at the beginning and can't accidentally make unintended call.
2. `handle()` calls will be more succinct, not needing to pass the option every time.

Pros if given at `handler()` time
1. More flexible, we technically just have need to construct one handler, and use it in however way the internal code sees fit.

tests/interfaces/records/handlers/records-write.spec.ts
```diff
@@ -1987,23 +1986,25 @@ describe('RecordsWriteHandler.handle()', () => {
     });
   });
 
-  it('should throw if `storageController.put()` throws unknown error', async () => {
-    const { requester, message, dataStream } = await TestDataGenerator.generateRecordsWrite();
-    const tenant = requester.did;
+  // Question for reviewers (mainly thehenrytai): Why are were doing this?
```
thehenrytsai, line 1989: I think it is a test case for this line of code: https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/interfaces/records/handlers/records-write.ts#L92

The intend for that line of code is so that we don't claim that it is a client error when don't know what that error is and it can very well be a server side error. A case can be made that we should return 500 instead.

src/interfaces/records/handlers/records-write.ts
diehuxx: Added a test. Keeping the existing `associate()` behavior where it returns undefined for now.

diehuxx:
> Pros if given at `handler()` time
> 
> 1. More flexible, we technically just have need to construct one handler, and use it in however way the internal code sees fit.

I do prefer passing `options` at `handler()` time personally, so keeping it that way. I'm down to change it in the future if we change our minds.

README.md
```diff
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-94.71%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-94.16%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.15%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-94.71%25-brightgreen.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-94.65%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-94.16%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.06%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-94.65%25-brightgreen.svg?style=flat)
```
diehuxx, line 6: I think the drop in coverage is because I'm removing `pruned-initial-records-write.ts` which had 100% coverage.

src/interfaces/records/handlers/records-delete.ts
```diff
@@ -60,7 +60,7 @@ export class RecordsDeleteHandler implements MethodHandler {
       });
     }
 
-    // return Not Found if record does not exist or is already deleted
+    // return Not Founnd if record does not exist or is already deleted
```
thehenrytsai, line 63:
```suggestion
    // return Not Found if record does not exist or is already deleted
```

src/interfaces/records/handlers/records-write.ts
```diff
@@ -77,21 +79,26 @@ export class RecordsWriteHandler implements MethodHandler {
     const indexes = await constructRecordsWriteIndexes(recordsWrite, isLatestBaseState);
 
     try {
-      this.validateUndefinedDataStream(dataStream, newestExistingMessage, message);
-
-      await this.storeMessage(this.messageStore, this.dataStore, this.eventLog, tenant, message, indexes, dataStream);
+      // try to store data, unless options explicitly say to skip storage
+      if (!options?.skipDataStorage) {
```
thehenrytsai, line 83: Optional nit: I find it a bit confusing with the double-negative...

```suggestion
      if (options === undefined || !options.skipDataStorage) {
```

thehenrytsai (dismissed):

thehenrytsai (approved):
🙏

Issue #365, opened 2023-05-15
RecordsDelete should return 404 if record does not exist or is already deleted


thehenrytsai:

thehenrytsai:
Done.

PR #366, opened 2023-05-15
#365 - RecordsDelete should return 404 if record does not exist or is already deleted

thehenrytsai:


frankhinek:
LGTM -- thanks for knocking this out so quickly @thehenrytsai ! 🔥

diehuxx (approved):
speedy

PR #371, opened 2023-05-16
Added a Q&A document

thehenrytsai:
* added a Q&A document
* opportunistically updated `CONTRIBUTING.md`

### DIFF
--- a/Q_AND_A.md
+++ b/Q_AND_A.md
@@ -1,0 +1,24 @@
+
+# DWN Q&A
+
+## Message ID (`messageCid`) and Record ID (`recordId`)
+
+- Why can't/don't we use the message ID (`messageCid`) of the initial `RecordsWrite` as the record ID?
+
+  Because the initial `RecordsWrite` contains the `recordId` property, so this would be a "chicken-and-egg" problem. See question [why is `recordId` needed in the `RecordsWrite`](#why-is-record-id-needed) for more info.
+
+- Why is `recordId` needed in a `RecordsWrite`?
+
+  This question should be further split into two:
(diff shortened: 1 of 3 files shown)


Q_AND_A.md
```diff
@@ -0,0 +1,24 @@
+
+# DWN Q&A
```
diehuxx, line 2: Nit: What do you think of `FAQ` instead?

Q_AND_A.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Message ID (`messageCid`) and Record ID (`recordId`)
+
+- Why can't/don't we use the message ID (`messageCid`) of the initial `RecordsWrite` as the record ID?
+
+  Because the initial `RecordsWrite` contains the `recordId` property, so this would be a "chicken-and-egg" problem. See question [why is `recordId` needed in the `RecordsWrite`](#why-is-record-id-needed) for more info.
```
diehuxx, line 8: For brevity, i think we can leave out the second sentence (`See question`). It's directly below so they'll just keep reading

diehuxx (dismissed):

Q_AND_A.md
thehenrytsai: I added it in case the question gets edited/shuffled, and the continuation is lost thus answer isn’t providing sufficient context

Q_AND_A.md
thehenrytsai: i didn’t do it because I didn’t think these are “frequently” asked questions…. LOL

diehuxx (approved):

PR #377, opened 2023-05-19
Renamed requester to author

thehenrytsai:
I have been meaning to rename "requester" to "author" for a long time and never found the bandwidth and gap to do so without introducing annoying conflicts due to the constant streams of PRs flowing in.

We initially used the two terms interchangeably months ago, and overtime we have converged on "author" being objectively the better term (especially in tests) except for some specific contexts, thus using this gap to quickly switch over.

### DIFF
--- a/src/interfaces/records/handlers/records-query.ts
+++ b/src/interfaces/records/handlers/records-query.ts
@@ -76,13 +76,13 @@ export class RecordsQueryHandler implements MethodHandler {
   /**
    * Fetches the records as a non-owner, return only:
    * 1. published records; and
-   * 2. unpublished records intended for the requester (where `recipient` is the requester)
+   * 2. unpublished records intended for the query author (where `recipient` is the query author)
    */
   private async fetchRecordsAsNonOwner(tenant: string, recordsQuery: RecordsQuery)
     : Promise<RecordsWriteMessageWithOptionalEncodedData[]> {
     const publishedRecords = await this.fetchPublishedRecords(tenant, recordsQuery);
-    const unpublishedRecordsForRequester = await this.fetchUnpublishedRecordsForRequester(tenant, recordsQuery);
-    const unpublishedRecordsByRequester = await this.fetchUnpublishedRecordsByRequester(tenant, recordsQuery);
+    const unpublishedRecordsForRequester = await this.fetchUnpublishedRecordsForQueryAuthor(tenant, recordsQuery);
+    const unpublishedRecordsByRequester = await this.fetchUnpublishedRecordsByAuthor(tenant, recordsQuery);
     const records = [...publishedRecords, ...unpublishedRecordsForRequester, ...unpublishedRecordsByRequester];
     return records;
   }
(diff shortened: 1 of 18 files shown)


src/interfaces/records/handlers/records-query.ts
```diff
@@ -76,13 +76,13 @@ export class RecordsQueryHandler implements MethodHandler {
...
+   * 2. unpublished records intended for the query author (where `recipient` is the query author)
    */
   private async fetchRecordsAsNonOwner(tenant: string, recordsQuery: RecordsQuery)
     : Promise<RecordsWriteMessageWithOptionalEncodedData[]> {
     const publishedRecords = await this.fetchPublishedRecords(tenant, recordsQuery);
-    const unpublishedRecordsForRequester = await this.fetchUnpublishedRecordsForRequester(tenant, recordsQuery);
-    const unpublishedRecordsByRequester = await this.fetchUnpublishedRecordsByRequester(tenant, recordsQuery);
+    const unpublishedRecordsForRequester = await this.fetchUnpublishedRecordsForQueryAuthor(tenant, recordsQuery);
+    const unpublishedRecordsByRequester = await this.fetchUnpublishedRecordsByAuthor(tenant, recordsQuery);
     const records = [...publishedRecords, ...unpublishedRecordsForRequester, ...unpublishedRecordsByRequester];
```
diehuxx, line 86: These variable names and a few others in this file still have `Requester` even though you changed the function names.

diehuxx (commented):
I notice you left `protocol-authorization.ts` alone. I'm guessing that you left the variable `requesterDid` as-is because we use the word `author` to refer to the author in `$action` rules? I don't feel very strongly either way on that particular file, just clarifying.

tests/event-log/event-log-level.spec.ts
```diff
@@ -25,15 +25,15 @@ describe('EventLogLevel Tests', () => {
...
 
   it('separates events by tenant', async () => {
-    const { requester, message } = await TestDataGenerator.generateRecordsWrite();
+    const { author, message } = await TestDataGenerator.generateRecordsWrite();
     const messageCid = await Message.getCid(message);
-    const watermark = await eventLog.append(requester.did, messageCid);
+    const watermark = await eventLog.append(author.did, messageCid);
 
-    const { requester: requester2, message: message2 } = await TestDataGenerator.generateRecordsWrite();
+    const { author: requester2, message: message2 } = await TestDataGenerator.generateRecordsWrite();
```
diehuxx, line 32:
```suggestion
    const { author: author2, message: message2 } = await TestDataGenerator.generateRecordsWrite();
```

src/interfaces/records/handlers/records-query.ts
thehenrytsai: Thanks for seeing, changing now!

tests/event-log/event-log-level.spec.ts
thehenrytsai: Okay, I know why missed all the above hits... I had "match whole world" turned on in my search, which ignores any hits with prefixes and suffixes... changing now!

diehuxx (approved):

PR #381, opened 2023-05-24
Allow newer ProtocolsConfigures to overwrite old ones

diehuxx:
Fixing a bug after a conversation with @csuwildcat this morning.

Currently, we allow overwrites of `ProtocolsConfigure` only if the message CID of the new message is lexicographically larger than that of the existing message. This is weird and not intuitive.

This PR changes it to be more like `RecordsWrite`. If a `ProtocolsConfigure` has a later `dateCreated` then it overwrites the existing. Lexicographic compare of message CIDs is used as a tie-breaker when the incoming and existing `ProtocolsConfigure` have identical `dateCreated`.


csuwildcat:
🧑‍🍳🤌

thehenrytsai:
Haven't read the entire PR, but think I grasped the essence. The necessary logic duplication in this PR is due to the semantic difference between `dateCreated` in `RecordsWrite` and `ProtocolsConfigure`, which suggests that we should consider making them the same.

Adding more context, I was felt the meaning of `dateCreated` was NOT clear, is it the date the record was created or the date the message was created? We need both information for `RecordsWrite` and due to us [insisting](https://github.com/TBD54566975/technical-proposals/pull/2) on including full state in every message so that we can return the latest message as the latest state of a record, I had to introduce `dateModified` to mean the date the message was created, which kinda makes sense in the context of `RecordsWrite`.

Circling back to `ProtocolsConfigure`, if we were to have the same semantic meaning (which I advocate) for `dateCreated` across interfaces, renaming `dateCreated` to `dateModified` or adding `dateModified` in `ProtocolsConfigure` "should" make things work without duplicating logic.

If `dateModified` is a horrible name for people, we could consider renaming it (along with `dateCreated`), perhaps `date<Message/Record>Created`.

Happy to chat on anything!

diehuxx (commented):
Per discussion yesterday, I updated the PR to rename `dateCreated` -> `dateModified` in `ProtocolsConfigure` messages, so we can share the logic for comparing new messages with `RecordsWrite`.

This is a temporary state. This PR addresses a bug which should be resolved ASAP. However, we are in the process of designing how to handle overwritten protocols and the records associated with a given version* of a protocol.

*Don't read into the word "version" too much, we're still figuring out terminology there.

src/interfaces/protocols/handlers/protocols-configure.ts
```diff
@@ -43,9 +43,9 @@ export class ProtocolsConfigureHandler implements MethodHandler {
     const existingMessages = await this.messageStore.query(tenant, query) as ProtocolsConfigureMessage[];
 
     // find lexicographically the largest message, and if the incoming message is the largest
```
thehenrytsai, line 45: Nit: remove?

thehenrytsai (dismissed):
🥇

thehenrytsai (dismissed):

src/interfaces/protocols/handlers/protocols-configure.ts
diehuxx: good catch. updated.

thehenrytsai (approved):

PR #382, opened 2023-05-25
Moved and renamed types files in a common types folder

thehenrytsai:


README.md
```diff
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-94.65%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-94.16%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.06%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-94.65%25-brightgreen.svg?style=flat)
```
diehuxx, line 6: nice

diehuxx, line 6: nice

diehuxx (commented):
Was the intent just to move all `types.ts` files into a common `types/` folder or also to migrate the types defined in files other than the many `types.ts`? E.g. `dwn.ts` defines the `DwnConfig` type.

thehenrytsai:
In my mind, min-bar is that if you were gonna go out of your way to create a file that contains types, put it in somewhere common and name it accordingly.

A separate discussion can be had about whether we want to enforce types to ALWAYS to be defined in a separate file in the types folder, that seems more contentious and can be in a follow up PR (can also be in this PR if you prefer also).

diehuxx (approved):

README.md
thehenrytsai: 🙏

PR #383, opened 2023-05-25
index-level refactoring

thehenrytsai:
Refactored the Level implementation of data indexing in preparation for inevitable Protocol participant requirements:

1.  Refactored `put()` by simplifying how indexes are stored for deletion purposes
1. Refactored `query()` (thanks @diehuxx !!)
1. Refactored `findExactMatches()` and `findRangeMatches()` by removing `findMatches()`
1. Refactored `delete()`
1. Removed `Match`
1. Added a lot of comments/documentation
1. A lot of renames

### DIFF
--- a/src/store/index-level.ts
+++ b/src/store/index-level.ts
@@ -72,8 +72,6 @@ export class IndexLevel {
   }
 
   async query(filter: Filter, options?: IndexLevelOptions): Promise<Array<string>> {
-    const missingPropertiesForID: { [id: string]: Set<string> } = { };
-
     // Note: We need to have an array of Promise<Matches> in order to support OR queries
     const propertyNameToPromise: { [key: string]: Promise<Matches>[] } = {};
 
@@ -109,15 +107,23 @@ export class IndexLevel {
       }
     }
 
+    // map of ID of all data/object -> list of missing property matches
+    // if list of missing property matches is 0, then it data/object is fully matches the filter
+    const missingPropertyMatchesForId: { [id: string]: Set<string> } = { };
+
     // Resolve promises and find the union of results for each individual propertyName DB query
     const matchedIDs: string[] = [ ];
     for await (const [propertyName, promises] of Object.entries(propertyNameToPromise)) {
       for (const promise of promises) {
         for (const [ _, id ] of await promise) {
-          missingPropertiesForID[id] ??= new Set<string>([ ...Object.keys(filter) ]);
+          // if first time seeing a property match for object, add all the properties in the filter to missingPropertyMatchesForId to track
+          missingPropertyMatchesForId[id] ??= new Set<string>([ ...Object.keys(filter) ]);
 
-          missingPropertiesForID[id].delete(propertyName);
-          if (missingPropertiesForID[id].size === 0) {
+          missingPropertyMatchesForId[id].delete(propertyName);
+          if (missingPropertyMatchesForId[id].size === 0) {
+            // full filter match, add it to return list
+            // NOTE: there appears to be an edge case where if the last filter match is an array with multiple matching values,
+            // the same data/object ID would be added multiple times
             matchedIDs.push(id);
           }
         }
--- a/src/store/index-level.ts
+++ b/src/store/index-level.ts
@@ -4,10 +4,6 @@ import type { LevelWrapperBatchOperation, LevelWrapperIteratorOptions } from './
 import { flatten } from '../utils/object.js';
 import { createLevelDatabase, LevelWrapper } from './level-wrapper.js';
 
-export type Entry = {
-  [property: string]: unknown
-};
-
 export interface IndexLevelOptions {
   signal?: AbortSignal;
 }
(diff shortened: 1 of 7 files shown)


src/store/index-level.ts
```diff
@@ -37,75 +33,124 @@ export class IndexLevel {
...
+    for (const propertyName in indexes) {
+      const value = indexes[propertyName];
+      const key = this.join(propertyName, this.encodeValue(value), id);
+      ops.push({ type: 'del', key });
     }
-    ops.push({ type: 'del', key: `__${id}__prefixes` });
+
+    ops.push({ type: 'del', key: `__${id}__indexes` });
 
     return this.db.batch(ops, options);
```
diehuxx, line 155: clarifying: The return type of this method is `Promise<void>` but we have this return statement here. Did you keep that for a reason?

src/store/index-level.ts
```diff
@@ -37,75 +33,124 @@ export class IndexLevel {
...
+    options?: IndexLevelOptions
+  ): Promise<void> {
+
+    indexes = flatten(indexes);
+
+    const operations: LevelWrapperBatchOperation<string>[] = [ ];
+
+    // create an index entry for each property in the `indexes`
+    for (const propertyName in indexes) {
+      const value = indexes[propertyName];
```
diehuxx, line 52: It took me a second to read properly that this `value` isn't the same as the `value: id` a couple line down. I'm trying to think of another variable name than `value` for this.

What do you think of renaming `value` to `propertyValue` or `indexValue`?

src/store/index-level.ts
```diff
@@ -37,75 +33,124 @@ export class IndexLevel {
...
     }
 
+    const indexes = JSON.parse(serializedIndexes);
+
+    // delete all indexes associated with the data of the given ID
     const ops: LevelWrapperBatchOperation<string>[] = [ ];
-    for (const prefix of JSON.parse(prefixes)) {
-      ops.push({ type: 'del', key: this.join(prefix, id) });
+    for (const propertyName in indexes) {
+      const value = indexes[propertyName];
```
diehuxx, line 148: Same as https://github.com/TBD54566975/dwn-sdk-js/pull/383/files?diff=unified&w=1#r1207596520

src/store/index-level.ts
```diff
@@ -37,75 +33,124 @@ export class IndexLevel {
...
+    // Note: We have an array of Promises in order to support OR (anyOf) matches when given a list of accepted values for a property
+    const propertyNameToPromises: { [key: string]: Promise<string[]>[] } = {};
 
+    // Do a separate DB query for each property in `filter`
+    // We will find the union of these many individual queries later.
     for (const propertyName in filter) {
       const propertyFilter = filter[propertyName];
 
-      if (typeof propertyFilter === 'object' && propertyFilter !== null) {
+      if (propertyFilter === null) {
```
diehuxx, line 85: tbh I don't think we even need this check. Typescript doesn't complain if I delete it.

src/store/index-level.ts
```diff
@@ -37,75 +33,124 @@ export class IndexLevel {
...
-
-      const prefix = this.join(property, this.encodeValue(value));
-      ops.push({ type: 'put', key: this.join(prefix, id), value: id });
-      prefixes.push(prefix);
+  /**
+   * Adds indexes for a specific data/object/content.
+   * @param id ID of the data/object/content being indexed.
+   */
+  async put(
+    id: string,
```
diehuxx, line 41: another variable naming idea. `id` feels vague to me, even though you define it in the comment above this method. What do you think of changing `id` -> `dataId` throughout this file?

src/store/index-level.ts
diehuxx: Now that I think about it, I'm on the fence about these variable names too. I'll leave it up to you.

src/store/index-level.ts
```diff
@@ -114,56 +159,63 @@ export class IndexLevel {
...
-      matches.set(key, value);
+      matches.push(id);
     }
+
+    if ('lte' in range) {
+      // When `lte` is used, we must also query the exact match explicitly because the exact match will not be included in the iterator above.
+      // This is due to the extra data (CID) appended to the (property + value) key prefix, e.g.
+      // key = 'dateCreated\u0000"2023-05-25T11:22:33.000000Z"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
+      // the value would be considered greater than { lte: `dateCreated\u0000"2023-05-25T11:22:33.000000Z"` } in the iterator options,
+      // thus would not included in such iterator even though we'd like it to be.
```
diehuxx, line 213: Very helpful. Thanks.

diehuxx (dismissed):
Love this. Way easier to understand. Still a dense chunk of code, but at least now I can read through it.

I left a question and a couple suggestions for variable names. I'll leave it up to you if you want to make those changes; I'd also be comfortable with you merging this as-is.

src/store/index-level.ts
thehenrytsai: No, just didn't notice/bother enough. But you are right, it's redundant, went ahead and removed.

src/store/index-level.ts
thehenrytsai: `propertyValue` sounds good. Changing.

src/store/index-level.ts
thehenrytsai: Changed.

src/store/index-level.ts
thehenrytsai: Removing.

src/store/index-level.ts
thehenrytsai: Glad you suggest it. I wasn't sure `dataId` sounds "okay" or not, but sounds like you like it. Changing now.

diehuxx (approved):

PR #384, opened 2023-05-26
Add smarter return typing to Dwn.processMessage()

diehuxx:
Thanks to @mistermoe for this genius idea. I am but a medium through which his divine knowledge becomes a PR.

The aim is that typescript can determine what kind of message reply will be returned from `Dwn.processMessage` based on what type of DWN message it is processing.

### Changes
1. Bump Version
2. Update `Dwn.processMessage` to use generics, in order to provide better typing for the message and message reply.
  a. This involves adding a message name parameter to `processMessage`. This is a significant change so we're bumping version number.
3. Remove common `MessageReply` type. Return either `BaseMessage = { status: ...}` from a handler or return a DWN message specific reply type, e.g. `RecordsQueryReply`.

This involved updating a LOT of tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-96.98%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-93.92%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.31%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-96.98%25-brightgreen.svg?style=flat)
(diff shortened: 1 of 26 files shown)


thehenrytsai:
I have no strong opinion on this, so whatever makes the most sense for the users of SDK! My impression is that this would be very favorable, but adding @frankhinek for awareness and comment.

It does provide very nice typing support for the one method that rules them all (`processMessage()`), but it also requires an additional signal parameter indicating what message type it is, so the lunch is not "completely free".

mistermoe:
@thehenrytsai if `method`s were still like `RecordsWrite`, `RecordsRead` etc. we could have gotten away with a free lunch like so! (contrived example)

```typescript
type BaseMessage = {
    descriptor: {
        method: keyof MessageOptionsMap
    }
}

function processMessage<M extends keyof MessageOptionsMap>(message: BaseMessage, options: MessageOptions<M>): Ret<M> {
    const handler = handlers[message.descriptor.method];
    return handler(options as any) as Ret<M>
}
```

diehuxx:
Closing because we decided the tradeoff of better typechecking vs an extra parameter in `processMessage` isn't worth it, at least for now.

Issue #299, opened 2023-04-06
Support unsigned `RecordsQuery`


thehenrytsai:
Support unsigned `RecordsQuery`.

When a `RecordsQuery` is unsigned, return only published records.

thehenrytsai:
This is done.

PR #393, opened 2023-06-07
#299 - added support for unauthenticated RecordsQuery

thehenrytsai:


src/interfaces/records/handlers/records-query.ts
```diff
@@ -114,12 +129,17 @@ export class RecordsQueryHandler implements MethodHandler {
...
       isLatestBaseState : true,
       published         : false
     };
-    const unpublishedRecordsForQueryAuthor = await StorageController.query(this.messageStore, this.dataStore, tenant, filter);
+    let unpublishedRecordsForQueryAuthor = await StorageController.query(this.messageStore, this.dataStore, tenant, filter);
+
+    // removing records that have its author being the same as the query author,
+    // because those records will be included in records returned by `fetchUnpublishedRecordsByAuthor()` already
+    // TODO: #392 - use a NOT query directly in the filter as a performance improvement (https://github.com/TBD54566975/dwn-sdk-js/issues/392)
+    unpublishedRecordsForQueryAuthor = unpublishedRecordsForQueryAuthor.filter(message => Message.getAuthor(message) !== recordsQuery.author);
```
diehuxx, line 141: I have a slight preference for doing the dedupe in `fetchRecordsAsNonOwner` instead of here. By doing the dedupe here, we omit some of the records that conceptually fall into the category of "records where the query author is the `recipient`".

Unless there's a performance reason you're doing it here, I think it would be more conceptually clean to remove this `.filter()` and do `const records = [...publishedRecords, ...unpublishedRecordsByAuthor, ...unpublishedRecordsForQueryAuthor];` in `fetchRecordsAsNonOwner()` above.

tests/interfaces/records/handlers/records-query.spec.ts
```diff
@@ -449,11 +449,44 @@ describe('RecordsQueryHandler.handle()', () => {
...
+
+      const recordsWrite1Reply = await dwn.processMessage(alice.did, record1Data.message, record1Data.dataStream);
+      expect(recordsWrite1Reply.status.code).to.equal(202);
+      const recordsWrite2Reply = await dwn.processMessage(alice.did, record2Data.message, record2Data.dataStream);
+      expect(recordsWrite2Reply.status.code).to.equal(202);
+
+      // test correctness for anonymous query
+      const anonymousQueryMessageData = await TestDataGenerator.generateRecordsQuery({
+        filter: { dateCreated: { from: '2000-01-01T10:20:30.123456Z' } }
+      });
```
diehuxx, line 472: Do you mind adding an option to `generateRecordsQuery` that will make it not generate a persona? I prefer that to `delete`ing `authorization`.

```suggestion
      const anonymousQueryMessageData = await TestDataGenerator.generateRecordsQuery({
        filter: { dateCreated: { from: '2000-01-01T10:20:30.123456Z' } },
        anonymous: true
      });
```

thehenrytsai: Sure, added. Didn't think was worth it due to some minor cascading changes.

src/interfaces/records/handlers/records-query.ts
thehenrytsai: Agree the method name can now mislead even though I explicitly update documentation here as well as method summary to setup expectation. It is indeed for perf reason that I did this, which is why I went as far as creating another issue to improve the inefficiency further.

I went ahead and implemented the dedup logic though, since you are right that my approach is probably a premature optimization!!! I was originally not convinced that iterating the returned entries to compute their hash as map key is the "right" approach as it seems quite heavy in the name of de-dup when you could avoid it in the first place. But sneakily I used `recordId` as the key for now, it might not suffice when we return commits, but we can worry about it when we cross that bridge.

tests/interfaces/records/handlers/records-query.spec.ts
thehenrytsai: Sure, added. Didn't think was worth the additional code and cascading effects.

diehuxx (approved):
ty

PR #398, opened 2023-06-16
flattening directory structure

thehenrytsai:
Flattening directory structure in preparation for a new set of interface methods to be added.

`src/interfaces/records/handlers/records-write.ts` ->`src/handlers/records-write.ts`
`src/interfaces/records/messages/records-write.ts` ->`src/interfaces/records-write.ts`

let me know if you think it ought to be `src/methods/records-write.ts`, I didn't do `src/messages/records-write.ts` to avoid `src/messages/messages-get.ts`, but okay with it if people prefer it.


diehuxx (approved):
works for me

PR #402, opened 2023-06-20
Decompose and expose preprocessing checks

diehuxx:
## Change
At the request of @mistermoe, we're making the preprocessing checks public. Since, `preprocessingChecks` is a vague name to expose publicly, we're decomposing the check into two parts: tenant gate check and message structure validation.

## Context
Currently message processing has roughly the following steps
1. check tenant gate
2. validate message structure
5. store message
6. send 20X response, barring other issues with the DWN message.

But if the data is considered invalid at the application-level, we do not want to send a 20X. 20X is misleading, and leads to unnecessary complexities when trying to un-accept the message. It is useful to insert a step between 2 and 3 where the application validates data. (dwn-sdk-js cannot/should not do this validation as the data may be encrypted).

By exposing the tenant gate check and message structure validation, we give application developers the option to insert a data validation step before they store the message and respond.

### DIFF
--- a/src/core/message.ts
+++ b/src/core/message.ts
@@ -28,6 +28,17 @@ export enum DwnMethodName {
   Delete = 'Delete'
 }
 
+export enum DwnMessageType {
+  EventsGet = 'EventsGet',
+  MessagesGet = 'MessagesGet',
+  ProtocolsConfigure = 'ProtocolsConfigure',
+  ProtocolsQuery = 'ProtocolsQuery',
+  RecordsDelete = 'RecordsDelete',
+  RecordsQuery = 'RecordsQuery',
+  RecordsRead = 'RecordsRead',
+  RecordsWrite = 'RecordsWrite',
+};
+
 export abstract class Message<M extends BaseMessage> {
   readonly message: M;
   readonly authorizationPayload: any;
--- a/src/core/message.ts
+++ b/src/core/message.ts
@@ -28,17 +28,6 @@ export enum DwnMethodName {
   Delete = 'Delete'
 }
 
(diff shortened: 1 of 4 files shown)


src/core/message.ts
```diff
@@ -28,6 +28,17 @@ export enum DwnMethodName {
   Delete = 'Delete'
 }
 
+export enum DwnMessageType {
```
diehuxx, line 31: Adding this since I've noticed that the `web5` package has a notion of dwn message type. i've been wanting to add this for a while, and I intend to do more with it in a tangentially related follow up PR.

In this PR, I'm only adding the DWN messages that are fully supported and ready for use. If I'm missing any or if reviewers would prefer I add everything all at once, i can do that too.

thehenrytsai: Can you just briefly describe where you'll be using it more to demonstrate the need?

Not super concerned either way but right now I am mainly seeing that one parameter is passed instead of two, to which I am not convinced it is worthwhile because of the overhead (albeit minor) of context knowledge of new type and maintenance of it.

thehenrytsai (dismissed):

src/core/message.ts
diehuxx: I was planning to use it to organize message types, reply types, and handlers, sorta similar to what we did in `dwn-types` in #384. After messing around with it for half an hour, I've convinced myself that it's not as useful for that purpose as I thought 🙈.

thehenrytsai (approved):

thehenrytsai (approved):

mistermoe:
thanks @diehuxx !!

Issue #406, opened 2023-06-23
Apply consistency to `dateCreated` and `dateModified` across all DWN interfaces


thehenrytsai:
Currently, `dateCreated` means date "record" created in Record interface, but mean date "message" created in others. `dateModified` always means date "message" created, but is used only in Record and Protocol interface - basically it's kind of a mess... = ]

There a number of proposals to fix this inconsistency, first we have this:

1. Require all DWN message to have `messageTimestamp`
2. `dateCreated` is a method specific property, ie. in `RecordsWrite`.
3. Remove/rename all other variant of date related properties accordingly.

```ts
export type RecordsWriteDescriptor = {
  interface: 'Records';
  method: 'Write';
  messageTimestamp: string;
  dateCreated: string;
  datePublished?: string;
  ...
};

export type RecordsQueryDescriptor = {
  interface: 'Records';
  method: 'Query';
  messageTimestamp: string;
  filter: RecordsQueryFilter;
  dateSort?: DateSort;
};

export type RecordsReadDescriptor = {
  interface: 'Records';
  method: 'Read';
  messageTimestamp: string;
  recordId: string;
};

export type RecordsDeleteDescriptor = {
  interface: 'Records';
  method: 'Delete';
  messageTimestamp: string;
  recordId: string;
};

export type ProtocolsConfigureDescriptor = {
  interface : 'Protocols';
  method: 'Configure';
  messageTimestamp: string;
  dateCreated: string; // assuming we want/need it at some point
  definition: ProtocolDefinition;
};

export type ProtocolsQueryDescriptor = {
  interface : 'Protocols',
  method: 'Query';
  messageTimestamp: string;
  filter?: ProtocolsQueryFilter
};

export type PermissionsRequestDescriptor = {
  interface: 'Permissions';
  method: 'Request';
  messageTimestamp: string;
  ...
};

export type PermissionsGrantDescriptor = {
  interface: 'Permissions';
  method: 'Grant';
  messageTimestamp: string;
  ...
};

export type SnapshotsCreateDescriptor = {
  interface : 'Snapshots';
  method: 'Create';
  messageTimestamp: string;
  definitionCid: string;
};

```

It might be worth entertaining the idea of taking the `messageTimestamp` out of the `descriptor` container.

Keeping the above semantic meaning exactly the same, an alternate naming scheme was also proposed:
1. But this would introduce a bunch of special `dateXYZCreated` property that is arguably ugly and unnecessary.

```ts
[...]

PR #417, opened 2023-06-28
Create and process PermissionsRevoke

diehuxx:
Next step in implementing [TP26](https://github.com/TBD54566975/technical-proposals/pull/10/files).

This adds creation and processing of `PermissionsRevoke` message, but does NOT add grant-based authorization. That will be in the next PR, which will also boost test coverage for some scenarios in `PermissionsRevokeHandler` which are not currently reachable.


src/handlers/permissions-revoke.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    // query for previous revocations of this grant
+    const query = {
+      interface : DwnInterfaceName.Permissions,
+      method    : DwnMethodName.Revoke,
+      permissionsGrantId,
+    };
+    const existingRevokesForGrant: PermissionsRevokeMessage[] = await this.messageStore.query(tenant, query) as PermissionsRevokeMessage[];
+
+    // Conflict 409 if the grant already has an older revoke
```
diehuxx, line 72: I imagine this will be the one of the most surprising/nuanced pieces of functionality.

Consider the scenario where Alice sends a `PermissionsRevoke` to her local DWN, then a moment later sends a `PermissionsRevoke` to her remote DWN for the same `PermissionsGrant`. Upon DWN sync, we should use the earlier `PermissionsRevoke` and delete/bounce the later `PermissionsRevoke`.

src/handlers/permissions-revoke.ts
```diff
@@ -0,0 +1,115 @@
...
+    // Store incoming PermissionsRevoke
+    const indexes: { [key: string]: string } = {
+      interface          : DwnInterfaceName.Permissions,
+      method             : DwnMethodName.Revoke,
+      permissionsGrantId : message.descriptor.permissionsGrantId,
+    };
+    await this.messageStore.put(tenant, message, indexes);
+    await this.eventLog.append(tenant, await Message.getCid(message));
+
+    // Delete existing revokes which are all older than the incoming message
```
diehuxx, line 89: Related to https://github.com/TBD54566975/dwn-sdk-js/pull/417/files?diff=unified&w=0#r1245751682

src/handlers/permissions-revoke.ts
```diff
@@ -0,0 +1,115 @@
...
+    const removedRevokeCids: string[] = [];
+    for (const existingRevoke of existingRevokesForGrant) {
+      const existingRevokeCid = await Message.getCid(existingRevoke);
+      await this.messageStore.delete(tenant, existingRevokeCid);
+      removedRevokeCids.push(existingRevokeCid);
+    }
+    await this.eventLog.deleteEventsByCid(tenant, removedRevokeCids);
+
+    // Delete grant-authorized messages with timestamp after revocation
```
diehuxx, line 98: Related to https://github.com/TBD54566975/dwn-sdk-js/pull/417/files?diff=unified&w=0#r1245751682

To make the scenario a bit more complex, consider the following sequence of events:
1. Alice issues a `PermissionsGrant` to Bob, and all of her DWNs have processed this grant.
2. Alice sends a `PermissionsRevoke` to her local DWN
3. Bob sends a message to Alice's remote DWN using his `PermissionsGrant`
4. Alice's remote DWN syncs with her local DWN, pulling the `PermissionsRevoke`.

Since the revocation was created before Bob's message, Bob's message should not actually have been authorized. So, Bob's message is deleted from Alice's DWN.

src/handlers/permissions-revoke.ts
thehenrytsai, line 89: sanity:
```suggestion
    // Delete existing revokes which are all newer than the incoming message
```

src/handlers/permissions-revoke.ts
```diff
@@ -0,0 +1,115 @@
...
+    const query = {
+      interface : DwnInterfaceName.Permissions,
+      method    : DwnMethodName.Revoke,
+      permissionsGrantId,
+    };
+    const existingRevokesForGrant: PermissionsRevokeMessage[] = await this.messageStore.query(tenant, query) as PermissionsRevokeMessage[];
+
+    // Conflict 409 if the grant already has an older revoke
+    const oldestExistingRevoke = PermissionsRevoke.getOldestRevoke(existingRevokesForGrant);
+    if (oldestExistingRevoke !== undefined && oldestExistingRevoke.descriptor.dateCreated < message.descriptor.dateCreated) {
```
thehenrytsai, line 74: Sanity: super edge, should we probably use lexicographical compare so there is never an equal? Else we can get into odd situation when timestamps are equal where different version of the same revoke maybe stored by different DWN, which may have no security implementation, but the byte-for-byte state of the DWN would not be deterministic.

src/handlers/permissions-revoke.ts
```diff
@@ -0,0 +1,115 @@
...
+      const existingRevokeCid = await Message.getCid(existingRevoke);
+      await this.messageStore.delete(tenant, existingRevokeCid);
+      removedRevokeCids.push(existingRevokeCid);
+    }
+    await this.eventLog.deleteEventsByCid(tenant, removedRevokeCids);
+
+    // Delete grant-authorized messages with timestamp after revocation
+    const grantAuthdMessagesQuery = {
+      permissionsGrantId,
+      dateCreated: { gt: message.descriptor.dateCreated },
```
thehenrytsai, line 101: `gte`?

src/interfaces/permissions-revoke.ts
```diff
@@ -0,0 +1,61 @@
...
+  public async authorize(permissionsGrantMessage: PermissionsGrantMessage): Promise<void> {
+    if (this.author !== permissionsGrantMessage.descriptor.grantedFor) {
+      // Until delegation is implemented, only the DWN owner may grant or revoke access to their DWN
+      throw new DwnError(DwnErrorCode.PermissionsRevokeUnauthorizedRevoke, 'Only the DWN owner may revoke a grant');
+    }
+  }
+
+  public static getOldestRevoke(permissionsRevokeMessages: PermissionsRevokeMessage[]): PermissionsRevokeMessage | undefined {
+    // Sort revoke messages by `dateCreated`
+    const sortedRevokes = permissionsRevokeMessages.sort((revoke1, revoke2) => {
```
thehenrytsai, line 47: Nit: O(nLog(n)) can be avoid with an O(n) here. Basically like how `getNewestMessage()` is done.

I know #406 is probably putting a lot uncertainty here, so I'd be fine if you just reference #406 with a TODO comment to kick the can, 'coz we could reuse much of the existing code.

src/handlers/permissions-revoke.ts
thehenrytsai: Thanks! Super helpful!

thehenrytsai (commented):
Looks good overall! A few small comments.

src/interfaces/permissions-revoke.ts
diehuxx: Good call. Fixed.

src/handlers/permissions-revoke.ts
diehuxx: fixed

src/handlers/permissions-revoke.ts
diehuxx: Fixed.

src/handlers/permissions-revoke.ts
alanhkarp: Are you implying that processing Bob's request at the remote DWN is delayed until it syncs with Alice's local DWN?  What if that sync is delayed?

The important thing to note is that there is an unavoidable race between Alice's revoke and Bob's message because their actions are concurrent.  After all, Bob could have sent his message just a little bit before Alice sent her revoke or just a little bit after.  That means it's important to carefully define what "before" means in this context.

src/interfaces/permissions-revoke.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return currentOldestMessage;
+  }
+
+  /**
+   * Checks if first message is older than second message.
+   * @returns `true` if `a` is older than `b`; `false` otherwise
+   */
+  public static async isCreatedOlder(a: PermissionsRevokeMessage, b: PermissionsRevokeMessage): Promise<boolean> {
+    // TODO: #406 - Deduplicate code with Message.isOlder (https://github.com/TBD54566975/dwn-sdk-js/issues/406)
```
thehenrytsai, line 65: Sanity: time to remove these methods?

(Also, I have a horrible copy & paste error in the original code, it should be `aIsOlder`, thank you if you can opportunistic rename!)

src/handlers/permissions-revoke.ts
```diff
@@ -70,8 +70,8 @@ export class PermissionsRevokeHandler implements MethodHandler {
     const existingRevokesForGrant: PermissionsRevokeMessage[] = await this.messageStore.query(tenant, query) as PermissionsRevokeMessage[];
 
     // Conflict 409 if the grant already has an older revoke
-    const oldestExistingRevoke = await PermissionsRevoke.getOldestRevoke(existingRevokesForGrant);
-    if (oldestExistingRevoke !== undefined && !(await PermissionsRevoke.isCreatedOlder(oldestExistingRevoke, message))) {
+    const oldestExistingRevoke = await Message.getOldestMessage(existingRevokesForGrant);
+    if (oldestExistingRevoke !== undefined && !(await Message.isNewer(oldestExistingRevoke, message))) {
```
thehenrytsai, line 74: Sanity: is this supposed to be `isOlder()`? If so, consider adding a test since this was not caught by existing tests.

diehuxx: It is supposed to be `isNewer`. We do have a test which failed when I was deduping the code. I made two mistakes that canceled each other out to produce correct behavior. First, I used `isCreatedOlder` when I should have used `isCreatedNewer`. Second, I wrote a bug in `isCreatedOlder` so it was returning whether the first message was newer.

The number of negations is making my head spin. Let me update it one more time so it's less confusing.

thehenrytsai (approved):
🎆

src/handlers/permissions-revoke.ts
diehuxx: No, Bob's request is not delayed. Alice's remote DWN processes Bob's request. Some time later Alice's DWNs sync, and her remote DWN receives the `PermissionsRevoke`, which has an earlier `messageTimestamp` than Bob's message. Even though Alice's remote processed Bob's message first, the `messageTimestamp` takes precedence, and Bob's message is deleted.

alanhkarp: If Bob's message is processed, then any action it specifies is processed as if no revoke action had taken place.  If you later delete Bob's message, does that undo any action taken when processing Bob's message?

PR #418, opened 2023-06-29
Updated BaseMessage in preparation for external Store interface implementations

thehenrytsai:
1. Removed Record specific properties from `BaseMessage`
1. Renamed `BaseMessage` to `GenericMessage` - Makes more sense from the perspective of the store implementor, as they don't care about inheritance/extension
1. Opportunistic rename of `Reply" types

### DIFF
--- a/src/core/message-reply.ts
+++ b/src/core/message-reply.ts
@@ -20,7 +20,7 @@ export function messageReplyFromError(e: unknown, code: number): BaseMessageRepl
 /**
  * Catch-all message reply type. It is recommended to use BaseMessageReply or a message-specific reply type whereever possible.
  */
-export type GenericMessageReply = BaseMessageReply & {
+export type UnionMessageReply = BaseMessageReply & {
   /**
    * Resulting message entries or events returned from the invocation of the corresponding message.
    * e.g. the resulting messages from a RecordsQuery
--- a/src/core/message-reply.ts
+++ b/src/core/message-reply.ts
@@ -6,11 +6,11 @@ type Status = {
   detail: string
 };
 
-export type BaseMessageReply = {
+export type GenericMessageReply = {
   status: Status;
 };
 
-export function messageReplyFromError(e: unknown, code: number): BaseMessageReply {
+export function messageReplyFromError(e: unknown, code: number): GenericMessageReply {
 
   const detail = e instanceof Error ? e.message : 'Error';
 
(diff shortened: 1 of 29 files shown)


src/core/message-reply.ts
```diff
@@ -6,21 +6,21 @@ type Status = {
...
 
   return { status: { code, detail } };
 }
 
 /**
- * Catch-all message reply type. It is recommended to use BaseMessageReply or a message-specific reply type whereever possible.
+ * Catch-all message reply type. It is recommended to use GenericMessageReply or a message-specific reply type whereever possible.
  */
-export type GenericMessageReply = BaseMessageReply & {
+export type UnionMessageReply = GenericMessageReply & {
```
diehuxx, line 23: I wish I had removed `UnionMessageReply` altogether, though I'm not sure if it's possible while keeping `processMessage` in its current form. This type feels messy. If we've gotta keep it, I like this naming.

diehuxx: Also, in `index.ts`, we're doing `export { UnionMessageReply as MessageReply }`. We should update to just `export { UnionMessageReply }`

src/store/storage-controller.ts
```diff
@@ -42,16 +42,20 @@ export class StorageController {
...
-  public static async delete(
+  /**
+   * Deletes a Record message.
+   */
+  public static async deleteRecordMessage(
     messageStore: MessageStore,
     dataStore: DataStore,
     tenant: string,
-    message: BaseMessage
+    message: GenericMessage
```
diehuxx, line 52: Taking `GenericMessage` in a method called `deleteRecordMessage` feels a bit odd. I'd prefer if we either 1. did `message: RecordsWriteMessage` to make this method stricter, or 2. left the method name as `delete`.

thehenrytsai: Agree it is a bit odd, I considered changing the type, but it will involve both `RecordsWriteMessage` and `RecordsDelete` message and in the future `RecordsCommit`, a union type here is super ugly and will also trigger cascading changes to the caller who only has `TimestampedMessage`. I also considered create a new type called `RecordsMessage`, but that'd not include `RecordsQueryMessage` and `RecordsReadMessage` so the name won't be accurate.

Because there is no non-record handler that calls this method (since they don't have data to delete), I renamed it to reflect that it is really meant for just record handlers. I was also considering to rename the class `RecordStorageController` as an alternative but that sounded unwieldy, so I passed.

I took your suggestion to rename back to delete, since it is the existing name so there is no regression. Let me know if you like any approaches above or have new ones!

diehuxx: Agreed that all of those other approaches are not quite right. Keeping the name as `delete` is my favorite for now. Depending on how #406 shakes out, my fingers are crossed that we can change/remove `TimestampedMessage` and hopefully this `delete()` method will become easier to name and type.

diehuxx (approved):

PR #419, opened 2023-06-30
Standardize message timestamp naming

diehuxx:
Taking approach 1 from #406.
> Require all DWN message to have messageTimestamp
dateCreated is a method specific property, ie. in RecordsWrite.
Remove/rename all other variant of date related properties accordingly.

* In `RecordsWrite`, `RecordsDelete`, and `ProtocolsConfigure`: `dateModified` -> `messageTimestamp`.
* `RecordsWrite` retains the `dateCreated` field, which refers to the time that the _record_ was created.
* In `RecordsRead`: `date` -> `messageTimestamp`
* `EventsGet` and `MessagesGet` do not have timestamps of any kind.
* In all other DWN messages, `dateCreated` -> `messageTimestamp`

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # Decentralized Web Node (DWN) SDK
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-97.31%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-94.26%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.68%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-97.31%25-brightgreen.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-97.31%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-94.01%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-93.68%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-97.31%25-brightgreen.svg?style=flat)
 
 ## Introduction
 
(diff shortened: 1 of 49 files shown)


thehenrytsai:
@diehuxx, everything looks great, though I think we should also add timestamp for `MessageGet` and `EventGet` to prevent replay. What do you think? Obviously we'd need to add more checks, but that can be done in a future PR.

diehuxx:
> add timestamp for `MessageGet` and `EventGet` to prevent replay

I don't see any reason why not, but not sure if I understand the reason. What do you mean by replay? Also, does the timestamp have any significance if we aren't storing the `MessagesGet` or `EventsGet` messages and we aren't using the timestamp for filtering purposes?

thehenrytsai:
> I don't see any reason why not, but not sure if I understand the reason. What do you mean by replay? Also, does the timestamp have any significance if we aren't storing the `MessagesGet` or `EventsGet` messages and we aren't using the timestamp for filtering purposes?

Imagine if a bad actor got hold of a legit (signed) `MessageGet` message. They can just send this message to a DWN to get the message back any time in the future (assuming no key rotation).

> we aren't using the timestamp for filtering purposes?

Yeah... that's the "we'd need to add more checks, but that can be done in a future PR" part I was referring to, as in, check the timestamp against the current server time, if it's outside of the acceptable "server time drift" range, ignore. This applies to all fetch-type messages that we have including `RecordsQuery` and `RecordsRead`.

diehuxx:
> Imagine if a bad actor got hold of a legit (signed) `MessageGet` message. They can just send this message to a DWN to get the message back any time in the future (assuming no key rotation).

Ahhh okay I see. Thanks for the explanation :). Just pushed a commit adding timestamp to `EventsGet` and `MessagesGet`.

thehenrytsai (dismissed):
🙏

thehenrytsai (approved):

PR #424, opened 2023-07-05
Add stronger typing and JSON schemas for authorization payloads

diehuxx:


build/compile-validators.js
```diff
@@ -60,7 +62,9 @@ const schemas = {
   ProtocolsQuery,
   RecordsRead,
   PublicJwk,
-  SnapshotsCreate
+  SnapshotsCreate,
+  BaseDecodedAuthorizationPayload,
+  RecordsWriteDecodedAuthorizationPayload
```
thehenrytsai, line 67: Consider leaving `Decoded` out: we can think of payload being either "as is" or "encoded".

```suggestion
  BaseAuthorizationPayload,
  RecordsWriteAuthorizationPayload
```

thehenrytsai (commented):
A couple of comments.

build/compile-validators.js
diehuxx: Updated

src/core/auth.ts
```diff
@@ -47,33 +43,28 @@ export async function validateAuthorizationIntegrity(
...
+  const dwnMethod = message.descriptor.method;
+
+  let schemaLookupKey: string;
+  if (dwnInterface === DwnInterfaceName.Records && dwnMethod === DwnMethodName.Write) {
+    schemaLookupKey = 'RecordsWriteDecodedAuthorizationPayload';
+  } else {
+    schemaLookupKey = 'BaseDecodedAuthorizationPayload';
+  }
+  validateJsonSchema(schemaLookupKey, payloadJson);
+
```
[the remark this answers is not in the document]
diehuxx: Update from a conversation we had offline:

* JSON Schema are worth keeping because they are documentation-like and they validate the shape of different auth payloads
* It doesn't make sense for `validateAuthorizationIntegrity` to check that all properties other than `descriptorCid` are CIDs. We'll leave that responsibility up to each DWN interface (currently, only `RecordsWrite.validateIntegrity`).
* We don't want this function to be DWN interface dependence, so as a middle-ground, we are passing the JSON schema key as a parameter.

thehenrytsai (approved):

PR #426, opened 2023-07-06
Exported test suite so that store-dependent tests can be run from external repositories

thehenrytsai:
Quite heavy-handed on both approach and implementation to get this support going after failing various experimentations....... so open to more elegant alternatives and/or improvements!

### DIFF
--- a/tests/handlers/records-write.spec.ts
+++ b/tests/handlers/records-write.spec.ts
@@ -41,1480 +41,1483 @@ import { Encryption, EncryptionAlgorithm } from '../../src/utils/encryption.js';
 
 chai.use(chaiAsPromised);
 
-describe('RecordsWriteHandler.handle()', () => {
-  let didResolver: DidResolver;
-  let messageStore: MessageStore;
-  let dataStore: DataStore;
-  let eventLog: EventLog;
-  let dwn: Dwn;
+export function testRecordsWriteHandler(): void {
+  describe('RecordsWriteHandler.handle()', async () => {
+    let didResolver: DidResolver;
+    let messageStore: MessageStore;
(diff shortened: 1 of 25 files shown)


tests/handlers/records-write.spec.ts
```diff
@@ -34,1487 +34,1490 @@ import { RecordsWrite } from '../../src/interfaces/records-write.js';
...
+export function testRecordsWriteHandler(): void {
+  describe('RecordsWriteHandler.handle()', async () => {
+    let didResolver: DidResolver;
+    let messageStore: MessageStore;
+    let dataStore: DataStore;
+    let eventLog: EventLog;
+    let dwn: Dwn;
 
-  describe('functional tests', () => {
+    describe.only('functional tests', () => {
```
diehuxx, line 52:
```suggestion
    describe('functional tests', () => {
```

diehuxx (commented):
Added one small comment. Here are my broader thoughts.

This is definitely improvement for people developing a custom store, but there's a slight QoL drawback for you and me. We can't do `test.only` or `test.skip`, which will make debugging annoying for us. We can pretty easily workaround by commenting out a few things while we debug. Though I wonder if custom store developers want the ability to run individual tests too. I can't find a graceful way to do that right now, so for the scope of this PR, LGTM.

tests/handlers/records-write.spec.ts
thehenrytsai: Thanks for catching this! Didn't realize `.only` stopped working as I was using it to test my changes quickly. Fixed.

diehuxx (approved):
niiiiice. lgtm

PR #428, opened 2023-07-08
Moved dependencies needed by exported test suite as SDK dependencies

thehenrytsai:


frankhinek:
On the surface, this seems like an odd scenario to be requiring a mocking/stubbing library, `sinon`, as a runtime dependency for the DWN SDK, which will result in every project that directly or transitively depends on the project to have a library only intended for tests.

My instinct is to think this is a sign we should come up with a different way to export the test suite.

thehenrytsai:
@frankhinek, definitely open to suggestions to viable alternatives. As the code currently stands, we either need to merge this, or roll back the PR that introduced test suite export.

thehenrytsai:
Including test dependencies is undesirable.

PR #432, opened 2023-07-10
Authorize ProtocolsConfigure with PermissionsGrant

diehuxx:
Another PR for [TP26](https://github.com/TBD54566975/technical-proposals/pull/10).

Finally, we are actually able to use `PermissionsGrants` for authorization. For the sake of keeping the diff at a reasonable size, this PR only updates `ProtocolsConfigure` to use `PermissionsGrants`. Why `ProtocolsConfigure`? Because it is much simpler than any of the `Records` DWN methods. There's no particular reason I did not choose `ProtocolsQuery`, which will be updated in a fast follow up PR.

The new `GrantAuthorization` class is meant for grant validation that is common to all relevant DWN message types. As we build out DWN method-specific `scope` options, validation will be implemented in files directly related to that DWN method, e.g. `interface/protocols-configure.ts` or `handlers/protocols-configure.ts`.


src/interfaces/protocols-configure.ts
```diff
@@ -71,4 +74,18 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
       types    : typesCopy,
     };
   }
+
+  public async authorize(tenant: string, messageStore: MessageStore): Promise<void> {
+    // if author is the same as the target tenant, we can directly grant access
+    if (this.author === tenant) {
+      return;
+    } else if (this.author !== undefined && this.authorizationPayload?.permissionsGrantId) {
+      await GrantAuthorization.authorizeGenericMessage(tenant, this, this.author, messageStore);
```
thehenrytsai, line 83: Consider moving checks beyond `this.author === tenant` within `authorizeGenericMessage()`?
```suggestion
      await GrantAuthorization.authorizeGenericMessage(tenant, this, messageStore);
```

src/core/grant-authorization.ts
```diff
@@ -0,0 +1,150 @@
...
+    messageStore: MessageStore,
+  ): Promise<void> {
+    // Check that incomingMessage is within the grant's timeframe
+    if (incomingMessage.descriptor.messageTimestamp < permissionsGrantMessage.descriptor.messageTimestamp) {
+      // grant is not yet active
+      throw new DwnError(
+        DwnErrorCode.GrantAuthorizationGrantNotYetActive,
+        `The message has a timestamp before the associated PermissionsGrant becomes active`,
+      );
+    } else if (permissionsGrantMessage.descriptor.dateExpires !== undefined) {
```
thehenrytsai, line 102: Consult @frankhinek and @csuwildcat if `dateExpires` should be optional.

thehenrytsai (commented):
Looks legit! Just a couple of small comments for consideration.

src/interfaces/protocols-configure.ts
diehuxx: Do you mind if I leave it as-is for now, maybe changing it in a later PR? I have an intuition that it'll be cleaner this way since other message types have different handling for anonymous messages. If it turns out to be messier as I implement grant-auth for other message types I'll move these checks into `authorizeGenericMessage()`.

I'm sorta following the pattern of `ProtocolAuthorization` where we check if protocol URI is present before even entering `ProtocolAuthorization.authorize()`.

src/core/grant-authorization.ts
diehuxx: Discussed at office hours. None of us seem to to feel super strongly, but we leaned towards requiring `dateExpires`. It's easier to require it for now and relax the requirement later than vice versa.

src/interfaces/protocols-configure.ts
thehenrytsai: OK with it.

thehenrytsai (dismissed):

thehenrytsai (approved):

PR #437, opened 2023-07-12
Updated flat-space encryption scheme

thehenrytsai:
Updated flat-space encryption scheme according to TP27, which is the simplest part of the TP27, so I am tackling it first to test the waters.

@csuwildcat, can you upload TP27 to https://github.com/TBD54566975/technical-proposals/pulls, thanks!


src/utils/records.ts
```diff
@@ -88,17 +88,21 @@ export class Records {
...
-      descriptor.dataFormat,
-      recordId
-    ];
-
-    return fullDerivationPath;
+    if (descriptor.schema !== undefined) {
+      return [
+        KeyDerivationScheme.DataFormats,
+        descriptor.schema, // this is as spec-ed on TP27, the intent is to support sharing the key for just a specific data type under a schema
+        descriptor.dataFormat
```
diehuxx, line 98: @thehenrytsai I remember we talked about the fact that the TP27 seems to spec this, but I'm still not convinced that this is or isn't the right thing to do. What was the reason/use case why we don't want to share a key for all records of a given `dataFormat`, including records with different `schemas`?

thehenrytsai: I extrapolate the reason being the same as why spec does not support sharing a key for all records of a given schema, including protocol-space records.

diehuxx (approved):

src/utils/records.ts
diehuxx: \I definitely agree with sharply dividing access of non-protocol schema from protocol records. So, tentatively, following that pattern for dataFormat works for me.

diehuxx (approved):

PR #442, opened 2023-07-17
Add exports for DataStore, MessageStore, and EventLog types

amika-sq:
Currently, DataStore, MessageStore, and EventLog aren't exporting all of the types that are used within their interfaces. 

This required me to re-define the types alongside my custom implementation. It'd be awesome if we just exported these from the sdk itself!

I ran into this when implementing custom SQL-backed stored here: https://github.com/TBD54566975/dwn-sql-store

I also added an export for the utility function `executeUnlessAborted`, as our TestSuite has expectations around throwing when an abort signal has been triggered. This helper function can be used in the custom implementations for easy use!


amika-sq:
Examples where I had to re-define these types in my custom store implementation:
* [EventLog](https://github.com/TBD54566975/dwn-sql-store/blob/main/src/event-log-sql.ts#L18)
* [DataStore](https://github.com/TBD54566975/dwn-sql-store/blob/main/src/data-store-sql.ts#L7)
* [MessageStore](https://github.com/TBD54566975/dwn-sql-store/blob/main/src/message-store-sql.ts#L9)

mistermoe (approved):

thehenrytsai (approved):

PR #443, opened 2023-07-17
Added support for multi-implementation test suite runs in one external repository

thehenrytsai:
@amika-sq reported the need to run the exported test suites multiple times, however this is not currently supported because
1. Any number of times the custom store overrides will all take place immediately before test suite is run regardless how many time it is supposed to run, resulting in only the last override taking effect for all test suite runs.
2. A test is modifying a protocol definition used across test suite runs, causing 3 tests dependent on the protocol definition to always fail  in subsequent test suite runs.


amika-sq (approved):
Thank you so much for digging into these issues! Really appreciate it!

diehuxx (approved):

PR #446, opened 2023-07-19
Added support for unauthenticated queries for published ProtocolsConfigures

thehenrytsai:
1. Added concept of published ProtocolsConfigures
1. Added support for unauthenticated queries for published ProtocolsConfigures
1. Retain `authorization` property when returning results for `ProtocolsQuery`
1. Refactored `RecordsQuery` handler for cleaner auth logic (should be helpful for adding permission auth)

### DIFF
--- a/src/handlers/records-query.ts
+++ b/src/handlers/records-query.ts
@@ -26,41 +26,53 @@ export class RecordsQueryHandler implements MethodHandler {
       return messageReplyFromError(e, 400);
     }
 
+    // if this is an anonymous query, query only published records
+    if (recordsQuery.author === undefined) {
+      let recordsWrites = await this.fetchPublishedRecords(tenant, recordsQuery);
+      recordsWrites = await sortRecords(recordsWrites, recordsQuery.message.descriptor.dateSort);
+
+      const entries = RecordsQueryHandler.removeAuthorization(recordsWrites);
+      return {
+        status: { code: 200, detail: 'OK' },
+        entries
+      };
+    }
+
     // authentication
     try {
-      if (recordsQuery.author !== undefined) {
-        await authenticate(message.authorization!, this.didResolver);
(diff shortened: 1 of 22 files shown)


src/handlers/records-query.ts
```diff
@@ -26,43 +26,55 @@ export class RecordsQueryHandler implements MethodHandler {
       return messageReplyFromError(e, 400);
     }
 
+    // if this is an anonymous query, query only published records
```
diehuxx, line 29: noice 👍

src/types/message-store.ts
```diff
@@ -22,7 +22,7 @@ export interface MessageStore {
   put(
     tenant: string,
     message: GenericMessage,
-    indexes: { [key: string]: string },
+    indexes: { [key: string]: string | boolean },
```
diehuxx, line 25: good call

diehuxx (dismissed):

src/interfaces/protocols-query.ts
```diff
@@ -24,7 +24,9 @@ export class ProtocolsQuery extends Message<ProtocolsQueryMessage> {
 
 
   public static async parse(message: ProtocolsQueryMessage): Promise<ProtocolsQuery> {
```
diehuxx, line 26: A couple lines above here, I added `readonly author!: string;` since `ProtocolsQuery` had previously required auth. We should remove that line now.

src/handlers/records-query.ts
thehenrytsai: 🙏

src/interfaces/protocols-query.ts
thehenrytsai: Nice catch! Removd.

diehuxx (approved):

PR #447, opened 2023-07-20
Lirancohen/records commit

LiranCohen:
Starting point for RecordsCommit interface.


diehuxx (commented):
Hi @LiranCohen! Thanks for the PR. Before I dive into this, I think we need a technical proposal to make sure we're all on the same page about what this feature should look like. We usually post technical proposals as PRs in the [technical-proposals](https://github.com/tbD54566975/technical-proposals) repo.

Skimming through this PR, some questions come to mind that I would want to answer before jumping into implementation. E.g.
- Why is it a new message type? Why not just an optional immutable property additional fields on RecordsWrite?
- Why is `parentid` required?
- What enforces that the commit strategy is actually used? Is it like `schema` where we just list the schema or protocolUri but dwn-sdk-js stores and indexes on it but doesn’t check it?
- Why isn’t this at the app layer? Why do we need commits?
- Are all `RecordsCommits` kept? Contrast with RecordsWrites where only the first+last are kept
- If I `RecordsQuery` or `RecordsRead`, can I get back `RecordsCommits` entries now or do I still just get `RecordsWrites`?
- As far as protocol definitions are concerned, does `RecordsCommit` fall into `write` or is it a separate action that requires a separate `$action` rule in the protocol definition?
- If an initial `RecordsWrite` has `commitStrategy` , can I only update the record with a commit? Or can I send subsequent writes to that `recordId`

LiranCohen:
Closing this for now while a more formal TP is being put together.

PR #448, opened 2023-07-20
Authorize RecordsRead of schema

diehuxx:
This sets us up for grant-authorized `RecordsRead`, `RecordsWrite`, and `RecordsDelete`, which all have similar grant scopes. This PR only authorizes a list of specific `recordIds` for `RecordsRead`. There are codepaths for `RecordsWrite` and `RecordsDelete` which as of this PR are unused, but will be used in the next PR.


json-schemas/interface-methods/permissions-grant.json
```diff
@@ -63,12 +63,28 @@
           "type": "string"
         },
         "scope": {
-          "$ref": "https://identity.foundation/dwn/json-schemas/permissions/defs.json#/definitions/scope"
+          "oneOf": [
+            {
+              "$ref": "https://identity.foundation/dwn/json-schemas/permissions/scopes.json#/definitions/protocols-configure-scope"
```
thehenrytsai, line 68: Sanity: I think this is no long needed on our earlier discussion? If you agree, please remove.

src/core/grant-authorization.ts
```diff
@@ -15,10 +15,16 @@ export class GrantAuthorization {
   public static async authorizeGenericMessage(
     tenant: string,
     incomingMessage: Message<GenericMessage>,
-    author: string,
     messageStore: MessageStore,
-  ): Promise<void> {
+  ): Promise<PermissionsGrantMessage> {
```
thehenrytsai, line 19: nit: consider adding `@return` in documentation above.

json-schemas/permissions/scopes.json
```diff
@@ -0,0 +1,92 @@
...
+        "method"
+      ],
+      "properties": {
+        "interface": {
+          "const": "Records"
+        },
+        "method": {
+          "const": "Delete"
+        },
+        "recordIds": {
```
thehenrytsai, line 41: The idea of granting someone the ability to delete Record X, Y, and Z seems strange to me.

src/core/grant-authorization.ts
```diff
@@ -15,10 +15,16 @@ export class GrantAuthorization {
...
     const permissionsGrantId: string = incomingMessage.authorizationPayload!.permissionsGrantId!;
+    if (incomingMessage.author === undefined) {
```
thehenrytsai, line 21: Sanity: I "think" this code path should not be reachable externally. Was there a particular code path that prompted this check? No push back if this is just a  defensive programming.

thehenrytsai, line 21: I am supportive for defensive check here as a fall back check. But what do you think of the pattern that I introduced where in general auth methods aren't called for unauthenticated calls? I find the code much easier to reason about. If you agree I can make a similar change that I did with `ProtocolsQuery` and `RecordsQuery` to `RecordsRead` in a future PR (I just forgot its existence at the time).

src/core/records-grant-authorization.ts
```diff
@@ -0,0 +1,47 @@
...
+    messageStore: MessageStore,
+  ): Promise<void> {
+
+    // authorize generic message
+    const permissionsGrantMessage = await GrantAuthorization.authorizeGenericMessage(tenant, incomingMessage, messageStore);
+
+    const grantScope = permissionsGrantMessage.descriptor.scope as RecordsPermissionScope;
+
+    // if recordIds, authorizeRecordIds
+    if (grantScope.recordIds !== undefined) {
```
thehenrytsai, line 27: It definitely seems useful to be able to grant access to multiple records, though we should get together tomorrow with @csuwildcat on the expectation around if a grant should align scoping to encryption key scope, which is one of: flat-space-schema, flat-space-schema-data-format, flat-space-data-format, protocol-context, protocol-path.

diehuxx: I decided to repurpose this PR to support flat-space-schema instead of a list of record IDs. We can circle back to record IDs in a couple weeks when we have time to sync on it again.

json-schemas/permissions/scopes.json
diehuxx: https://github.com/TBD54566975/dwn-sdk-js/pull/448#discussion_r1282442605

src/core/grant-authorization.ts
diehuxx: Updated. I structured it slightly different than in `RecordsQuery` and `ProtocolsQuery`, but we can make it exactly the same in a future PR if you think that's better.

PR #454, opened 2023-07-24
Bundling strategy

mistermoe:


build/create-cjs-bundle.cjs
```diff
@@ -0,0 +1,48 @@
...
+const baseConfig = {
+  platform : 'node',
+  format   : 'cjs',
+  bundle   : true,
+  external : excludeList,
+};
+
+const indexConfig = {
+  ...baseConfig,
+  entryPoints : ['./dist/esm/src/index.js'],
```
thehenrytsai, line 35: Sanity: so we are now building `cjs` from the `esm` module built correct?

package.json
```diff
@@ -42,35 +41,29 @@
...
-    "./dist/cjs/src/index.js": "./dist/bundles/browser.js"
-  },
   "@comment react-native": [
     "directive for metro (react native build tool)."
   ],
-  "react-native": {
-    "./dist/esm/src/index.js": "./dist/esm/src/index.js",
-    "./dist/cjs/src/index.js": "./dist/esm/src/index.js"
-  },
+  "react-native": "./dist/cjs/index.js",
```
thehenrytsai, line 66: Sanity: I thought the react-native folks had more success depending on the `esm`? Is `cjs` good for them to use again?

thehenrytsai (approved):
Thank you thank you thank you. 
🙏 🙏 🙏 
🎖️ 🏅 🥇

build/create-cjs-bundle.cjs
mistermoe: that's correct! doesn't work any other way 😞

package.json
mistermoe: for now! it seems to work better since the bundle now injects all of the third party dependencies that don't include cjs distributions or `react-native` import paths which prevents folks from having to remap anything themselves. previously they'd have to do [this](https://github.com/TBD54566975/web5-react-native-metro-config). not 100% sure this will end up holding up though! if it doesn't we'll switch back to esm

build/create-cjs-bundle.cjs
thehenrytsai: 🙏

Issue #453, opened 2023-07-24
(optimization) reduce # of underlying db queries needed to fulfill a `RecordsQuery`


mistermoe:
The storage architecture of DWNs separates record metadata from record data by storing each in entirely separate data stores. the design rationale behind this is motivated by DWN’s aim to support small and large amounts application data (e.g. images) within an individual `RecordsWrite`. 

The cost of this architectural decision is incurred at query time. First, all records that match a given query are found. Then, an individual query is executed to fetch the data for each individual record. This means that a `RecordsQuery` matching 1000 records requires 1001 underlying sql queries.

repro:
```typescript
/* eslint-disable @typescript-eslint/no-unused-vars */

import ms from 'ms';
import { createPool } from 'mysql2';
import { faker } from '@faker-js/faker';
import { Dwn, DataStream, DidKeyResolver, Jws, RecordsWrite, RecordsQuery } from '@tbd54566975/dwn-sdk-js';
import { MysqlDialect, MessageStoreSql, DataStoreSql, EventLogSql } from '@tbd54566975/dwn-sql-store';

const mysqlDialect = new MysqlDialect({
  pool: async () => createPool({
    host     : 'localhost',
    port     : 3306,
    database : 'dwn',
    user     : 'root',
    password : 'dwn',
    debug    : ['ComQueryPacket', 'RowDataPacket']
  })
});

const messageStore = new MessageStoreSql(mysqlDialect);
const dataStore = new DataStoreSql(mysqlDialect);
const eventLog = new EventLogSql(mysqlDialect);

const dwn = await Dwn.create({ messageStore, dataStore, eventLog });
const didKey = {
  'did'     : 'did:key:z6MkfC8pVLTxqGa7UXPs3JLxVSyUwoch3bALhzRSdo9xFYm6',
  'keyId'   : 'did:key:z6MkfC8pVLTxqGa7UXPs3JLxVSyUwoch3bALhzRSdo9xFYm6#z6MkfC8pVLTxqGa7UXPs3JLxVSyUwoch3bALhzRSdo9xFYm6',
  'keyPair' : {
    'publicJwk': {
      'alg' : 'EdDSA',
      'kty' : 'OKP',
      'crv' : 'Ed25519',
      'x'   : 'CvvN8BqLXSVkUf7Ek89Z5j1HxTgG9bqrQo3xQu2cQXk'
    },
    'privateJwk': {
      'alg' : 'EdDSA',
      'kty' : 'OKP',
[...]

PR #456, opened 2023-07-27
Data below a threshold should not be stored in DataStore.

LiranCohen:
Data below a certain threshold should not be stored in `DataStore` but instead directly in `MessageStore` as an `encodedData` property.

This was brought up as a potential solution for: https://github.com/TBD54566975/dwn-sdk-js/issues/453

The way you deal with RecordsWrite/RecordsRead/RecordsQuery stays the same, this only affects the storage of messages.

Prior to this change any data below 10K would be embedded within the RecordsQuery response as `encodedData`, this will now be bumped to the 50k limit.

I opened this as a Draft PR, I still have some tests to do and would like to review it myself some more.

Any input/feedback would be appreciated.


thehenrytsai (changes requested):
🔥 🙏 🎸 

Thanks @LiranCohen! Pretty much there, with a few comments.

I have not looked at the tests thoroughly because my comments impacts the tests written!

src/handlers/messages-get.ts
```diff
@@ -68,18 +68,22 @@ export class MessagesGetHandler implements MethodHandler {
...
 
       // RecordsWrite specific handling
-      const recordsWrite = message as RecordsWriteMessage;
+      const recordsWrite = message as RecordsWriteMessageWithOptionalEncodedData;
       const dataCid = recordsWrite.descriptor.dataCid;
       const dataSize = recordsWrite.descriptor.dataSize;
 
-      if (dataCid !== undefined && dataSize! <= DwnConstant.maxDataSizeAllowedToBeEncoded) {
+      // Leaving this for backwards compatibility
+      if (dataCid !== undefined && dataSize! <= DwnConstant.maxDataSizeAllowedToBeEncoded && recordsWrite.encodedData === undefined) {
```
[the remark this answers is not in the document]
LiranCohen: Awesome, sounds good! I was actually thinking about back-populating but thought that was overkill. Good to know we can at this stage still just assume backward compatibility, definitely creates cleaner code.

src/handlers/records-write.ts
```diff
@@ -113,6 +119,54 @@ export class RecordsWriteHandler implements MethodHandler {
...
+    newestExistingMessage?: RecordsWriteMessage | RecordsDeleteMessage
+  ): Promise<RecordsWriteMessageWithOptionalEncodedData> {
+    let dataBytes;
+    if (dataStream === undefined) {
+      // dataStream must be included if message contains a new dataCid
+      if ( newestExistingMessage === undefined ||
+          newestExistingMessage?.descriptor.method === DwnMethodName.Delete ||
+          newestExistingMessage?.descriptor.dataCid !== message.descriptor.dataCid) {
+        throw new DwnError(
+          DwnErrorCode.RecordsWriteMissingDataStream,
```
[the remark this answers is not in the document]
LiranCohen: Yup agreed, sounds good, will do that.

src/handlers/messages-get.ts
```diff
@@ -68,18 +65,10 @@ export class MessagesGetHandler implements MethodHandler {
       }
```
diehuxx, line 65: Nit: There's a comment a few lines above here that says `! NOTE: this is somewhat duplicate code that also exists in StorageController.query.`. Do you mind removing that since it's not relevant anymore?

src/handlers/records-write.ts
```diff
@@ -113,13 +126,49 @@ export class RecordsWriteHandler implements MethodHandler {
     return messageReply;
   };
 
+  /**
+   * Embeds the given data into the `encodedData` property to be stored in the MessageStore instead of DataStore
```
diehuxx, line 130: Slight elaboration, since it took me a second to understand just reading the code.

```suggestion
   * Embeds the record's data into the `encodedData` property.
   * If dataStream is present, it uses the dataStream. Otherwise, uses the `encodedData` from the most recent RecordsWrite.
```

src/handlers/records-write.ts
```diff
@@ -79,23 +81,34 @@ export class RecordsWriteHandler implements MethodHandler {
...
+    if (options === undefined || !options.skipDataStorage) {
+      if (dataStream === undefined && newestExistingMessage?.descriptor.method === DwnMethodName.Delete) {
+        return messageReplyFromError(new DwnError(DwnErrorCode.RecordsWriteMissingDataStream, 'No data stream was provided with the previous message being a delete'), 400);
       }
 
-      // else throw
-      throw error;
+      try {
+        if (message.descriptor.dataSize <= DwnConstant.maxDataSizeAllowedToBeEncoded) {
+          message = await this.processEncodedData(
```
diehuxx, line 92: Can we use a variable other than `message`? I try to stay away from modifying function parameters, especially for a long one like this.
```suggestion
          messageWithOptionalEncodedData = await this.processEncodedData(
```

diehuxx (commented):
I missed the initial discussion around this PR, but I'm curious about an alternative approach. It makes sense that we need to reduce the number of queries, and moving smaller `data` into the `messageStore` accomplishes that. Have we considered adding more indexes to the message store? Currently store the data blob with `messageCid` as the only index. If we added a list of `indexes` to `dataStore.put`, we could `dataStore.query` on the same filters that we use for `messageStore.query`. This would take a more significant refactor of `data-store-level` to support, so I'm comfortable with merging this PR for now if we need to solve this n+1 query issue today.

Part of why I'm suggesting an alternate approach is that storing data in the message store just feels weird to me. While reviewing I found it confusing that the `messageStore.query` or `messageStore.get` would returns `RecordsWriteWithOptionalEncodedData` instead of the actual message.

LiranCohen:
@diehuxx I agree that it might be a bit confusing that RecordsWriteMessage can optionally have encodedData as an internal mechanism.

We should definitely have a more in-depth discussion at some point about this.

I added some comments in addition to your suggestions to hopefully make things a little clearer in certain places when reading the code.

diehuxx (dismissed):

diehuxx (approved):

thehenrytsai (approved):

Issue #449, opened 2023-07-21
Implement v2 encryption scheme


thehenrytsai:

thehenrytsai:
Done. Close.

PR #459, opened 2023-07-31
#449 - Replaced existing key derivation function using a hardened scheme

thehenrytsai:
This is the last PR for the re-worked DWN encryption.

Definitely could still use some work to make things more bearable such as more utility methods, documentation, code refactoring etc, will get to those when I am back!


diehuxx (approved):
Short and sweet. LGTM.

Issue #458, opened 2023-07-31
IndexLevel filter not working correctly


amika-sq:
I've been porting over the IndexLevel code to React Native, and ran into a bug with the range filters used by the `IndexLevel` class. 

An iterator is used to find what values are "less than" or "greater than" the target value. This breaks down, because the LevelDB database sorts these keys in byte order.

Here's an example of a failing test (`index-level.spec.ts`)
```
it('supports range queries with variable digit count', async () => {
  await index.put('1', { digit: 1 });
  await index.put('2', { digit: 2 });
  await index.put('10', { digit: 10 });

  const resp = await index.query({
    digit: {
      gte: '10'
    }
  });

  expect(resp.length).to.equal(1);
});
```

Instead of just one record being returned, we instead see that all three records are returned to the user. This is because the byte order of these keys are actually: `[10, 1, 2]`

---

This bug can also be seen with the `EqualFilter` as well with the following test:
```
it('supports equality queries', async () => {
  await index.put('1', { digit: 1 });
  await index.put('2', { digit: 2 });
  await index.put('10', { digit: 10 });

  const resp = await index.query({
    digit: 1
  });

  expect(resp.length).to.equal(1);
});
```

In this test, both the `"1"` and `"10"` records are returned, even though only one of those actually has a digit equal to 1.

thehenrytsai:
Closing, thanks @amika-sq!

PR #462, opened 2023-08-01
Allow for explicit number range filters in IndexLevel

amika-sq:
https://github.com/TBD54566975/dwn-sdk-js/issues/458

When querying the `IndexLevel` store, the user is allowed to provide a filter. One of the filters that the user can provide is a `RangeFilter`, which allows the user to specify that they want to query for values between given ranges. These ranges are provided via `gt`, `gte`, `lt`, and `lte` values.

Previously, these `gt`, `gte`, `lt`, `lte` value were only allowed to be `string` values. This caused some really subtle bugs. For example, let's say that the user wanted all records that have a particular key with values greater than 100, but less than 1000 (like in the test that is included within this PR). 

With string based comparisons, this was really difficult to determine. And in fact, it returned zero results in cases where there should actually be results. 

To accomodate, and make it extremely clear to the caller when doing a number filter, I added the ability to define a `RangeFilter` with a `number` in addition to the `string`. This way, it's clear to the caller when they provide the string `"100"` in a range filter, they're doing a string comparisons. 

To accomodate actual number comparisons, the logic had to change slightly compare the actual values instead of relying on the string ordering that LevelDB is doing under the covers with iterators. For example: the key `digit\u0000100` comes before the key `digit\u00002`, even though the value `100` is greater than `2`.


diehuxx (approved):
Thanks for the quick and clean fix!

diehuxx (dismissed):

diehuxx (approved):

PR #466, opened 2023-08-04
cjs compatibility

mistermoe:
This PR removes use of es6 `#` private fields (which were introduced in a recent commit) to regain cjs compatibility. we're retaining cjs due to electron's lack of esm support.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tbd54566975/dwn-sdk-js",
-  "version": "0.2.0",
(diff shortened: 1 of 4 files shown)


csuwildcat (approved):

frankhinek (dismissed):
Frustration continues to build while we wait for this PR to be merged https://github.com/electron/electron/pull/37535

but otherwise, a necessary workaround for now...

diehuxx (approved):

Issue #463, opened 2023-08-02
More granular protocol capabilities


diehuxx:
## Problem
The `$action` rules for `author` and `recipient` use the `of` keyword. The current implementation of `of` is limited and unintuitively specific given its generic name.

```json=
{
    "$actions": [
        {
            "who": "author",
            "of": "thread",
            "can": "read"
        },
        {
            "who": "recipient",
            "of": "thread",
            "can": "write"
        }
    ]
}
```

## Proposal
We will refactor `of` from protocol definitions, renaming it to `ofRecord` and turning it into an object. `ofRecord.atPath` will contain a string protocol path. `inGroup` will contain a string from the enum `'any' | 'ancestors' | 'context' | 'descendants'`.

```javascript
{
  "who": 'recipient',
  "ofRecord": {
    "atPath": 'foo/bar',
    "inGroup": 'any' | 'ancestors' | 'context' | 'descendants'
  },
  "can": 'write'
}
```

### Compatibility
Bad news, this is a breaking change. Good news, the existing functionality is still supported albeit with different syntax. To refactor your protocol definition, this find-and-replace regex worked for me in VSCode.
* Find: `([ ])"of": (".*")`
* Replace:
     ```
     $1"ofRecord": {
     $1  "atPath": $2,
     $1  "inGroup": "ancestors"
     $1}

### Group options
#### `ancestors`
`inGroup: 'ancestor'` is a renaming of the current behavior of `of`. `from: 'ancestor'` looks up the ancestor chain to find a record with a matching `who` and with `incomingMessage.protocolPath === action.ofRecord.atPath`.

#### `descendants`
`inGroup: 'descendants'` queries for descendants of the incoming record, looking for matching `who` and with `incomingMessage.protocolPath === action.ofRecord.atPath`. `action.ofRecord.atPath` will be the most troublesome to write a performant implementation because it involves first getting a list of children with matching `who` and `protocolPath`, then successively querying up each ancestor chain to ensure that the child record is actually a descendant of the incoming record message. Implementing this will also get us half way to #285.

Example: A podcaster stores episodes of their podcast in a DWN protocol. When a listener pays for an episode, the podcaster adds a `podcast/subscription` record with the listener as recipient. Using the `descendants` rule, the listener gains the ability to read the `podcast`.

```json=
"structure": {
    "podcastSeries": {
        "podcast": {
            "$actions": [
                {
                    "who": "recipient",
                    "ofRecord": {
[...]

PR #469, opened 2023-08-09
Support additional protocol action rule groups

diehuxx:
This PR implements the `any` and `context` rules of #463.

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -366,6 +366,7 @@ export class ProtocolAuthorization {
 
     // If this is reached, there is likely an issue with the protocol definition.
     // The protocolPath to the actionRule should start with actionRule.ofRecord.atPath
+    // consider moving this check to ProtocolsConfigure message ingestion
     if (ancestorRecordsWrite === undefined) {
       return false;
     }
@@ -399,7 +400,7 @@ export class ProtocolAuthorization {
     };
 
     if (actionRule.who === ProtocolActor.Recipient) {
-      // Find matching messages with recipient is the author of the inbound message
+      // Find matching messages where recipient is the author of the inbound message
       filter.recipient = author;
     } else { // actionRule.who === ProtocolActor.Author
       // Find matching messages authored by the author of the inbound message
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -1,9 +1,10 @@
+import type { Filter } from '../types/message-types.js';
 import type { MessageStore } from '../types/message-store.js';
 import type { RecordsRead } from '../interfaces/records-read.js';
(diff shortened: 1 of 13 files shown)


src/core/protocol-authorization.ts
```diff
@@ -283,62 +261,59 @@ export class ProtocolAuthorization {
    * Verifies the actions specified in the given message matches the allowed actions in the rule set.
    * @throws {Error} if action not allowed.
    */
-  private static verifyAllowedActions(
+  private static async verifyAllowedActions(
```
diehuxx, line 264: This method here is the centerpoint of the PR. Looking for feedback here and in functions called herein.

src/core/protocol-authorization.ts
```diff
@@ -283,62 +261,59 @@ export class ProtocolAuthorization {
...
 
     const actionRules = inboundMessageRuleSet.$actions;
-    if (actionRules === undefined) {
-      // if no action rule is defined, owner of DWN can do everything
-      if (author === tenant) {
-        return;
-      } else {
-        throw new Error(`no action rule defined for ${incomingMessageMethod}, ${author} is unauthorized`);
-      }
+    if (author === tenant) {
```
diehuxx, line 276: This addresses a tangential bug that @csuwildcat and I noticed. Previously, if there were no `$actions` rules, the tenant was allowed access; or if there were rules, the rules determined the tenant's access. Now the tenant always has access.

src/core/protocol-authorization.ts
```diff
@@ -18,24 +19,25 @@ export class ProtocolAuthorization {
...
    * Performs protocol-based authorization against the given message.
+   * @param recordsWrite Either the incomingMessage itself if the incoming is a RecordsWrite,
+   *                     or the latest RecordsWrite associated with the recordId being read.
    * @throws {Error} if authorization fails.
    */
   public static async authorize(
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
-    author: string | undefined,
+    recordsWrite: RecordsWrite,
```
LiranCohen, line 29: nit: Maybe there's a better name for what this record represents?

I'm notoriously bad at naming things, so I'm not sure if this is a better suggestion, but maybe something like `rootMessage` or `authorizationTarget`?

LiranCohen (commented):
Still wrapping my head around it all but I really like this feature!

Looks really good!

src/core/protocol-authorization.ts
diehuxx: how bout `newestRecordsWrite`?

csuwildcat (approved):

src/core/protocol-authorization.ts
diehuxx: Since I explain the variable in a comment, I'm gonna leave the name as-is for now. If we decide it's too confusing, we can change it later.

Issue #468, opened 2023-08-09
Tweak reads to not require a recordId, and just return the latest created record matching a query filter


csuwildcat:
We could tweak Reads to not require a recordId, such that if someone passed a protocol/protocolPath it would just return the latest record to be added for that indicated filter.

diehuxx:
What's the use case for this? This is not an intuitive behavior for a `Read` to me. This feels like a specific query.

csuwildcat:
The use case is so that if I have a record type `avatar` under a Profile protocol, I can do a Read of Profile Protocol + the `avatar` protocolPath and know I am getting the bytes for the singular latest record under that bucket.

diehuxx:
Does having `limit: 1` or possibly `latest: true` on queries solve this? If this could be solved with a small addition to RecordsQuery, I don't think we should add it to RecordsRead. My further worry is that it will set a precedent for other query-like parameters on read, muddying the difference between read and query.

@LiranCohen's notes from #470 show that this is already the intention.
> There is a potential to add additional fields to the description such as author, recipient, published etc for further filtering here.

> I will add contextId as an option when doing a Read based on protocol + protocolPath

frankhinek:
+1 to @diehuxx 's comments

Modifying `RecordsRead` in this way seems like an attempt to work around a limitation in `RecordsQuery` that ought to be addressed.

This use case also causes me to wonder: If we're adding query parameters to `RecordsQuery`, should we consider a parameter that returns the data payload regardless of the size if `limit: 1`?

thehenrytsai:
I shared similar sentiment as @frankhinek and @diehuxx, this seems like programming sugar to me, it also feels like we are attempting to workaround the design decision of generated `recordId` to mimic the benefit of predefined record ID.

LiranCohen:
We had some discussion around this during office hours and mentioned that `RecordsRead` is almost analogous to an HTTP GET req, which fits my mental model as well.

I think the main limitation of using `RecordsQuery ` is not necessarily limiting it to a single record, but rather actually reading the data that comes with that record as @frankhinek pointed out.

Maybe it would be worthwhile to discuss what type of improvements we can make to DataStore in order to support something like this, as we would need to allow streaming of large data with `RecordsQuery`. If that ends up being the case, it almost seems to eliminate any need for `RecordsRead` altogether.

[...]

Issue #467, opened 2023-08-09
Add a '$singleton' option for graph points to keep only the latest single record


csuwildcat:
Many times a graph point is intended to only have one record stored, never multiple. If we added a '$singleton' boolean option it would allow the dev to specify that an object in the graph was to only have one record, the latest one, kept and all others discarded. Thing of the case of having a blog protocol where you have an `index` html record where you only ever want one, don't really care about the recordId, and just want the latest kept. The $singleton option would make that possible.

diehuxx:
1. Can `$singleton` apply to records other than the root record?
  * If yes, how does that work? Say `foo/bar` is a `$singleton`. Can there only ever be one record at path `foo/bar` across contexts? If I try to write a new `foo/bar`, does that get rejected or delete the existing `foo/bar`, which may be in a different context.
  * If no, we need to validate upon protocol ingestion. It seems like the point of `$singleton` is to restrict a protocol to only one context.
1. I want to unpack this phrase `just want the latest kept`. If I have `$singleton` for protocol path `foo`, and a tree of descendant records below it, will writing a new record to path `foo` delete the entire existing tree? That's pretty drastic and dangerous behavior.

LiranCohen:
I've been working on how this would work and have some WIP PRs around it.

As discussed with @csuwildcat and I believe suggested by @diehuxx, a "$keep" property that is a positive integer greater than 0 rather than a "$singleton" property to only denote 1 would bring better flexibly.

Some thoughts on @diehuxx's questions above:

Yes, I think"$keep" records should be able to be nested.Currently when a new record is written the older ones are purged.

However the reject path is also interesting, and I could see how that might be useful. Maybe behind a protocol definition property that denotes the behavior of $keep, with default being purge. Maybe something like this:
```
"$keep" : {
    "limit" : 1,
    "strategy": "purge" | "reject"
}
```

For anything that is a child of a protocol context, the parentId and contextId are required upon creation of that record, currently I "$keep" the limit number of records within that context.

So you could have "foo/bar" with a a "$keep" limit of 5, you would then keep 5 "foo/bar"s for each parent instance of "foo".

If you had 5 "foo" records, you would then have a total of 25 "foo/bar" records. 
If you query only on protocolPath 'foo/bar' without a contextId you will get all 25 records.

Would like to get some input on this.

diehuxx:
@LiranCohen Looks good!

We discussed at office hours. I'll summarize:
[...]

PR #470, opened 2023-08-10
Tweak reads to not require a recordId

LiranCohen:
From https://github.com/TBD54566975/dwn-sdk-js/issues/468

This is a first pass, currently it accepts either `recordId` or `protocol` with `protocolPath` for a `RecordsRead`.
If the latest record is disallowed by protocol rules or not implicitly/explicitly allowed it will fail with a 401.

There is a potential to add additional fields to the description such as author, recipient, published etc for further filtering here.

Will think about how this fits in with https://github.com/TBD54566975/dwn-sdk-js/issues/467


diehuxx:
I'm not sure about this feature. This feels like we're conflating query and read. Especially given your comment @LiranCohen about possible adding "additional fields to the description such as author, recipient, published, etc." I added a comment in https://github.com/TBD54566975/dwn-sdk-js/issues/468.

LiranCohen:
After yesterday's office hours discussion I think this is ready to be reviewed.

As I was learning how protocols work, I had context + parent mixed up so i changed allowing to issue a Read based on a parentId along with a path instead of just the path.

Path based RecordsRead will look for the most recent Record which has not been Deleted. If no records are found at all it will 404.

src/interfaces/records-read.ts
```diff
@@ -57,6 +67,32 @@ export class RecordsRead extends Message<RecordsReadMessage> {
...
+   */
+  public static createFilter(descriptor: Partial<Omit<RecordsReadDescriptor, 'method'>>): Filter {
+    const filter: Filter = { interface: DwnInterfaceName.Records, isLatestBaseState: true };
+    const { recordId, protocol, protocolPath, parentId } = descriptor;
+    if (recordId !== undefined) {
+      return { ...filter, recordId };
+    } else if (protocol !== undefined && protocolPath !== undefined) {
+      if (parentId !== undefined) {
+        filter.parentId = parentId;
+      }
```
thehenrytsai, line 86: What is the use case for supporting `parentId`?

json-schemas/interface-methods/records-read.json
```diff
@@ -37,8 +36,30 @@
...
+        "protocolPath": {
+          "type": "string"
+        },
+        "parentId": {
+          "type": "string"
+        }
+      },
+      "$comment": "describe rule requiring either `recordId` OR `protocol` and `protocolPath`",
+      "allOf": [
+        {
```
thehenrytsai, line 52: Sanity: the `allOf` outer wrapping doesn't appear to be necessary?

src/interfaces/records-read.ts
```diff
@@ -57,6 +67,32 @@ export class RecordsRead extends Message<RecordsReadMessage> {
...
 
+  /**
+   * Creates a Records interface Filter using either the `recordId` or `protocol` & `protocolPath` of the incoming Message.
+   * Also filters on an optional `parentId` for `protocolPath` based reads
+   * @param descriptor message descriptor with optional properties `recordId`, `protocol` and `protocolPath`
+   *
+   * @returns {Filter} with a Records interface as well as the appropriate filter params
+   * @throws {DwnError} if `recordId` or `protocol` and `protocolPath` are missing
+   */
+  public static createFilter(descriptor: Partial<Omit<RecordsReadDescriptor, 'method'>>): Filter {
```
thehenrytsai, line 78: The ugly types like `Partial<Omit<RecordsReadDescriptor, 'method'>>` would typically suggest improvement or refactoring somewhere.

Perhaps this can be improved after the merge of #474

src/interfaces/records-read.ts
```diff
@@ -57,6 +67,32 @@ export class RecordsRead extends Message<RecordsReadMessage> {
     return new RecordsRead(message);
   }
 
+  /**
+   * Creates a Records interface Filter using either the `recordId` or `protocol` & `protocolPath` of the incoming Message.
+   * Also filters on an optional `parentId` for `protocolPath` based reads
+   * @param descriptor message descriptor with optional properties `recordId`, `protocol` and `protocolPath`
+   *
+   * @returns {Filter} with a Records interface as well as the appropriate filter params
+   * @throws {DwnError} if `recordId` or `protocol` and `protocolPath` are missing
```
thehenrytsai, line 76:
```suggestion
   * @throws {DwnError} if `recordId` and either `protocol` or `protocolPath` are missing
```

tests/interfaces/records-read.spec.ts
```diff
@@ -22,6 +23,93 @@ describe('RecordsRead', () => {
...
+      const readPromise = RecordsRead.create({
+        protocolPath                : 'some/path',
+        authorizationSignatureInput : Jws.createSignatureInput(alice),
+        date                        : currentTime
+      });
+
+      await expect(readPromise).to.be.rejectedWith('must have required property \'recordId\'');
+    });
+
+    it('should require `recordId` when `protocol` is set and `protocolPath` is not set', async () => {
```
thehenrytsai, line 50: Should we make it such that `recordId` and `protocol + protocolPath` are mutually exclusive?

src/interfaces/records-read.ts
```diff
@@ -57,6 +67,32 @@ export class RecordsRead extends Message<RecordsReadMessage> {
...
+   */
+  public static createFilter(descriptor: Partial<Omit<RecordsReadDescriptor, 'method'>>): Filter {
+    const filter: Filter = { interface: DwnInterfaceName.Records, isLatestBaseState: true };
```
thehenrytsai, line 79: Nit:
```suggestion
    const commonFilter: Filter = { interface: DwnInterfaceName.Records, isLatestBaseState: true };
```

src/interfaces/records-read.ts
LiranCohen: Was thinking this would be useful if you have a protocol with multiple children paths that you may want to read a record from.

Say you have a game protocol with multiple `game` paths and you'd like to read the latest `game/score` from a particular game, you would need that game's `parentId`. You could read the score of the latest updated game without it, or target a specific game.

As I mentioned in the Issue discussion https://github.com/TBD54566975/dwn-sdk-js/issues/468#issuecomment-1682174577, I'd be open to removing this if it doesn't fit the intended purpose of this change.

src/interfaces/records-read.ts
LiranCohen: Yeah, I really didn't love that either.

And yup! Looking at #474, it would eliminate this from being used outside the context of `RecordsRead`.

json-schemas/interface-methods/records-read.json
LiranCohen: Thanks for catching this, I removed it.

I was trying to have the schema validation respond with a message that was more accurate rather than `recordId` is mossing, but it became too verbose and was not much better in messaging.

Simplified here, but replied separately to the testing in another comment.

src/interfaces/records-read.ts
LiranCohen: Yeah this was worded poorly by me, ended up going with:
```
@throws {DwnError} when either `recordId` is missing, or both `protocol` and `protocolPath` are missing.
```

I think that's a better description.

tests/interfaces/records-read.spec.ts
LiranCohen: Yup, that's actually how the behavior is expected, but these tests were structured/worded poorly.

I've gone ahead and restructured the tests to better represent the intention, but this brought out that the error returned by the JSON Schema validator always asks for a `recordId`, which isn't really descriptive enough.

I was thinking about adding a specific DwnError here for create and failing fast before it even gets down to the Schema validation. This way the error can be more descriptive similar to the one thrown when receiving a bad RecordsRead during handle.

I'll still keep the rules that exist within the JSON Schema, but would just throw a better error before it gets there.

LiranCohen: I ended up doing this and added a DwnError: `RecordsReadMissingCreateProperties`

src/interfaces/records-read.ts
thehenrytsai: Let's remove it unless @csuwildcat have other thoughts. My understanding is that the whole point of this feature is to avoid needing to learning a CID that requires prior query, a scenario that needs `parentId` seems to defeat that entire purpose thus is out of scope.

src/interfaces/records-read.ts
thehenrytsai: Sounds good, let's resolve the conflicts now that #474 is in, I'll do a final pass on the PR!

json-schemas/interface-methods/records-read.json
```diff
@@ -37,8 +36,19 @@
...
+        "protocol": {
+          "type": "string"
+        },
+        "protocolPath": {
+          "type": "string"
         }
-      }
+      },
+      "$comment": "describe rule requiring either `recordId` OR both `protocol` and `protocolPath`",
+      "anyOf": [
```
thehenrytsai, line 48: Should we do `oneOf`?

src/interfaces/records-read.ts
```diff
@@ -57,6 +75,28 @@ export class RecordsRead extends Message<RecordsReadMessage> {
...
+    if (recordId !== undefined) {
+      return { ...commonFilter, recordId };
+    } else if (protocol !== undefined && protocolPath !== undefined) {
+      return { ...commonFilter, protocol, protocolPath };
+    } else {
+      throw new DwnError(
+        DwnErrorCode.RecordsReadMissingDescriptorProperties,
+        'missing required properties from RecordsRead descriptor, expected either `recordId`, or both `protocol` and `protocolPath` to be present'
+      );
+    }
```
thehenrytsai, line 97: Sanity: don't think we need this check, as JSON schema check would have caught this condition prior to this function call, so this is a condition only triggered by test.

Consider removing this condition, or if you prefer this for defensive programming, rename the error to something like `DwnErrorCode.RecordsReadCreateFilterUnreachableCode`

src/interfaces/records-read.ts
```diff
@@ -57,6 +75,28 @@ export class RecordsRead extends Message<RecordsReadMessage> {
...
   }
 
+  /**
+   * Creates a Records interface Filter using either the `recordId` or `protocol` & `protocolPath` of the incoming Message.
+   * @param descriptor message descriptor with optional properties `recordId`, `protocol` and `protocolPath`
+   *
+   * @returns {Filter} with a Records interface as well as the appropriate filter params
+   * @throws {DwnError} when either `recordId` is missing, or both `protocol` and `protocolPath` are missing.
+   */
+  public static createFilter(descriptor: RecordsReadDescriptor): Filter {
```
thehenrytsai, line 85: Let's move it into `handlers/reads-read.ts` now that this is only used by the handler.

tests/handlers/records-read.spec.ts
```diff
@@ -632,7 +633,292 @@ export function testRecordsReadHandler(): void {
...
+              const deleteRecord = await TestDataGenerator.generateRecordsDelete({ recordId: record.recordId, author: alice });
+              const deleteResponse = await dwn.processMessage(alice.did, deleteRecord.message);
+              expect(deleteResponse.status.code).to.equal(202);
+            }
+
+            const latestEmailReply2 = await dwn.handleRecordsRead(alice.did, latestEmail.message);
+            expect(latestEmailReply2.status.code).to.equal(404);
+          });
+
+          it('returns the latest RecordsWrite to the specified protocolPath when multiple parents exist in path', async () => {
```
thehenrytsai, line 822:
```suggestion
          it('returns the latest RecordsWrite to the specified protocolPath when multiple children exist in path', async () => {
```

src/handlers/records-read.ts
```diff
@@ -37,13 +37,9 @@ export class RecordsReadHandler implements MethodHandler {
...
     }
 
-    // get existing messages matching `recordId` so we can perform authorization
-    const query = {
-      interface : DwnInterfaceName.Records,
-      recordId  : message.descriptor.recordId
-    };
+    // get existing messages matching `recordId` or `protocol` with `protocolPath` so we can perform authorization
+    const query: Filter = RecordsReadHandler.createFilter(recordsRead.message.descriptor);
     const existingMessages = await this.messageStore.query(tenant, query) as TimestampedMessage[];
```
thehenrytsai, line 42: With this PR, the returned messages could be in 100s or even 1000s, and is subject to sort/paging. What were the scenario that prompted this change again? If it's for path that will ever have only one record, should we add a check for it?

tests/interfaces/records-read.spec.ts
```diff
@@ -22,6 +22,51 @@ describe('RecordsRead', () => {
...
+        authorizationSignatureInput : Jws.createSignatureInput(alice),
+      });
+
+      await expect(readPromiseSuccess).to.not.be.rejected;
+    });
+
+    it('should reject if only one of `protocol` or `protocolPath` are set', async () => {
+      const alice = await TestDataGenerator.generatePersona();
+      // with only protocolPath
+      const protocolPathOnlyP = RecordsRead.create({
```
diehuxx, line 48: What is the `P` at the end of `protocolPathOnlyP` and `protocolOnlyP`?

src/handlers/records-read.ts
```diff
@@ -37,13 +37,9 @@ export class RecordsReadHandler implements MethodHandler {
...
-
     const newestExistingMessage = await Message.getNewestMessage(existingMessages);
```
diehuxx, line 43: This will return the most recently _updated_ record, not the most recently created record. Is this the expected behavior?

tests/handlers/records-read.spec.ts
```diff
@@ -1096,6 +1382,22 @@ export function testRecordsReadHandler(): void {
         expect(readReply.status.code).to.equal(404);
       });
 
+      it('should return 404 on read without `recordId` or `protocol` and `protocolPath` to be set', async () => {
```
diehuxx, line 1385:
```suggestion
      it('should reject with 400 if neither `recordId` not `protocol`+`protocolPath` are set', async () => {
```

src/handlers/records-read.ts
```diff
@@ -86,4 +82,21 @@ export class RecordsReadHandler implements MethodHandler {
...
     return messageReply;
   };
+
+  /**
+   * Creates a filter using `recordId` in given descriptor, if not given, `protocol` & `protocolPath` are used to create the filter instead.
+   * @param descriptor message descriptor with optional properties `recordId`, `protocol` and `protocolPath`
+   *
+   * @returns {Filter} with a Records interface as well as the appropriate filter params
+   */
+  public static createFilter(descriptor: RecordsReadDescriptor): Filter {
```
diehuxx, line 92: Why `public`? If it's just for testing, I'd prefer to make this private, since testing this separately feels like testing implementation rather than behavior.

tests/handlers/records-read.spec.ts
```diff
@@ -1614,6 +1917,49 @@ export function testRecordsReadHandler(): void {
...
+
+        expect(filter['recordId']).to.equal('some-id');
+      });
+
+      it('should not throw if `protocol` and `protocolPath` are set', async () => {
+        const filter = RecordsReadHandler.createFilter({
+          interface        : DwnInterfaceName.Records,
+          method           : DwnMethodName.Read,
+          messageTimestamp : '2023-08-19T00:00:00.000000Z',
+
```
diehuxx, line 1938: extra whitespace here

tests/handlers/records-read.spec.ts
```diff
@@ -1614,6 +1917,49 @@ export function testRecordsReadHandler(): void {
...
+        expect(filter['protocol']).to.equal('some-protocol');
+        expect(filter['protocolPath']).to.equal('protocol/path');
+      });
+
+      it('should not set protocol filters if `recordId` is provided', async () => {
+        const filter = RecordsReadHandler.createFilter({
+          interface        : DwnInterfaceName.Records,
+          method           : DwnMethodName.Read,
+          messageTimestamp : '2023-08-19T00:00:00.000000Z',
+
```
diehuxx, line 1952: extra whitespace here too

tests/handlers/records-read.spec.ts
```diff
@@ -1614,6 +1917,49 @@ export function testRecordsReadHandler(): void {
...
+          protocol     : 'some-protocol',
+          protocolPath : 'protocol/path',
+          recordId     : 'some-id'
+        });
+
+        expect(filter['recordId']).to.equal('some-id');
+        expect(filter['protocol']).to.be.undefined;
+        expect(filter['protocolPath']).to.be.undefined;
```
diehuxx, line 1960: Any reason why you're using `['']` notation instead of dot notation here? I prefer dot notation since the typechecker isn't good with square bracket notation.

tests/handlers/records-read.spec.ts
```diff
@@ -1614,6 +1917,49 @@ export function testRecordsReadHandler(): void {
       });
     });
 
+    describe('createFilter()', async () => {
+      it('should not throw if only `recordId` is set', async () => {
```
diehuxx, line 1921: Sanity: This method never throws. Why test that it doesn't throw under these conditions?

tests/handlers/records-read.spec.ts
```diff
@@ -1614,6 +1917,49 @@ export function testRecordsReadHandler(): void {
...
+          method           : DwnMethodName.Read,
+          messageTimestamp : '2023-08-19T00:00:00.000000Z',
+
+          recordId: 'some-id'
+        });
+
+        expect(filter['recordId']).to.equal('some-id');
+      });
+
+      it('should not throw if `protocol` and `protocolPath` are set', async () => {
```
diehuxx, line 1933: Same as https://github.com/TBD54566975/dwn-sdk-js/pull/470/files#r1299587478
Sanity: This method never throws. Why test that it doesn't throw under these conditions?

tests/interfaces/records-read.spec.ts
```diff
@@ -22,6 +22,51 @@ describe('RecordsRead', () => {
...
+    });
+
+    it('should not reject if only `recordId` is passed', async () => {
+      const alice = await TestDataGenerator.generatePersona();
+      const readPromiseSuccess = RecordsRead.create({
+        recordId                    : 'some-id',
+        authorizationSignatureInput : Jws.createSignatureInput(alice),
+      });
+
+      await expect(readPromiseSuccess).to.not.be.rejected;
```
diehuxx, line 42: Nit: Rather than just testing that it doesn't reject, can you check that `recordId` is present once it resolves?

tests/interfaces/records-read.spec.ts
diehuxx, line 1: Please add a test for when all of `recordId`, `protocol`, and `protocolPath` are present.

tests/handlers/records-read.spec.ts
LiranCohen: Filter's type is generalized to `[property: string]: ...` so there is no strong typing of any of the filter properties.

src/handlers/records-read.ts
LiranCohen: The scenario is just to get the latest record under the path, but it doesn't have to be a "$singleton" record, we are moving away from singleton and toward $keep instead, which could have a value greater than 1.

So I don't think there are any checks here to provide, however we would add a limit to 1 and sort appropriately when that is exposed.

tests/interfaces/records-read.spec.ts
LiranCohen: The `P` stands for `Promise`, I usually put that as a naming convention when I haven't awaited the promise.

I can avoid that if you think it adds confusion (or use the full word).

src/handlers/records-read.ts
LiranCohen: Good call out, this was my understanding of the intended behavior, do you think that's unclear?

When you think about using this feature, do you find yourself thinking of scenarios that latest Created should be returned instead of updated?

src/handlers/records-read.ts
LiranCohen: It originally lived on RecordsRead interface and was used in multiple places before the Grants change, since it now lives within the handler and is only used here this could be private.

tests/handlers/records-read.spec.ts
LiranCohen: Henry made an update removing the explicitly thrown message here since JSON Schema will throw an exception elsewhere.

I'll update these tests to reflect that better.

tests/interfaces/records-read.spec.ts
thehenrytsai: I'd go for full word.

src/interfaces/records-read.ts
thehenrytsai: Hey @LiranCohen, yeah I understand the intent, but as it was, it was just for tests to hit the code, the actual running of DWN would never hit it thus I removed it.

As mentioned above, if you want to have the check in the same place, better to rename the error. If you want to move it so that it is actually called at runtime, then totally open to hear your proposal, but I suspect it will be a much more involved discussion.

src/handlers/records-read.ts
thehenrytsai: > The scenario is just to get the latest record under the path

This sounds more like a "requirement" but not scenario to me, do you know what the/an actual scenario that called for this requirement was?

src/handlers/records-read.ts
diehuxx: I was expecting the latest created to be returned tbh. What's the rationale for returning the latest updated vs latest created?

@csuwildcat What say you? To borrow/adapt one of your examples, I imagine someone has a blog that they add new blog posts to. They want readers to be able to `RecordsRead` `protocolPath: 'blog/post'` to get the latest created blog post. If the blog owner makes a spelling fix to an older post, they wouldn't want that older post to be returned.

tests/handlers/records-read.spec.ts
diehuxx: Filter's type is `[property: string]: EqualFilter | OneOfFilter | RangeFilter`, so there is typing of the value. For example, the following does not build because `{ bar: { baz: 1 } }` is not a `EqualFilter | OneOfFilter | RangeFilter`

```ts
const filter: Filter = {
  foo: { bar: { baz: 1 } },
};
```

src/handlers/records-read.ts
diehuxx: Word, makes sense. Can you make it private?

tests/handlers/records-read.spec.ts
```diff
@@ -632,7 +633,292 @@ export function testRecordsReadHandler(): void {
...
+            expect(readReply.status.detail).to.contain('must have required property \'recordId\'');
+          });
+
+          it('should retrieve the latest protocol record matching the path', async () => {
+            const testCases = [];
+            const alice = await DidKeyResolver.generate();
+
+            const protocolDefinition = emailProtocolDefinition as ProtocolDefinition;
+
+            // Install social-media protocol on Alice's DWN
```
diehuxx, line 689:
```suggestion
            // Install protocol on Alice's DWN
```
Since this the `email` protocol fixture, not `social-media`

LiranCohen:
After today's office hours call we hammered out some final details for this PR and agreed on the following:

`RecordsRead` will now take a `filter` with a `RecordsFilter` type within its descriptor. If that filter returns more than one result it will error back to the user, otherwise it will return the single result.

@thehenrytsai @diehuxx this is ready to be re-reviewed.

src/handlers/records-read.ts
```diff
@@ -37,15 +37,23 @@ export class RecordsReadHandler implements MethodHandler {
...
+      isLatestBaseState : true,
+      ...Records.convertFilter(message.descriptor.filter)
     };
-    const existingMessages = await this.messageStore.query(tenant, query) as TimestampedMessage[];
+    const existingMessages = await this.messageStore.query(tenant, query) as RecordsWriteMessage[];
 
-    const newestExistingMessage = await Message.getNewestMessage(existingMessages);
+    // ensure that the returned query only contains a unique record
+    try {
+      RecordsReadHandler.enforceSingleRecordRule(existingMessages);
```
diehuxx, line 51: Since we have `isLatestBaseState: true`, I think we can optimize by doing `if existingMessages.length > 1 then error`.

src/handlers/records-read.ts
```diff
@@ -37,15 +37,23 @@ export class RecordsReadHandler implements MethodHandler {
...
+    } catch (e) {
+      return messageReplyFromError(e, 400);
+    }
 
+    const newestExistingMessage = await Message.getNewestMessage(existingMessages);
     // if no record found or it has been deleted
     if (newestExistingMessage === undefined || newestExistingMessage.descriptor.method === DwnMethodName.Delete) {
```
diehuxx, line 58: Another optimization as a result of filtering by `isLatestBaseState: true`
```suggestion
    if (newestExistingMessage === undefined) {
```

src/handlers/records-read.ts
```diff
@@ -37,15 +37,23 @@ export class RecordsReadHandler implements MethodHandler {
...
     };
-    const existingMessages = await this.messageStore.query(tenant, query) as TimestampedMessage[];
+    const existingMessages = await this.messageStore.query(tenant, query) as RecordsWriteMessage[];
 
-    const newestExistingMessage = await Message.getNewestMessage(existingMessages);
+    // ensure that the returned query only contains a unique record
+    try {
+      RecordsReadHandler.enforceSingleRecordRule(existingMessages);
+    } catch (e) {
+      return messageReplyFromError(e, 400);
```
diehuxx, line 53: I've been wondering myself what status code we should send. Why 400? My gut tells me this isn't a "bad" request, but it's definitely not a 401. 404 also seems not quite right. Maybe 418? I'll put my musings aside and trust your judgement that 400 is good here. Just want to have it written somewhere if there's a reason.

diehuxx (commented):
I like this design a lot better :). Just a couple more small things then we can merge.

src/handlers/records-read.ts
LiranCohen: 🤣 yeah I had the same thoughts, although not 418 haha!

I landed on 400 because because it feels like a request error from the user's end. They provided some sort of incorrect filter that caused it.

I was thinking 404 at one point, but it felt not descriptive enough since there is a record there, just not only 1.

src/handlers/records-read.ts
LiranCohen: Great catch! That actually simplified a lot more here. 🙏

LiranCohen:
@diehuxx Yeah, I'm actually really happy where we landed with this design! It took a little to grow on me, but I think we made the right choices here.

src/handlers/records-read.ts
```diff
@@ -37,23 +37,26 @@ export class RecordsReadHandler implements MethodHandler {
       return messageReplyFromError(e, 401);
     }
 
-    // get existing messages matching `recordId` so we can perform authorization
-    const query = {
-      interface : DwnInterfaceName.Records,
-      recordId  : message.descriptor.recordId
+    // get the latest active records matching the supplied filter
```
thehenrytsai, line 40:
```suggestion
    // get the latest active messages matching the supplied filter
```

src/handlers/records-read.ts
```diff
@@ -37,23 +37,26 @@ export class RecordsReadHandler implements MethodHandler {
...
-    if (newestExistingMessage === undefined || newestExistingMessage.descriptor.method === DwnMethodName.Delete) {
+    const existingMessages = await this.messageStore.query(tenant, query) as RecordsWriteMessage[];
+    if (existingMessages.length === 0) {
       return {
         status: { code: 404, detail: 'Not Found' }
       };
+    } else if (existingMessages.length > 1) {
+      return messageReplyFromError(new DwnError(
+        DwnErrorCode.RecordsReadReturnedMultiple,
+        'Multiple records exist for the requested RecordRead filter'
```
thehenrytsai, line 55:
```suggestion
        'Multiple records exist for the RecordsRead filter'
```

src/core/dwn-error.ts
```diff
@@ -52,6 +52,7 @@ export enum DwnErrorCode {
   RecordsProtocolContextDerivationSchemeMissingContextId = 'RecordsProtocolContextDerivationSchemeMissingContextId',
   RecordsProtocolPathDerivationSchemeMissingProtocol = 'RecordsProtocolPathDerivationSchemeMissingProtocol',
   RecordsSchemasDerivationSchemeMissingSchema = 'RecordsSchemasDerivationSchemeMissingSchema',
+  RecordsReadReturnedMultiple = 'RecordsReadReturnedMultiple',
```
thehenrytsai, line 55: Super nit: should be line 54. 🤣

thehenrytsai (commented):
Looks good to me! I got nothing for ya except a few nits, then we get a final stamp of approval for me!

LiranCohen:
Alright, I think this is ready unless there's anything else that was missed! 🚀

src/handlers/records-read.ts
```diff
@@ -37,23 +37,26 @@ export class RecordsReadHandler implements MethodHandler {
...
-    if (newestExistingMessage === undefined || newestExistingMessage.descriptor.method === DwnMethodName.Delete) {
+    const existingMessages = await this.messageStore.query(tenant, query) as RecordsWriteMessage[];
+    if (existingMessages.length === 0) {
       return {
         status: { code: 404, detail: 'Not Found' }
       };
+    } else if (existingMessages.length > 1) {
+      return messageReplyFromError(new DwnError(
+        DwnErrorCode.RecordsReadReturnedMultiple,
+        'Multiple records exist for the RecordRead filter'
```
thehenrytsai, line 55:
```suggestion
        'Multiple records exist for the RecordsRead filter'
```

thehenrytsai (approved):
🎆 🍾 🎖️ 🚀 🎸

thehenrytsai (approved):

diehuxx (approved):

PR #474, opened 2023-08-14
Grant read and write access to protocol records

diehuxx:
`PermissionsGrants` for `RecordsRead` and `RecordsWrites` now have a field in the `scope` called `protocol`. Specifying the `protocol` field allows the grantee to read or write records from the protocol without restriction.

This PR does NOT yet support grants with specific `contextId` or `protocolPath` scopes.


src/core/protocol-authorization.ts
```diff
@@ -274,6 +275,17 @@ export class ProtocolAuthorization {
...
+      return;
+    } else if (incomingMessage.author !== undefined && incomingMessage.authorizationPayload?.permissionsGrantId !== undefined) {
+      // PermissionsGrant gives the author explicit access to this record
+      await RecordsGrantAuthorization.authorizeRecordsGrant(
+        tenant,
+        incomingMessage,
+        recordsWrite,
+        incomingMessage.author,
+        messageStore
+      );
```
thehenrytsai, line 288: Couple of questions for my understanding:
1. Why would a record written using a permission grant end up in this code path? I guess another way to ask the question: I saw in code further down that you flip the order or grant authorization and protocol authorization, what's the trigger of that?
1. Does a permission grant not have the capability to grant either read or read-write capability? If not, then it seems like we ought to have it for what seems to be very common scenarios.

src/core/records-grant-authorization.ts
```diff
@@ -25,36 +24,60 @@ export class RecordsGrantAuthorization {
...
+    grantScope: RecordsPermissionScope
+  ): void {
+    if (grantScope.protocol === undefined) {
+      throw new DwnError(
+        DwnErrorCode.RecordsGrantAuthorizationScopeNotProtocol,
+        'Grant for protocol record must specify protocol in its scope'
+      );
+    } else if (grantScope.protocol !== recordsWrite.message.descriptor.protocol) {
+      throw new DwnError(
+        DwnErrorCode.RecordsGrantAuthorizationScopeProtocol,
```
thehenrytsai, line 52:
```suggestion
        DwnErrorCode.RecordsGrantAuthorizationScopeProtocolMismatch,
```

src/core/protocol-authorization.ts
diehuxx: 1. The context is that `ProtocolAuthorization` does more than just the actual authorization of `RecordsWrites`. It also validates the structure of the `RecordsWrites`, e.g. validating the `protocolPath`. So, protocol `RecordsWrites` must enter `ProtocolAuthorization` even if a `PermissionsGrant` exists. After the previous steps of `ProtocolsAuthorization` are done, then actual authorization happens where we verify the grant.

2. A grant can be for `RecordsRead` or for `RecordsWrite`, but not both. Two separate grants are needed in order to give read AND write access.

src/core/records-grant-authorization.ts
```diff
@@ -16,6 +14,7 @@ export class RecordsGrantAuthorization {
   public static async authorizeRecordsGrant(
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite | RecordsDelete,
+    recordsWrite: RecordsWrite,
```
thehenrytsai, line 17: Would be nice to add a `@param` above, like what you've done in line 23, as it is unclear at a glance. Though I personally would probably prefer to structure the code slightly differently, with at least incoming `RecordsWrite` being on its own, even though there would be more code duplication.

src/core/protocol-authorization.ts
thehenrytsai: @diehuxx got it, thanks! So the order is critical,  must use `ProtocolAuthorization.authorize()` whenever `protocol` exists, even if a `permissionsGrantId` also exist.

diehuxx: Added a comment

src/core/records-grant-authorization.ts
diehuxx: I decided to split this into two separate methods, `authorizeRead` and `authorizeWrite`, which makes the interface a bit clearer. There is a small amount of code duplication but I was able to DRY it up in private methods.

src/core/records-grant-authorization.ts
```diff
@@ -1,27 +1,46 @@
...
 export class RecordsGrantAuthorization {
+  public static async authorizeWrite(
+    tenant: string,
+    incomingMessage: RecordsWrite,
+    author: string,
+    messageStore: MessageStore,
+  ): Promise<void> {
+    const permissionsGrantMessage = await GrantAuthorization.authorizeGenericMessage(tenant, incomingMessage, author, messageStore);
+
+    RecordsGrantAuthorization.authorizeScope(incomingMessage, permissionsGrantMessage);
```
thehenrytsai, line 18: Nit: can we rename to align with `verifyGrantActive()` and `verifyGrantScopeInterfaceAndMethod()`

I also want to rename `authorizeGenericMessage()` but can't think of a good name that covers all that occurs... so fine to leave as is for now.
```suggestion
    RecordsGrantAuthorization.verifyScope(incomingMessage, permissionsGrantMessage);
```

src/core/records-grant-authorization.ts
```diff
@@ -1,27 +1,46 @@
...
   ): Promise<void> {
-
-    // authorize generic message
     const permissionsGrantMessage = await GrantAuthorization.authorizeGenericMessage(tenant, incomingMessage, author, messageStore);
 
+    RecordsGrantAuthorization.authorizeScope(newestRecordsWrite, permissionsGrantMessage);
+  }
+
+  /**
+   * @param recordsWrite The source of the record being authorized. If the incoming record is a write,
```
thehenrytsai, line 37:
```suggestion
   * @param recordsWrite The source of the record being authorized. If the incoming message is a write,
```

src/core/records-grant-authorization.ts
```diff
@@ -1,27 +1,46 @@
...
+   *                     then this is the incoming. Otherwise, it is the newest existing RecordsWrite.
```
thehenrytsai, line 38:
```suggestion
   *                     then this is the incoming RecordsWrite. Otherwise, it is the newest existing RecordsWrite.
```

src/core/protocol-authorization.ts
```diff
@@ -274,6 +275,21 @@ export class ProtocolAuthorization {
 
     const actionRules = inboundMessageRuleSet.$actions;
     if (author === tenant) {
+      // tenant is always authorized
+      return;
+    } else if (incomingMessage.author !== undefined && incomingMessage.authorizationPayload?.permissionsGrantId !== undefined) {
+      // PermissionsGrant gives the author explicit access to this record
```
thehenrytsai, line 281: Sanity: sorry for the monkey-wrench, I noticed `verifyActionCondition()` call in line 73 will seem to cause problem if the initial content is written not by the author of the incoming message.

I am fine if we file an issue just to get this PR merged and address this in an immediate follow up PR.

diehuxx: Let's address this in a follow up PR. The tenant has this same restriction. We should all get on the same page the desired behavior for who can update a protocol record they did not initially author.
https://github.com/TBD54566975/dwn-sdk-js/issues/476

src/interfaces/records-read.ts
```diff
@@ -69,10 +69,10 @@ export class RecordsRead extends Message<RecordsReadMessage> {
     } else if (this.author !== undefined && this.author === descriptor.recipient) {
       // The recipient of a message may always read it
       return;
-    } else if (this.author !== undefined && this.authorizationPayload?.permissionsGrantId !== undefined) {
-      await RecordsGrantAuthorization.authorizeRecordsGrant(tenant, this, this.author, messageStore);
     } else if (descriptor.protocol !== undefined) {
       await ProtocolAuthorization.authorize(tenant, this, newestRecordsWrite, messageStore);
```
thehenrytsai, line 73:
```suggestion
      // All protocol RecordsWrites must go through protocol auth, because protocolPath, contextId, and record type must be validated
      await ProtocolAuthorization.authorize(tenant, this, newestRecordsWrite, messageStore);
```

thehenrytsai (approved):

PR #475, opened 2023-08-16
Update tenant-gate.ts

nearlyjuly:
The list of tenant DIDs for a DWN may be anything from single figures to many hundreds (perhaps thousands?). It's likely that there will be a custom solution to manage large numbers in particular, to set permissions or create a tenant removal process. This draft PR is a suggestion to make this a little easier, by having a custom tenantGate that can be set using `tenantGate   : new ExternalTenantGate(externalTenantCheck)` when initialising the DWN instance.

This would make it straightforward to pass in a function that checks if the DID is a tenant in any way, so long as the response is true or false.

Although it is straightforward to create a custom tenantGate, I feel this is a simple way to convey what is needed if someone is unfamiliar with this project. 

(If this is not a massive issue then I'll delete)

### DIFF
--- a/src/core/tenant-gate.ts
+++ b/src/core/tenant-gate.ts
@@ -16,13 +16,3 @@ export class AllowAllTenantGate implements TenantGate {
     return true;
   }
 }
-/**
- * A tenant gate that returns the value of externalTenantCheck, which requires a custom tenant checking function.
- */
-export class ExternalTenantGate implements TenantGate {
-  constructor(private externalTenantCheck: Function) {}
-  public async isTenant(did: string): Promise<boolean> {
-    const tenantCheck = await this.externalTenantCheck(did);
-    return tenantCheck;
-  }
-}
--- a/src/core/tenant-gate.ts
+++ b/src/core/tenant-gate.ts
@@ -16,3 +16,14 @@ export class AllowAllTenantGate implements TenantGate {
     return true;
   }
 }
+/**
+ * A tenant gate that returns the value of externalTenantCheck, which requires a custom tenant checking function.
+ * The default response is 'false'.
+ */
+export class ExternalTenantGate implements TenantGate {
+  constructor(private externalTenantCheck: Function) {}
+  public async isTenant(did: string): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


diehuxx (dismissed):
Thanks for the contribution @nearlyjuly ! You're right that we don't provide any explanation for custom tenant gating as far as I can find. I think a better place to provide an example is in the README. I'll push a commit to this PR in just a sec.

src/core/tenant-gate.ts
```diff
@@ -15,4 +15,14 @@ export class AllowAllTenantGate implements TenantGate {
...
     return true;
   }
-}
\ No newline at end of file
+}
+/**
+ * A tenant gate that returns the value of externalTenantCheck, which requires a custom tenant checking function.
+ */
+export class ExternalTenantGate implements TenantGate {
+  constructor(private externalTenantCheck: Function) {}
```
thehenrytsai, line 23: Not seeing a huge win here because the user will still need to implement `externalTenantCheck` somewhere. I see this as a convenience improvement and willing to approve it if `Function` is replaced with actual method signature, else it is arguably even less obvious (and error-prone) what is expected of the class.

Also agree with @diehuxx that adding explicit public facing documentation will help with understanding the intent and usage of tenant gate.

src/core/tenant-gate.ts
```diff
@@ -15,4 +15,14 @@ export class AllowAllTenantGate implements TenantGate {
   public async isTenant(_did: string): Promise<boolean> {
     return true;
   }
-}
\ No newline at end of file
+}
+/**
+ * A tenant gate that returns the value of externalTenantCheck, which requires a custom tenant checking function.
+ */
+export class ExternalTenantGate implements TenantGate {
```
thehenrytsai, line 22: Need to add to `index.ts`

src/core/tenant-gate.ts
thehenrytsai: Removed ExternalTenantGate.

diehuxx (approved):
thanks @thehenrytsai for the update

PR #477, opened 2023-08-18
Move level store exports to index.ts

amika-sq:
React Native currently cannot import `MessageStoreLevel`, `DataStoreLevel`, or `EventLogLevel`, as they are in a named export called `./stores`. 

React Native currently does not support named exports (but it is coming in a future update), so we have to manually import those classes from the `dist/esm/src/index-stores` directory. Example: https://github.com/TBD54566975/web5-wallet/pull/52#discussion_r1291668116

This is not ideal, but really causes issues when other packages use the `./stores` import. For example, in @frankhinek's agent refactor PR, we're using the `/stores` named export here: https://github.com/TBD54566975/web5-js/pull/168/files#r1298468034. React Native is not able to do this, causing the `@web5/agent` package to also be incompatible with React Native.

We have a couple of options here:
1. Move the `MessageStoreLevel`, `DataStoreLevel`, and `EventLogLevel` exports back into the primary `dwn-sdk-js` package (what this PR is doing).
2. Create a separate package just for the stores.

Perhaps there are more options than this, and I'd be open to other suggestions!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,3 +48,6 @@ export { RecordsRead, RecordsReadOptions } from './interfaces/records-read.js';
 export { SnapshotsCreate, SnapshotsCreateOptions } from './interfaces/snapshots-create.js';
 export { Secp256k1 } from './utils/secp256k1.js';
 export { SignatureInput } from './types/jws-types.js';
+export { DataStoreLevel } from './store/data-store-level.js';
+export { EventLogLevel } from './event-log/event-log-level.js';
+export { MessageStoreLevel } from './store/message-store-level.js';
--- a/README.md
+++ b/README.md
@@ -219,8 +219,7 @@ const result = await window.web5.dwn.processMessage({
 ### Custom Tenant Gating
 By default, all DIDs are allowed as tenants. A custom tenant gate implementation can be provided when initializing the DWN.
 ```ts
-import { Dwn, TenantGate } from '@tbd54566975/dwn-sdk-js';
-import { DataStoreLevel, EventLogLevel, MessageStoreLevel } from '@tbd54566975/dwn-sdk-js/stores';
+import { Dwn, TenantGate, DataStoreLevel, EventLogLevel, MessageStoreLevel } from '@tbd54566975/dwn-sdk-js';
 
 class CustomTenantGate implements TenantGate {
   public async isTenant(did): Promise<void> {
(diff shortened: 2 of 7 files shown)


shamilovtim (dismissed):

thehenrytsai:
@amika-sq, sorry that I am confused. The reason that we introduced named exports for Level implementations of stores and even took that path of less-convenient `Dwn` creation was precisely so that React Native can import the DWN SDK without Level stores, because I was told that Level isn't supported in React Native.

If a web5 library took a dependency on Level stores and blows up React Native, shouldn't the fix be not having level store as a dependency in its default export?

amika-sq:
`MessageStoreLevel`, `DataStoreLevel`, and `EventLogLevel` themselves do not directly use `Level`. Instead, it interacts with `LevelWrapper`, which is an `AbstractLevel`

Only the default behavior for `createLevelDatabase` [here](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/store/level-wrapper.ts#L14) directly interacts with `Level`.

React Native definitely supports abstract-level, as it's abstract. That's what is allowing us to provide an override for the `createLevelDatabase`, which creates an AbstractLevel instances that _is_ compatible with React Native ([example here](https://github.com/TBD54566975/web5-wallet/blob/main/src/features/dwn/dwn-service.ts#L23)).

thehenrytsai:
@amika-sq thanks for the clarification, so to summarize and make sure my understanding is correct:

`MessageStoreLevel`, `DataStoreLevel`, and `EventLogLevel` are not inherently incompatible with React Native, it's the default params to `createLevelDatabase` which creates an browser version of (but can be overridden) that is incompatible with React Native.

Follow up question:
Why is Level DB chosen? Level DB backed store was implemented because that's the only choice for browsers, I would have thought other types of DBs with better query capabilities such as SQL DB be better choices.

src/index.ts
```diff
@@ -47,4 +47,7 @@ export { RecordsQuery, RecordsQueryOptions } from './interfaces/records-query.js
 export { RecordsRead, RecordsReadOptions } from './interfaces/records-read.js';
 export { SnapshotsCreate, SnapshotsCreateOptions } from './interfaces/snapshots-create.js';
 export { Secp256k1 } from './utils/secp256k1.js';
-export { SignatureInput } from './types/jws-types.js';
\ No newline at end of file
+export { SignatureInput } from './types/jws-types.js';
+export { DataStoreLevel } from './store/data-store-level.js';
+export { EventLogLevel } from './event-log/event-log-level.js';
+export { MessageStoreLevel } from './store/message-store-level.js';
```
thehenrytsai, line 53: Need to update import paths in README https://github.com/TBD54566975/dwn-sdk-js#usage

amika-sq: Thanks for the callout. Updated this PR with the following commit: https://github.com/TBD54566975/dwn-sdk-js/pull/477/commits/92f35bc05f76f953fd0aa6c6a2131ac4d3f62b40

thehenrytsai (approved):

Issue #274, opened 2023-03-21
Consider supporting sort in Message Store


thehenrytsai:
Right now date sort is implemented after messages are fetched from the Message Store. This maybe undesirable when the returned dataset is large and we want/need to page the result.

thehenrytsai:
Thanks @LiranCohen for writing a proposal for this:
https://hackmd.io/@lirancohen/H1WYRrYsn

thehenrytsai:
This is done.

PR #478, opened 2023-08-18
MessageStore interfaces for sorting and pagination

LiranCohen:
Starting point for https://github.com/TBD54566975/dwn-sdk-js/issues/274

Modified the `DateSort` enum into the following types:
```
export enum RecordsDateSort {
  CreatedAscending = 'createdAscending',
  CreatedDescending = 'createdDescending',
  PublishedAscending = 'publishedAscending',
  PublishedDescending = 'publishedDescending',
}

export enum TimestampDateSort {
  TimestampAscending = 'timestampAscending',
  TimestampDescending = 'timestampDescending'
}

export type DateSort = RecordsDateSort | TimestampDateSort;

```

This allows explicit types for Records sorting(by created or published) vs generic message sorting (by timestamp).

Currently RecordsQuery specify the default sort to be `RecordsDateSort.CreatedDescending`
Everything else defaults to `TimestampDateSort.TimestampDescending`

Either as a part of this effort or as a different PR there can be some enhancements to certain parts of the code that either look for only 1 message, or attempt to sort the message they're looking for via Message.getNewestMessage or getOldestMessage.


src/types/message-types.ts
```diff
@@ -70,3 +70,22 @@ export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
 export type Filter = {
   [property: string]: EqualFilter | OneOfFilter | RangeFilter
 };
+
+export type Pagination = {
+  offset?: number
```
thehenrytsai, line 75: I am actually leaning towards cursor based pagination. Any reason why you lean towards offset based?

src/types/message-types.ts
```diff
@@ -70,3 +70,22 @@ export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
...
+};
+
+export enum RecordsDateSort {
+  CreatedAscending = 'createdAscending',
+  CreatedDescending = 'createdDescending',
+  PublishedAscending = 'publishedAscending',
+  PublishedDescending = 'publishedDescending',
+}
+
+export enum TimestampDateSort {
```
thehenrytsai, line 86: Unless there is requirement for exposing this, I think we can skip.

src/types/message-store.ts
```diff
@@ -35,7 +35,13 @@ export interface MessageStore {
   /**
    * Queries the underlying store for messages that match the provided filter.
    */
-  query(tenant: string, filter: Filter, options?: MessageStoreOptions ): Promise<GenericMessage[]>;
+  query(
+    tenant: string,
+    filter: Filter,
+    dateSort?: DateSort,
```
thehenrytsai, line 41: This is probably a point of contention, and I myself flip back & forth multiple times:

I am currently leaning towards a generic interface for the message store that minimizes the awareness of higher level constructs such as specific sorting enumerations. What do you think of:

```suggestion
    dateSort: { [key: string]: 1 | -1 },
```

This is purely intended as a point of discussion, the proposed change would:
1. Make the sort construct generic (though it could be harder to implement)
1. Follows design precedence of the filter being generic.
1. Follows design precedence of MongoDB (which filter also follows)
1. Make the parameter optional at this level, no specific reason, just my tendency of preferring required params if at all possible.

LiranCohen: I like this! Thanks for the suggestion, this is much more flexible... I'll look into implementing this.

src/types/message-types.ts
LiranCohen: I think you're right, especially given the general abstraction you mentioned below for sort.

src/types/message-types.ts
thehenrytsai: We probably will be asked to support both if we are successful, which would be a great problem to have. In the meantime though, I prefer to prioritize cursor based pagination mainly because:

1. yes, more performant, but also
2. the records retrieved is not impacted by records being added or removed in real-time during retrieval of pages: with offset pagination, you can miss records if records are being deleted and see duplicate records if records are being added in real-time.

Adding @csuwildcat and @diehuxx to see if they have their own preferences.

LiranCohen:
I did some more thinking about what this will look like when fully implemented with level, this is what I came away with and would love your thoughts @thehenrytsai & @diehuxx.

I generalized the `MessageSort` type as per @thehenrytsai's suggestion, however as I was thinking about truly generalizing the implementation to be inspired by MongoDB (which accepts multiple fields, and generalized sorting of fields) I wasn't sure if that kind of overhead was really necessary.

For every type of property we are sorting by, the number of indexes grows. So for now we should really only sort by `dateCreated` and `datePublished` where they are available (currently `RecordsWrite`), and sort by `messageTimestamp` as a default for everything else.

I think leaving the interface generalized is a good idea so that we can add other higher level abstractions with different types of dateSort fields to different message types and only need to add indexes.

Also thinking about pagination, cursor based pagination started making a lot more sense within the DWN context. I think we should start there and move to offset based pagination if there is a strong demand for it, as it will also add overhead to indexing and performance that might not be necessary.

Right now I am thinking that the cursor pointer for pagination be the `messageCid` and I went ahead and added that to the response in `RecordsQuery` since `authorization` is stripped. This cursor applies generically to all types of data that are stored within the message store and gives us an easy way to look up index starting points given the existing indexing scheme.

Would be happy to discuss this more in depth over a call. But for now figured looking at the interface and naive implementation will be a good place to start.

LiranCohen:
@thehenrytsai I think this is ready for a review at this point after yesterday's conversation.

I was going to add some types/properties to describe the sortable index fields somewhere, but it felt like premature optimization.

src/types/message-types.ts
```diff
@@ -70,3 +70,17 @@ export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
 export type Filter = {
   [property: string]: EqualFilter | OneOfFilter | RangeFilter
 };
+
+export type Pagination = {
+  messageCid?: string
```
thehenrytsai, line 75: Should we rename this to something like `cursor`? Even though a good choice is to use `messageCid`, it remains an implementation detail and another implementation may choose to use something else.

src/types/records-types.ts
```diff
@@ -83,6 +83,7 @@ export type UnsignedRecordsWriteMessage = {
  */
 export type RecordsQueryReplyEntry = RecordsWriteMessage & {
   encodedData?: string;
+  messageCid: string;
```
thehenrytsai, line 86: Design: I think we should have the cursor value outside of the array of entries, as metadata to be used for subsequent query.

src/types/message-types.ts
LiranCohen: I originally had it as cursor but wanted it to be more descriptive.

That's also an interesting viewpoint, I didn't really consider this as an implementation detail, would we want different implementations using a different type of cursor value?

Initially I didn't think so, it felt like the same results should be expected per the message regardless of implementation.

I'm still kind of leaning in that direction personally. This is why `messageCid` made sense, as every message in the system has a unique one.

But definitely open to changing my mind.

src/types/records-types.ts
LiranCohen: I think if we do land on a more general 'cursor' type from the above conversation, this is the right move.

But if we stay with specifically `messageCid` as the cursor, I think it might be a good idea to keep this here.

I might want to fetch the page prior to what I have again (sort the opposite direction with the first record's cursor).

src/types/message-store.ts
```diff
@@ -35,7 +35,13 @@ export interface MessageStore {
   /**
    * Queries the underlying store for messages that match the provided filter.
    */
-  query(tenant: string, filter: Filter, options?: MessageStoreOptions ): Promise<GenericMessage[]>;
+  query(
+    tenant: string,
+    filter: Filter,
+    dateSort: MessageSort,
+    pagination: Pagination,
```
thehenrytsai, line 42: Design: it seems to be a bit of anti-pattern to require a pagination object, then to have all its properties optional: we might as well make the `pagination` optional if that's the case to make calls cleaner.

Same comment for `dataSort`.

Seeing how many lines need to change by making `dataSort` and `pagination` required, I am willing to bend my default stance of making them required, however it does mean that "higher-level" logic (specifically a default for `dataSort`) will have to bleed into implementation. This is a point of discussion I suppose!

LiranCohen: I agree actually, I wanted to make it required due to a prior suggestion of yours, but it did feel off going through and adding empty objects everywhere.

I'll go ahead and change this to optional on the interface and where it's been modified.

I do think further down, specifically when this isn't just sugar, we will want to go in to those specific places and change some of those calls to be more specific (many of them really just want 1 item, sorted a specific way).

src/types/message-types.ts
thehenrytsai: Given this a bit more thought, and I still feel like while I agree the returned results should be the same, the marker/cursor/watermark used to denote progress should be left to the implementor, `messageCid` is really a byproduct/necessity of LevelDB implementation.

src/store/message-store-level.ts
```diff
@@ -89,7 +100,69 @@ export class MessageStoreLevel implements MessageStore {
...
+   * `dateCreated`, `datePublished`, `messageTimestamp`
+   *
+   * If sorting is based on date published, records that are not published are filtered out.
+   * @param messages - Messages to be sorted if dateSort is present
+   * @param sort - Sorting scheme
+   * @returns Sorted Messages
+   */
+  private sortRecords(
+    messages: GenericMessage[],
+    sort: MessageSort
```
thehenrytsai, line 143: After seeing this implementation I think your original ideas were better. If the implementor resorts into hardcoding the sortable index as was done here, we might as well define explicit interface to make expectation clear...

Perhaps something like:

export type MessageSort = {
  dateCreated?: SortOrder;
  datePublished?: SortOrder;
  messageTimestamp?: SortOrder;
};

src/types/records-types.ts
thehenrytsai: Upon further discussion, agree with you that `messageCid` as the cursor has the better trade offs.

json-schemas/interface-methods/records-query.json
```diff
@@ -38,6 +38,19 @@
         "filter": {
           "$ref": "https://identity.foundation/dwn/json-schemas/records-filter.json"
         },
+        "pagination": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "limit": {
+              "type": "number",
+              "minimum": 0
```
thehenrytsai, line 47: Sanity: what does it mean to have limit being zero? Should the minimum be 1?

src/handlers/records-query.ts
```diff
@@ -26,39 +27,38 @@ export class RecordsQueryHandler implements MethodHandler {
...
     }
 
-    let recordsWrites: RecordsWriteMessageWithOptionalEncodedData[];
-    if (recordsQuery.author === tenant) {
-      recordsWrites = await this.fetchRecordsAsOwner(tenant, recordsQuery);
-    } else {
-      recordsWrites = await this.fetchRecordsAsNonOwner(tenant, recordsQuery);
+    let paginationMessageCid: string|undefined;
+    const lastRecord = recordsWrites.at(-1);
+    if (lastRecord) {
```
thehenrytsai, line 52: Should we only return cursor if paging is specified in the query? Going even further would be to only return cursor if we KNOW that there are more entries to be returned, but this would further suggest that the cursor be returned from the query call: e.g. 

```ts
  query(
    tenant: string,
    filter: Filter,
    dateSort?: MessageSort,
    pagination?: Pagination,
    options?: MessageStoreOptions
  ): Promise<{ messages: GenericMessage[], paginationMessageCid: string }>;
  ```

File: src/handlers/records-query.ts [hunk 10]
```diff
@@ -75,19 +75,43 @@ export class RecordsQueryHandler implements MethodHandler {
...
+  /**
+   * Convert an incoming DateSort to a sort type accepted by MessageStore
+   * Defaults to 'dateCreated' in Descending order if no sort is supplied.
+   *
+   * @param dateSort the optional DateSort from the RecordsQuery message descriptor.
+   * @returns {MessageSort} for MessageStore sorting.
+   */
+  private convertDateSort(dateSort?: DateSort): MessageSort {
+    switch (dateSort) {
+    case DateSort.CreatedAscending:
```
thehenrytsai, line 87: No action item to this PR, but seems like the DWN specification is missing `dateModified`, which would be sorted against `messageTimestamp`

json-schemas/interface-methods/records-query.json
LiranCohen: This is an oversight, it should be 1.

src/handlers/records-query.ts
LiranCohen: Hrm... I think you're right we should only return the cursor if pagination is specified.

But I'm not sure about going the next step as only including it if we know there are more records. I think of it as just a cursor, not really knowing knowledge of what's after or before it.

Either way, I do think that having `paginationMessageCid` returned with `query` makes the most sense.
We would also at some point want to expose pagination on the other Query interfaces and it feels right for it to come from the query method of the store.

src/store/message-store-level.ts
```diff
@@ -89,7 +100,113 @@ export class MessageStoreLevel implements MessageStore {
...
+    const sortedRecords = await this.sortMessages(messages, messageSort);
+    return this.paginateRecords(sortedRecords, pagination);
+  }
+
+  private async paginateRecords(
+    messages: GenericMessage[],
+    pagination: Pagination = { }
+  ): Promise<GenericMessage[]> {
+    const { messageCid, limit = 0 } = pagination;
+    if (messageCid === undefined && limit > 0) {
```
thehenrytsai, line 112: Sanity: given the earlier than that limit will always be greater than zero, we should probably update the check
```suggestion
    if (messageId === undefined && limit !== undefined) {
```

src/handlers/records-query.ts
thehenrytsai: I updated the code according to my comment, see if you like/hate it.

src/store/message-store-level.ts
thehenrytsai: I updated the code according to my comment, see if you like/hate it:


https://github.com/TBD54566975/dwn-sdk-js/pull/478/commits/1125755bcf30e1e961bbe06ea3995dcccbbcd942

src/handlers/records-query.ts
thehenrytsai: @LiranCohen, responding to the money wrench you brought up regarding our code using multiple queries internally [(code)](https://github.com/TBD54566975/dwn-sdk-js/blob/de43a09c70eaef0aa5dc7520c5fdaae631dbff63/src/handlers/records-query.ts#L111) to serve the external query, complicating pagination logic...

Super glad that you catch this, we need to have a test that reproduce this so that we can confirm the scenario is handled and more importantly we don't regress.

The way I see it, there are two ways to approach this:

1. Combine multiple query into one somehow then pass it to the message store to execute. This requires message store to support OR predicates, something we've deliberately avoided to keep things simple and I still think it is the right decision given the resourcing we have. This is a can of worm I think we should avoid.
2. Add trimming logic accordingly in the query handler. Yes, we will need to incur 3 DB calls instead of 1 like we already do, but functionally it should/can work with some juggling, but avoidance of complexity increase is a worthwhile trade off IMO. This is what I am leaning towards.

Back to the discussion on whether we should have the message store query return the cursor, I am leaning towards it only if it's a low-enough hanging fruit, what I think is useful ultimately is a hint as to "if there are more message to fetch" or not, so technically a boolean would suffice. But keeping the code as is: externally always return a `messageCid` even though there might not be new messages is NOT the end of the world if you'd prefer this behavior, the ramification is "only" that the caller will need to make one extra call to know if there are more messages.

src/store/message-store-level.ts
```diff
@@ -77,19 +82,150 @@ export class MessageStoreLevel implements MessageStore {
...
+      return { messages }; // return all without pagination pointer.
+    }
+
+    // this is an optimization hack, we are passing the pagination message object for an easier lookup
+    // since we know this object exists within the the array if passed, we can assume that it will always have a value greater than -1
+    // if no pagination message is passed, we start from the beginning of the array, so we set the "cursor" to -1
+    const cursorIndex = paginationMessage ? messages.indexOf(paginationMessage) : -1;
+
+    // the first element of the returned results is always the message immediately following the cursor.
+    const start = cursorIndex > 0 ? cursorIndex + 1 : 0;
```
thehenrytsai, line 132: Sanity: is it not `>=` to handle the case when limit = 1?

```suggestion
    const start = cursorIndex >= 0 ? cursorIndex + 1 : 0;
```

src/store/message-store-level.ts
```diff
@@ -77,19 +82,150 @@ export class MessageStoreLevel implements MessageStore {
...
+    const end = limit === undefined ? undefined : limit + start;
+    const results = messages.slice(start, end);
+
+    // we only return a pagination MessageCid cursor if there are more results
+    const hasMoreResults = end !== undefined && end < messages.length;
+    let paginationMessageCid: string|undefined;
+    if (hasMoreResults) {
+      // we extract the cid of the last message in the result set.
+      const lastMessage = results.at(-1);
+      paginationMessageCid = lastMessage ? await Message.getCid(lastMessage) : undefined;
```
thehenrytsai, line 142: Sanity: why would the `lastMessage` be `undefined` if there are more results?

```suggestion
      paginationMessageCid = await Message.getCid(lastMessage!);
```

src/store/message-store-level.ts
thehenrytsai: Push minor changes for this, see if you like/hate it. Please resolve if it looks good to you!

src/store/message-store-level.ts
thehenrytsai: Push minor changes for this, see if you like/hate it. Please resolve if it looks good to you!

src/store/message-store-level.ts
```diff
@@ -77,19 +82,149 @@ export class MessageStoreLevel implements MessageStore {
...
+    pagination: Pagination = { }
+  ): Promise<{ messages: GenericMessage[], paginationMessageCid?: string } > {
+    const { limit } = pagination;
+    if (paginationMessage === undefined && limit === undefined) {
+      return { messages }; // return all without pagination pointer.
+    }
+
+    // this is an optimization, we are passing the pagination message object for an easier lookup
+    // since we know this object exists within the array if passed, we can assume that it will always have a value greater than -1
+    const cursorIndex = paginationMessage ? messages.indexOf(paginationMessage) : undefined;
```
thehenrytsai, line 128: Regarding this optimization of finding the cursor index to trim the result set, I would have thought a more performant alternative (if set it large) is to modify the queries themselves by looking up the corresponding sorted property value of the message. Is this considered? I am okay if you agree it's a good idea but want to file a `good-first-issue` and address this in a subsequent PR.

src/store/index-level.ts
```diff
@@ -73,60 +73,67 @@ export class IndexLevel {
...
+
+
+      for (const [propertyName, promises] of Object.entries(propertyNameToPromises)) {
+        // acting as an OR match for the property, any of the promises returning a match will be treated as a property match
+        for (const promise of promises) {
+          for (const dataId of await promise) {
+            // already matched
+            if (matchedIDs.includes(dataId)) {
+              continue;
+            }
```
thehenrytsai, line 123: Note to self: not understanding this short circuit.

src/store/message-store-level.ts
LiranCohen: Yup! I think after some thought and our discussion about sorting/pagination in general within Level (beyond what it's doing right now) I have some ideas, but they would definitely be suited for their own PR.

I agree that the optimization you mentioned would be a great `good-first-issue` as a first step in the overall pagination/sorting enhancements.

src/handlers/records-query.ts
LiranCohen: 💯 That is definitely a useful sort to expose.

src/store/message-store-level.ts
LiranCohen: Yup! looks clean! 🙏

src/handlers/records-query.ts
thehenrytsai: Reached agreement to go for option 3: introduce query array and let the store do the pagination and sorting.

src/store/message-store-level.ts
thehenrytsai: Added a TODO to #506 in code.

thehenrytsai (approved):
🥇 🎸 🚀 Long time coming!

Issue #479, opened 2023-08-18
Allow Range Filtering on negative numbers in LevelIndex


LiranCohen:
Recently there was a change made to address the ability to range digits within MessageStore/IndexStore, however this approach doesn't work with negative numbers.

I have a fix that will both support negative number range filters, as well as put us in the direction of allowing sorting and pagination to happen within the MessageStore/IndexStore.

PR #480, opened 2023-08-18
Allow Range filtering on negative numbers

LiranCohen:
https://github.com/TBD54566975/dwn-sdk-js/issues/479

This is ready to review, although there is coverage for it already from a higher level test, i'd like to write some function specific unit tests for `encodeNumberValue` and `extractValueFromKey`.


src/store/index-level.ts
```diff
@@ -180,61 +180,72 @@ export class IndexLevel {
...
+      for (const dataId of await this.findExactMatches(propertyName, rangeFilter.lte, options)) {
         matches.push(dataId);
       }
     }
 
     return matches;
   }
 
+  private NEGATIVE_OFFSET = Math.abs(Number.MIN_SAFE_INTEGER);
+  private NEGATIVE_PREFIX = '!'; // this will be sorted below positive numbers lexicographically
```
thehenrytsai, line 224: Can you use "earlier/later" instead of "below"?

src/store/index-level.ts
```diff
@@ -180,61 +180,72 @@ export class IndexLevel {
...
+      // thus would not be included in the iterator even though we'd like it to be.
+      for (const dataId of await this.findExactMatches(propertyName, rangeFilter.lte, options)) {
         matches.push(dataId);
       }
     }
 
     return matches;
   }
 
+  private NEGATIVE_OFFSET = Math.abs(Number.MIN_SAFE_INTEGER);
```
thehenrytsai, line 223: Sanity: any reason why not:
```suggestion
  private NEGATIVE_OFFSET = Number.MAX_SAFE_INTEGER;
```

src/store/index-level.ts
```diff
@@ -180,61 +180,72 @@ export class IndexLevel {
...
+  private NEGATIVE_PREFIX = '!'; // this will be sorted below positive numbers lexicographically
+  private PADDING_LENGTH = String(Number.MAX_SAFE_INTEGER).length;
```
thehenrytsai, line 225: use `static` or `const` for constants.

src/store/index-level.ts
```diff
@@ -180,61 +180,72 @@ export class IndexLevel {
...
       // For example, `'\x00'` becomes `'\\u0000'`.
       return `"${value}"`;
+    case 'number':
+      return this.encodeNumberValue(value);
+    default:
+      return String(value);
     }
+  }
+
+  private encodeNumberValue(n: number): string {
```
thehenrytsai, line 240: Would be nice to show an example of positive and negative encoded number either inline, or even better in method summary.

tests/store/index-level.spec.ts
```diff
@@ -266,63 +266,146 @@ describe('Index Level', () => {
...
     });
 
-    it('should return records that match provided number equality filter', async () => {
-      await index.put('a', { digit: 1000 });
-      await index.put('b', { digit: 100 });
-      await index.put('c', { digit: 10 });
+    describe('numbers', () => {
+      const randomInt = (min: number, max: number): number => {
+        return Math.floor(Math.random() * ( max - min) + min);
+      };
```
thehenrytsai, line 272: Consider moving this into `TestDataGenerator`.

src/store/index-level.ts
```diff
@@ -180,61 +180,72 @@ export class IndexLevel {
...
       // We can't just `JSON.stringify` as that'll affect the sort order of strings.
       // For example, `'\x00'` becomes `'\\u0000'`.
       return `"${value}"`;
+    case 'number':
+      return this.encodeNumberValue(value);
+    default:
+      return String(value);
     }
+  }
+
```
thehenrytsai, line 239: Would be nice to show an example of positive and negative encoded number either inline, or even better in method summary.

LiranCohen:
Thanks for the review @thehenrytsai! Cleaned up a bit as per your suggestions and added/fixed some tests.

thehenrytsai (approved):
🙏 🙏 🙏

Issue #481, opened 2023-08-21
Intermittent ProtocolsQuery grant test failures.


LiranCohen:
I get the following tests intermittently failing, even on main:
`rejects with 401 an external partys attempt to ProtocolsQuery if the grant is not yet active`

Noticed on this PR run:
https://github.com/TBD54566975/dwn-sdk-js/actions/runs/5928353825/job/16073770372?pr=480

Tried to re-create on `main` by re-running tests back to back and was eventually able to trigger the ProtocolsQuery failure on main locally.

I put some logging in before catching it and got the following:

protocolsQueryTimestamp: 2023-08-21T16:14:37.169477Z
dateGranted: 2023-08-21T16:14:37.169477Z

It seems that putting sleep(1) still produces an identical timestamp in some rare cases.
https://github.com/TBD54566975/dwn-sdk-js/blob/5b7ebb836b1c11498a41653a984f930b49823844/tests/handlers/protocols-query.spec.ts#L276C1-L276C1

Will update this to sleep for 2 ms.

PR #482, opened 2023-08-21
Sleep for 2ms to avoid time-sensitive test failures

LiranCohen:
Experienced this issue and tracked it down:
https://github.com/TBD54566975/dwn-sdk-js/issues/481

This might be some sort of precision bug within Temporal itself, I'm not really sure how various NodeJS systems handle time precision.

But in these cases there simply needs to be a gap in time for the testing logic, increasing the gap by an additional millisecond will make sure these tests are accurate.

### DIFF
--- a/tests/utils/test-data-generator.ts
+++ b/tests/utils/test-data-generator.ts
@@ -37,6 +37,7 @@ import type { PrivateJwk, PublicJwk } from '../../src/types/jose-types.js';
 import * as cbor from '@ipld/dag-cbor';
 import { CID } from 'multiformats/cid';
 import { DataStream } from '../../src/utils/data-stream.js';
+import { getCurrentTimeInHighPrecision } from '../../src/utils/time.js';
 import { PermissionsGrant } from '../../src/interfaces/permissions-grant.js';
 import { PermissionsRequest } from '../../src/interfaces/permissions-request.js';
 import { PermissionsRevoke } from '../../src/interfaces/permissions-revoke.js';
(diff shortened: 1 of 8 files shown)


thehenrytsai (dismissed):
🙏 
I wonder if it's worthwhile introducing a common util method like `minimalSleep()`.

LiranCohen:
@thehenrytsai I really like that idea, let me put that in this PR so we can set precedent.

tests/utils/test-data-generator.ts
```diff
@@ -821,4 +821,12 @@ export class TestDataGenerator {
...
     };
   }
+
+  /**
+   * We must sleep for at least 2ms to avoid timestamp collisions during testing.
+   * https://github.com/TBD54566975/dwn-sdk-js/issues/481
+   */
+  public static async minimalSleep(): Promise<void> {
+    await sleep(2);
+  }
```
thehenrytsai, line 831: Sorry for the final nit:

Let's put this directly in `time.ts`!

tests/utils/test-data-generator.ts
```diff
@@ -821,4 +821,12 @@ export class TestDataGenerator {
       didDocumentMetadata: {}
     };
   }
+
+  /**
+   * We must sleep for at least 2ms to avoid timestamp collisions during testing.
+   * https://github.com/TBD54566975/dwn-sdk-js/issues/481
```
thehenrytsai, line 827: Love that you provide context and issue number/URL for historical reference!!

thehenrytsai (approved):
🙏 🙏 🙏

PR #484, opened 2023-08-24
Issue PermissionsGrants reading or writing of protocolPath or contextId

diehuxx:
`PermissionsGrants` for `RecordsRead` or `RecordsWrite` can narrow their scope to either a specific `contextId` in a `protocol` or a specific `protocolPath` in a protocol. The properties `contextId` and `protocolPath` are mutually exclusive in order to align with record encryption scopes.


src/core/records-grant-authorization.ts
```diff
@@ -69,8 +69,21 @@ export class RecordsGrantAuthorization {
     } else if (grantScope.protocol !== recordsWrite.message.descriptor.protocol) {
       throw new DwnError(
         DwnErrorCode.RecordsGrantAuthorizationScopeProtocolMismatch,
-        `Grant scope specifies different protocol than in record with recordId ${recordsWrite.message.recordId}`
+        `Grant scope specifies different protocol than what appears in the record`
       );
+    } else {
```
thehenrytsai, line 74: Optional style nit: do you think four flat `if` without `else` or nesting is nicer to the eye? Okay if you prefer to leave as is.

src/types/permissions-types.ts
```diff
@@ -9,9 +9,11 @@ export type PermissionScope = {
...
 export type RecordsPermissionScope = {
   interface: DwnInterfaceName.Records;
-  method: DwnMethodName.Read | DwnMethodName.Write | DwnMethodName.Delete;
-  protocol?: string;
-  schema?: string;
+  method: DwnMethodName.Read | DwnMethodName.Write;
+  protocol?: string; // May only be present when `schema` is undefined
+  contextId?: string; // May only be present when `protocol` is defined and `protocolPath` is undefined
+  protocolPath?: string; // May only be present when `protocol` is defined and `contextId` is undefined
+  schema?: string; // May only be present when `protocol` is undefined
```
thehenrytsai, line 16: Might as well use `/** ... */ notation for the property comments so that the comments are picked up by IntelliSense and doc generator.

src/interfaces/permissions-grant.ts
```diff
@@ -104,4 +106,34 @@ export class PermissionsGrant extends Message<PermissionsGrantMessage> {
       );
     }
   }
+
+  public static validateScope(permissionsGrantMessage: PermissionsGrantMessage): void {
+    const scope = permissionsGrantMessage.descriptor.scope;
+    if (scope.interface === DwnInterfaceName.Records) {
```
thehenrytsai, line 112: Sanity: this line seems to read: "scope can be anything if interface is not `Records`", consider just remove, or add `else`.

src/interfaces/permissions-grant.ts
```diff
@@ -104,4 +106,34 @@ export class PermissionsGrant extends Message<PermissionsGrantMessage> {
...
+      // RecordsRead and RecordsWrite
+      if (scope.method === DwnMethodName.Read || scope.method === DwnMethodName.Write) {
```
thehenrytsai, line 114: Sanity: this two outer most `ifs` seem to read: "scope can be anything if interface is not `Records`" and "scope can be anything if methods are not read or write".

I removed the two outer `ifs` and all tests pass.

Consider just remove since condition is not reachable, or throw error in `else` if intent is defensive programming.

src/interfaces/permissions-grant.ts
```diff
@@ -104,4 +106,34 @@ export class PermissionsGrant extends Message<PermissionsGrantMessage> {
...
+        if (recordsScope.schema !== undefined) {
+          if (recordsScope.protocol !== undefined || recordsScope.contextId !== undefined || recordsScope.protocolPath) {
+            throw new DwnError(
+              DwnErrorCode.PermissionsGrantScopeSchemaProhibitedFields,
+              'PermissionsGrants for RecordsRead and RecordsWrite that have `schema` present may not also have protocol-related properties present'
+            );
+          }
+        }
+
+        if (recordsScope.protocol !== undefined) {
```
thehenrytsai, line 127: Sanity: does this imply `schema` is allowed when `protocol` is defined? Or this is disallowed else where in the code? (me lazy to hunt for it in code/tests)

src/interfaces/permissions-grant.ts
diehuxx: I'll remove for now. But here's some context: Grants for other DWN methods have scopes that can only contain `interface` and `method`. This is validated in JSON schema. In other words, grant scopes for DWN methods other than `RecordsRead` and `RecordsWrite` currently shouldn't ever have any of the fields checked inside the outer two `ifs`.

src/interfaces/permissions-grant.ts
diehuxx: The previous `if` statement (`if (recordsScope.schema !== undefined)`), we make sure that `protocol` is undefined when `schema` is present.

I'll add tests to `PermissionsGrantHandler` to make sure that we check these conditions upon ingest.

diehuxx: Added tests

src/interfaces/permissions-grant.ts
diehuxx: Removed

src/types/permissions-types.ts
diehuxx: Updated

src/core/records-grant-authorization.ts
diehuxx: Updated

thehenrytsai (approved):
🙏

PR #493, opened 2023-09-06
Refactor JSON Schema PermissionsRequest to use same scope as Grants

diehuxx:


json-schemas/interface-methods/permissions-grant.json
```diff
@@ -63,20 +63,7 @@
...
-              "$ref": "https://identity.foundation/dwn/json-schemas/permissions/scopes.json#/definitions/records-read-scope"
-            },
-            {
-              "$ref": "https://identity.foundation/dwn/json-schemas/permissions/scopes.json#/definitions/records-delete-scope"
-            },
-            {
-              "$ref": "https://identity.foundation/dwn/json-schemas/permissions/scopes.json#/definitions/records-write-scope"
-            }
-          ]
+          "$ref": "https://identity.foundation/dwn/json-schemas/permissions/defs.json#/definitions/scope"
```
thehenrytsai, line 66: For my understanding: why is this no longer the `oneOf` definition?

diehuxx: The `oneOf` got moved down a layer into `defs.json#/definitions/scope`. So it's still being used, just not in this file

thehenrytsai: Got it, would be nice if `defs.json` was named the same as `permissions-definitions.json` for consistency, but won't block the PR for it.

thehenrytsai (approved):

PR #494, opened 2023-09-07
Prevent resurrection of records that have been deleted

diehuxx:


src/handlers/records-write.ts
```diff
@@ -50,20 +50,25 @@ export class RecordsWriteHandler implements MethodHandler {
       recordId  : message.recordId
     };
     const existingMessages = await this.messageStore.query(tenant, query) as (RecordsWriteMessage|RecordsDeleteMessage)[];
+    const newestExistingMessage = await Message.getNewestMessage(existingMessages);
 
     // if the incoming write is not the initial write, then it must not modify any immutable properties defined by the initial write
     const newMessageIsInitialWrite = await recordsWrite.isInitialWrite();
     if (!newMessageIsInitialWrite) {
+      const initialWrite = await RecordsWrite.getInitialWrite(existingMessages);
+      if (initialWrite === undefined || newestExistingMessage?.descriptor.method === DwnMethodName.Delete) {
```
thehenrytsai, line 59: Returning "initial write is not found" when latest existing message is a delete seems to be misleading?

Is it possible to give specific error? Maybe with a check for the "latest message being a delete" after line 53?

src/handlers/records-write.ts
```diff
@@ -50,20 +50,30 @@ export class RecordsWriteHandler implements MethodHandler {
       recordId  : message.recordId
     };
     const existingMessages = await this.messageStore.query(tenant, query) as (RecordsWriteMessage|RecordsDeleteMessage)[];
+    const newestExistingMessage = await Message.getNewestMessage(existingMessages);
+    if (newestExistingMessage?.descriptor.method === DwnMethodName.Delete) {
+      return {
+        status: { code: 400, detail: `Record with id ${message.recordId} has been deleted` }
```
thehenrytsai, line 56: What about:

```suggestion
        status: { code: 400, detail: `Updating a deleted record (record ID ${message.recordId} ) is not allowed` }
```

diehuxx: I like it

diehuxx: Updated

LiranCohen (commented):
This looks good!

I'm on board with this general idea/behavior, but I have some sanity questions as this changes quite a bit WRT the system's behavior.

Currently we have some logic and a test for `RecordsDelete` around `should be disallowed if there is a newer RecordsWrite already in the DWN`.

The scenario is as follows:

- The user creates a RecordsWrite and sends it to their DWN. (accepted)
- The user creates a RecordsDelete for that same `recordId` and does not send it to their DWN.
- The user creates a 2nd RecordsWrite to update the original record referencing the same `recordId` and sends it to their DWN. (accepted)
- Alice sends the RecordsDelete which is older than the previously accepted `RecordsWrite`. Currently the DWN rejects this message.

I think this could create some inconsistency issues where one DWN might receive the `RecordsDelete` and be accepted and then 2nd `RecordsWrite` would be rejected, and a different DWN gets the messages in a different order and ends up in a different state.

Should we be purging the 2nd Write if a Delete with an earlier timestamp then appears, instead of rejecting it? Does this force us to rely on `messageTimestamp` too much? We already do rely on it a bit it seems.

diehuxx:
This was originally intended to deal with a constraint of `role` records, but now that I'm actively working on protocol `roles`, I don't think there's a need any more.

Here's the nuance:
If a `role` record is deleted with a retention list, then it saves what records the role has authorized. If that role is restored with a newer `RecordsWrite`, we purge the `RecordsDelete` and the corresponding retention list from the DB. Initially, I was worried about that, but I think it's fine since the role recipient has been given power again. If it is imperative that the retention list is retained, a new `role` record with a separate `recordId` can be created.

Issue #483, opened 2023-08-24
Investigate and choose new timer for DWN store open and close


frankhinek:
**Update: This has been partially addressed by #501, which fixed `LevelWrapper#open`. It is a good first issue to fix `LevelWrapper#closed` by following the same pattern as #501.**

While trying to track down a performance issue with the Web5 JS SDK, I noticed that a fairly significant portion of browser page load was attributable to instantiating a DWN instance (`Dwn.create()`).

Today it takes an average of 835ms

I got curious as to when this had increased since I didn't remember it taking this long the last time I had done rudimentary benchmarking

I found that up until `0.0.26` (Mar 2023) creating a DWN had averaged `220ms` , which had been the case as far back as I tested (Oct 2022)

Two commits in `0.0.25`/`0.0.26` nearly tripled the time (`220` -> `630` ms)
and another in `0.0.29` added an additional 33% (`630` -> `835` ms)

Altogether, it takes nearly 4x longer to instantiate a DWN as a result of these changes made in Mar/Apr.

<img width="1855" alt="Screenshot 2023-08-24 at 9 31 08 AM" src="https://github.com/TBD54566975/dwn-sdk-js/assets/134693/af339233-705b-477d-b5d8-981b2f7ec438">

frankhinek:
@mistermoe found the following `sleep()` calls in `Dwn.open()` and `Dwn.close()`:

https://github.com/TBD54566975/dwn-sdk-js/blob/285036b386a14e354e8c3cebaca6d7d9aabcb1b6/src/store/level-wrapper.ts#L51

https://github.com/TBD54566975/dwn-sdk-js/blob/285036b386a14e354e8c3cebaca6d7d9aabcb1b6/src/store/level-wrapper.ts#L71

Quick testing:
- with `200ms` sleep timers it takes [809ms](url) in a Node environment to `Dwn.create()`
- with `5ms` sleep timers it takes `24ms`
- with `1ms` sleep timers it takes just  `7ms`

thehenrytsai:
This is done by #501.

PR #501, opened 2023-09-14
Remove sleep() from db.open()

diehuxx:
Addresses #483.

Currently, when we open our `LevelWrapper` db, we `sleep(200)` repeatedly until it is open. Per the [abstract-level documentation](https://github.com/Level/abstract-level#events), the `db` emits events for `open`, `opening`, `closed`, and `closing`. So we simply need to wrap the event emitter in a promise and await that promise..

### DIFF
--- a/src/store/level-wrapper.ts
+++ b/src/store/level-wrapper.ts
@@ -60,7 +60,7 @@ export class LevelWrapper<V> {
     // If db is still opening, wait until the 'open' event is emitted
     case 'opening':
       return new Promise((resolve) => {
-        this.db.once('open', () => resolve());
+        this.db.once('open', resolve);
       });
 
     // If db is closing, wait until it is closed then await `db.open()`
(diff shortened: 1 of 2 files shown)


src/store/level-wrapper.ts
```diff
@@ -47,19 +47,43 @@ export class LevelWrapper<V> {
...
+
+      case 'closing':
+        // if db is closing, wait until it is closed then await `db.open()`
+        this.db.on('closed', onClosed);
+        break;
+
+      case 'closed':
+        // if db is closed, await `db.open`
+        await this.db.open();
+        resolve();
```
nitro-neal, line 83: you sure you want to reslove here? Will this trigger the onOpened above? Or those only trigger when you do 

```this.db.on('open', onOpened);``` ?

diehuxx: > Or those only trigger when you do `this.db.on('open', onOpened);` ?

This^^^

`onOpened` won't get triggered here since we didn't add the event listener.

thehenrytsai (dismissed):
This is dope! You are not imagining, the tests are running a lot faster! Or rather, back to the speed before the sleep was introduced.

Looks good to me. I just hope the event handlers are there for environments like react native...

Also considering filing a good first issue for following the pattern for `close()` also.

LiranCohen:
This is great! 🎸 🔥

thehenrytsai (approved):
Sensational!

Issue #206, opened 2023-01-31
Revisit `SignatureInput` to allow custom signers to sign messages


thehenrytsai:
`SignatureInput` currently requires private key. However there would be cases where private key is not available (e.g. key securing services, TPM, hardware-wallets).

A better model could be allowing a custom `Signer` to be injected, with perhaps a default implementation still expecting a private key as input.

frankhinek:
@thehenrytsai Would it be possible to take this approach?

Rather than injecting a custom `Signer` that the DWN SDK would call instead of the default implementation...
- IF a private key isn't passed in `authorizationSignatureInput ` to a given `.create()` method
- THEN an unsigned DWeb messages is returned

The developer would then have to handle signing with their "keystore" (TPM, secure enclave, hardware wallet, etc.) before passing the message to `processMessage()`.

This would allow for UI interactions that might be asynchronous and require human intervention, such as signing with a hardware wallet.

thehenrytsai:
Hey @frankhinek, definitely sounds reasonable in principle, though I would point out that it does mean that messages created by `.create()` methods could be "half-baked" and allows the situation where some instance methods/properties can't be used, (ie. will blow up if called) due to absence of the `authorization` property.

nearlyjuly:
Maybe there could be a parameter for the .create() methods with the default value being how it is now (private key was present and signatureMaterial has already been created), and then something like 'wallet' to read the private key/alg and create the signatureMaterial value within the method, plus a third option 'signer' where the signatureMaterial value will be passed in (as in it has been generated externally)

thehenrytsai:
This is now done.

PR #502, opened 2023-09-15
#206 - Introduced support for custom signer

thehenrytsai:
- Introduced new `Signer` interface for custom signer
- Replaced `PrivateJwk` with `Signer` in `SignatureInput`
- Renamed existing `Signer` to `SignatureAlgorithm` for disambiguation
- Renamed `GeneralJwsSigner` to `GeneralJwsBuilder`


src/utils/private-key-signer.ts
```diff
@@ -0,0 +1,31 @@
...
+import { signatureAlgorithms } from '../jose/algorithms/signing/signature-algorithms.js';
+import { DwnError, DwnErrorCode } from '../index.js';
+
+/**
+ * A signer that signs using a private key.
+ */
+export class PrivateKeySigner implements Signer {
+  private signatureAlgorithm;
+
+  public constructor(private privateJwk: PrivateJwk) {
```
diehuxx, line 13: Should we define/standardize what type the constructor takes in `Signer`?

diehuxx (commented):
A couple thoughts.

First, less pressing. In a follow up PR, we should migrate onto `@web5/crypto`. Thanks to @adam4leos that package has 100% test coverage. I can start on that after we get this PR merged.

Second, people are gonna have a hard time figuring out how to use a custom signer if we don't make it super clear in documentation. It's nested pretty deep in the options for `<Message>#create` where to pass a custom `Signer`. Do you mind adding a test that creates a DWN message with a custom signer just so we have it written down somewhere how to do it? We don't need the test coverage but it'll be useful to have something that we can link to. I anticipate people in Discord asking us how to use custom signers.

I'm wondering if there's a way to restructure the input to `#create()` methods that will make this more ergonomic or more self-documenting. Maybe flattening `authorizationSignatureInput: SignatureInput` into `protectedHeader?: JwsHeaderParameters; customSigner?: Signer;`?

Let's sync tomorrow. I've got some more thoughts based on how I see messages being created in web5-js that I can't express concisely right now.

README.md
```diff
@@ -234,6 +234,31 @@ const tenantGate = new CustomTenantGate();
 const dwn = await Dwn.create({ messageStore, dataStore, eventLog, tenantGate });
 ```
 
+### Custom Signature Signer
+Instead of using `PrivateKeySigner` to perform messger signing when you have the private key readily available, you can implement a customer signer to interface with external signing service, API, HSM, TPM etc and use it for signing your DWN messages:
```
diehuxx, line 238:
```suggestion
Instead of using `PrivateKeySigner` to perform message signing when you have the private key readily available, you can implement a customer signer to interface with external signing service, API, HSM, TPM etc and use it for signing your DWN messages:
```

diehuxx: Since we don't mention `PrivateKeySigner` elsewhere in the README, maybe phrasing it like:
"If you have the private key readily available, it is recommended to use the built-in `PrivateKeySigner`. Otherwise, you can implement a custom signer ..."

thehenrytsai: Good suggestion! Done. Orthogonally, will working on modifying the `Signer` interface as we discussed.

src/types/jws-types.ts
```diff
@@ -29,9 +29,22 @@ export type SignatureEntry = {
...
+   * Valid signature algorithm values can be found at https://www.iana.org/assignments/jose/jose.xhtml
+   */
   alg: string
-  /** JWS "kid" (Key ID) Parameter. */
+
+  /**
+   * JWS "kid" (Key ID) Parameter.
+   *
+   * This property is not a required property per JWS specification, but is required for DWN authentication.
+   * This needs to be a fully-qualified ID (ie. prefixed with DID) so that author can be parsed out for processing such as `recordId` computation.
```
LiranCohen, line 46: nit: Maybe show an example of what the expected fully-qualified `kid` looks like. 

```suggestion
   * This needs to be a fully-qualified ID (ie. prefixed with DID) so that author can be parsed out for processing such as `recordId` computation.
   * Example: did:example:alice#key1
```

LiranCohen (dismissed):
This looks great! The updated naming of Builder/Signer/SignatureAlgorithm definitely reads really nicely.

🎸 🚀

src/types/jws-types.ts
thehenrytsai: Thanks for the suggestion! Added.

thehenrytsai:
@diehuxx and @LiranCohen pushed an overhaul replacing `SignatureInput` with `Signer` directly. Tons of files are impacted, but the refactoring streamlines the code quite a lot IMO.

src/interfaces/records-write.ts
```diff
@@ -111,21 +112,21 @@ export type CreateFromOptions = {
...
   /**
    * Valid JSON message representing this RecordsWrite.
-   * @throws `DwnErrorCode.RecordsWriteMissingAuthorizationSignatureInput` if the message is not signed yet.
+   * @throws `DwnErrorCode.RecordsWriteMissingauthorizationSigner` if the message is not signed yet.
    */
   public get message(): RecordsWriteMessage {
     if (this._message.authorization === undefined) {
       throw new DwnError(
-        DwnErrorCode.RecordsWriteMissingAuthorizationSignatureInput,
+        DwnErrorCode.RecordsWriteMissingauthorizationSigner,
```
diehuxx, line 129: super nit: capitalization
```suggestion
        DwnErrorCode.RecordsWriteMissingAuthorizationSigner,
```

diehuxx (dismissed):

src/interfaces/records-write.ts
thehenrytsai: Not nit at all, keen eyes! Fixing!

diehuxx (approved):

PR #503, opened 2023-09-15
Remove extraneous TimestampedMessage type

diehuxx:
Previous `TimestampedMessage` referred to messages with `dateModified` in the descriptor. We removed `dateModified` and now all messages use `messageTimestamp`. So we can use type `GenericMessage` instead of `TimestampedMessage`.


thehenrytsai (approved):
Love this! Indication that we know "more" what we are doing over time.

PR #509, opened 2023-09-19
Bump version to v0.2.2

diehuxx:


thehenrytsai (dismissed):
the lock file might need to be added it after running non install

PR #514, opened 2023-09-22
Export Sort and Pagination types.

LiranCohen:
Exporting Filter, Sort and Pagination types for consumption by external MessageStores.

Lower max maxDataSizeAllowedToBeEncoded. We are currently checking this against the dataSize, which is the binary data size. However when we store it we are base64 encoding the data, which has an approximate increase of 33%, this was causing some issues going right above the limit in SQL.

Upon some rough calculations of the 65k limit for most SQL - 33% reduction, I rounded down to the nice number of 42_000.

### DIFF
--- a/src/core/dwn-constant.ts
+++ b/src/core/dwn-constant.ts
@@ -1,7 +1,9 @@
 export class DwnConstant {
   /**
-   * The maximum size in bytes of raw data that will be returned as `encodedData`.
-   * this is also the maximum size that we will store within MessageStore.
+   * The maximum size of raw data that will be returned as `encodedData`.
+   *
+   * We chose 30k, as after encoding it would give plenty of headroom up to the 65k limit in most SQL variants.
+   * We currently encode using base64url which is a 33% increase in size.
    */
-  public static readonly maxDataSizeAllowedToBeEncoded = 42_000;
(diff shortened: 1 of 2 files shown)


src/core/dwn-constant.ts
```diff
@@ -3,5 +3,5 @@ export class DwnConstant {
    * The maximum size in bytes of raw data that will be returned as `encodedData`.
    * this is also the maximum size that we will store within MessageStore.
    */
-  public static readonly maxDataSizeAllowedToBeEncoded = 50_000;
+  public static readonly maxDataSizeAllowedToBeEncoded = 42_000;
```
diehuxx, line 6: the meaning of life 💫

diehuxx: I wonder if we should make this configurable in `Dwn.create()`. Made an issue so we can revisit the idea when we have time. #515

diehuxx (approved):

src/core/dwn-constant.ts
thehenrytsai: Consider adding the context info above as comments directly in the code

LiranCohen: made the number more round with more headroom and added some comments to explain the size choice better.

thehenrytsai (approved):

PR #516, opened 2023-09-23
offline did resolution in tests were not working

LiranCohen:
Refactored to include the fetch method within the DidIonResolver class.

Now sinon can stub out fetch when the network is unavailable.

### DIFF
--- a/src/did/did-ion-resolver.ts
+++ b/src/did/did-ion-resolver.ts
@@ -1,22 +1,27 @@
-import type { DidMethodResolver, DidResolutionResult } from './did-resolver.js';
-
 import crossFetch from 'cross-fetch';
-// supports fetch in: node, browsers, and browser extensions.
-// uses native fetch if available in environment or falls back to a ponyfill.
-// 'cross-fetch' is a ponyfill that uses `XMLHTTPRequest` under the hood.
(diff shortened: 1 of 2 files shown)


src/did/did-ion-resolver.ts
```diff
@@ -1,22 +1,27 @@
...
-// indirectly.
-const fetch = globalThis.fetch ?? crossFetch;
+import type { DidMethodResolver, DidResolutionResult } from './did-resolver.js';
 
 /**
  * Resolver for ION DIDs.
  */
 export class DidIonResolver implements DidMethodResolver {
+
+  // since we are not always using global fetch, we set our fetch method within the constructor.
```
thehenrytsai, line 9: Sanity: I don't understand the comment on global fetch, initializing the fetch in the constructor certainly works, but is not the only way? (You can set it directly on line 10 too).

Maybe remove the comment or:
```suggestion
  // member field allows for test stubbing
```

LiranCohen: Yeah, there really is no purpose to putting it into the constructor unless we wanted to inject a fetch.

Since there's no need for that, for now I just set fetch directly within the class, and also added your comment suggestion about it being a member field.

thehenrytsai (approved):

PR #524, opened 2023-09-29
Authorize RecordsWrites with protocol roles

diehuxx:


src/core/protocol-authorization.ts
```diff
@@ -347,10 +342,7 @@ export class ProtocolAuthorization {
     }
 
     // Get role being invoked. Currently only Reads support role-based authorization
```
thehenrytsai, line 344: Remove.

src/interfaces/records-write.ts
```diff
@@ -663,7 +664,12 @@ export class RecordsWrite {
     if (contextId !== undefined) { authorizationPayload.contextId = contextId; } // assign `contextId` only if it is defined
     if (attestationCid !== undefined) { authorizationPayload.attestationCid = attestationCid; } // assign `attestationCid` only if it is defined
     if (encryptionCid !== undefined) { authorizationPayload.encryptionCid = encryptionCid; } // assign `encryptionCid` only if it is defined
-    if (permissionsGrantId !== undefined) { authorizationPayload.permissionsGrantId = permissionsGrantId; }
+    if (additionalProperties?.permissionsGrantId !== undefined) {
+      authorizationPayload.permissionsGrantId = additionalProperties?.permissionsGrantId;
+    }
+    if (additionalProperties?.protocolRole !== undefined) {
+      authorizationPayload.protocolRole = additionalProperties?.protocolRole;
+    }
```
thehenrytsai, line 672: Design: signature payload include properties that are not included in the descriptor (such as grant ID), but if protocol role is included descriptor, then is doesn't need to be included again in the payload.

But it seems that we ought to not tie role information used mainly for authorization purpose to the record itself, so we probably should remove the role info from the descriptor (seen below) like how grants are handled. Maybe confirm with @csuwildcat with this thinking.

README.md
```diff
@@ -48,7 +48,7 @@ Here's to a thrilling Hacktoberfest voyage with us! 🎉
 # Decentralized Web Node (DWN) SDK <!-- omit in toc -->
 
 Code Coverage
```
thehenrytsai, line 50: On code coverage: I noticed that there are a few placed related to roles (and grants) that missed branch (very small) coverage. Would be nice to catch them up so they don't fall further behind.

src/interfaces/records-write.ts
diehuxx: Updated -- Totally agree it should just be in `authorization`, I had meant to clean it up from the descriptor before opening the PR.

src/core/protocol-authorization.ts

README.md
diehuxx: Did some minor restructuring that covers those branches and hopefully improves readability somewhat.

src/interfaces/records-write.ts
```diff
@@ -399,7 +400,7 @@ export class RecordsWrite {
       await ProtocolAuthorization.authorize(tenant, this, this, messageStore);
     } else if (this.author === tenant) {
       // if author is the same as the target tenant, we can directly grant access
-    } else if (this.author !== undefined && this.authorizationPayload?.permissionsGrantId !== undefined) {
+    } else if (this.author !== undefined && this.authorizationPayload!.permissionsGrantId !== undefined) {
```
diehuxx, line 403: Correct me if I'm wrong, but I think if `author` is present, then `authorizationPayload` will be present. Or maybe your currently open PR changes things.

thehenrytsai: You are right that `authorizationPayload` should be present.

src/interfaces/records-write.ts
```diff
@@ -653,23 +653,19 @@ export class RecordsWrite {
...
+    const authorizationPayload: RecordsWriteAuthorSignaturePayload = {
+      recordId,
+      descriptorCid,
+      contextId,
+      attestationCid,
+      encryptionCid,
+      permissionsGrantId : additionalProperties?.permissionsGrantId,
+      protocolRole       : additionalProperties?.protocolRole
+    };
+    removeUndefinedProperties(authorizationPayload);
```
thehenrytsai, line 668: Good call! 🙏

README.md
thehenrytsai: Just did a sanity run there are a couple more annoying `?.` in `protocol-authorization.ts`, interfaces/protocols-`query.ts` and `interfaces/records-read.ts`.

And a more real block in `handlers/permissions-revoke.ts`.

Fine if you'd rather do this in a follow up PR.

thehenrytsai (dismissed):

README.md
diehuxx: I'll grab em in a follow up PR

thehenrytsai (approved):

Issue #439, opened 2023-07-13
Add JWS for 'owner', and separate it from existing 'author' JWS attribution


csuwildcat:

diehuxx:
Writing down our thoughts from our chat this afternoon:

### What use case does this solve?
It allows a given protocol context tree to be shared between multiple DWNs.

### Example
If Alice and Bob have a context tree of records with some `contextId`, they may want to invite Carol into the fold. Carol needs to establish a protocol context with the same `contextId` that Alice and Bob have, meaning she needs the exact same root `RecordsWrite` message that Alice and Bob have.

But how will Carol's DWN accept such a message if Alice/Bob pass her the exact message that they have? Alice signed the message as author, but Carol's protocol definition doesn't necessarily say "allow anyone to write a root record". I suppose Carol could create a `PermissionsGrant` to accept the root `RecordsWrite`, but that's heavy-handed. We can do better.

So we propose this flow:
1. Alice sends a `RecordsWrite` to Bob
2. Bob sends the `RecordsWrite` to Carol
3. Carol adds her own signature to the `authorization`. Now Alice's signature is in the `author` JWS, and Carol's signature is in the `owner` JWS.
4. Carol passes this updated `RecordsWrite` to her DWN, which sees that `owner === tenant` and accepts the message.

How does Carol decide whether she wants to add a signature and pass it to her DWN? We leave that to the application level.

csuwildcat:
@thehenrytsai has a PR out for this as of today.

PR #531, opened 2023-10-03
#439 - Added `RecordsWrite` support for retainer signature

thehenrytsai:
1. Added `RecordsWrite` support for retainer signature
2. Added `signAsRetainer()` method
3. A number refactoring and renaming due to this new feature

Note that I originally refactored the code with the assumption that upcoming delegate feature will make use `PermissionsGrant` but walked back most of the changes because the PR got more complicated due to increased scope.


frankhinek:
After reading this PR and not being part of the prior conversations, I was confused by the meaning of `retainer`.  At least for me, it was not intuitive.

I had to search through Slack messages and found a conversation explaining some of the back and forth -- and a couple calls to discuss the details.

After reading through the prior discussion on this it seemed like `tenant` received a ➕ 👍  from @kirahsapong / @csuwildcat

Adding my 2 🪙 in to say that `tenant' is much more intuitive than `retainer` in terms of informing me of which actor was responsible for signing.  Particularly in the case of a DWN given that it is inherently multi-tenant.

thehenrytsai:
@frankhinek and @csuwildcat, will not die on this hill, I'll give it one last shot at pushing back then will change it if no further response is received:

1. There is a reason why the current DWN spec has ZERO mentioning/use of the word "tenant": it’s not applicable in the context of a logical DWN and its specification. The term ‘tenant’ only becomes relevant in the context of a DWN service/implementation that supports multi-tenancy. For this reason alone, an alternative term is preferable, IMO.
2. ‘Message retainer’ conveys the intent much more effectively than ‘message tenant’, in my opinion: I would argue that it’s challenging to explain why ‘tenant’ is in ‘authorization’ without using the word ‘retain’ (or its synonyms), but the intent can be easily described without using the term ‘tenant’.

Daniel originally pointed out the potential confusion with the term ‘retention list’, which is a valid concern. However, the ‘retention list’ is no longer being considered.

tests/handlers/records-write.spec.ts
```diff
@@ -2622,59 +2767,57 @@ export function testRecordsWriteHandler(): void {
         expect(recordsWriteReply.status.detail).to.contain(DwnErrorCode.RecordsWriteMissingDataAssociation);
       });
 
-      describe('reference counting tests', () => {
```
diehuxx, line 2625: no pushback, just curious why you're removing flattening out this `describe()` block

src/types/protocols-types.ts
```diff
@@ -119,6 +119,7 @@ export type ProtocolRuleSet = {
 };
 
 export type ProtocolsConfigureMessage = GenericMessage & {
+  authorization: AuthorizationModel; // overriding `GenericMessage` with `authorization` being required
```
diehuxx, line 122: is this an opportunistic change or is it necessary for the PR somehow? no pushback, just asking for my own understanding.

diehuxx (approved):
LGTM -- no surprises, looks like what I expected

diehuxx (commented):
I don't have a very strong opinion on whether we call it `tenant` or `retainer` or `fooBarFlyingCircus`, but if I were to pick it myself, I'd call it `override` or `overrideSignature`.

frankhinek:
@thehenrytsai good points made on the downside to using `tenant`

Don't want to drag this out too long either but if you had to describe the semantic meaning of this property (regardless of what it is is called) how would you explain it to someone in 1 or 2 sentences?

thehenrytsai:
@frankhinek, it'd be something like:

"A signature that overrides the author signature used for authentication and authorization if the DWN owner wishes to retain a copy of a message that the owner did not author. The `author` field will still be used to enforce the structure of records within the protocol space"

Originally I did not like "owner" 'coz I thought we'd need to qualify what "owner" means but it could be interpreted as "owner of the message", which doesn't make any sense. I must admit that I am being a MORON overthinking this and concede that `owner` is a good compromise and that's literally what we started with (sorry @diehuxx !!)... but it is definitely better than "tenant" IMO.

I am also down with `override` over `tenant` for sure.

I am also down for renaming `author/owner` to `author/ownerSignature` to disambiguate `author/owner` being typically the DID string, but I'd do it in a separate PR as this PR is already touching 36 files and don't want to touch more.

Again, I apologize for overcomplicating things!!!

src/types/protocols-types.ts
thehenrytsai: Necessary, I realized that this condition is never hit in real code path (except by a unit test, which I also removed):

https://github.com/TBD54566975/dwn-sdk-js/pull/531/files/2c79913beec68c5a7f846cd749de0061182c7357#diff-ebe8b6ad983e00a9df763de521756d085f755da1cf9819d13766a1c84d485d22L36

The reason we had it was purely because of ambiguous typing, thus the "fix".

tests/handlers/records-write.spec.ts
thehenrytsai: I do NOT remember doing this at all... must've been a mistake... checking and reverting!

thehenrytsai: Undone. I probably did it unconsciously because there was only one test under the grouping? Anyway, reverted.

diehuxx (approved):
naming is hard

PR #540, opened 2023-10-05
Use contextRoles to authorize Reads and Writes

diehuxx:


src/core/protocol-authorization.ts
```diff
@@ -408,16 +415,27 @@ export class ProtocolAuthorization {
...
+      if (inboundMessageRuleSet.$globalRole) {
+        throw new DwnError(
+          DwnErrorCode.ProtocolAuthorizationDuplicateGlobalRoleRecipient,
+          `DID '${recipient}' is already recipient of a $globalRole record at protocol path '${protocolPath}`
+        );
+      } else {
+        // $contextRole
+        throw new DwnError(
+          DwnErrorCode.ProtocolAuthorizationDuplicateContextRoleRecipient,
+          `DID '${recipient}' is already recipient of a $context record at protocol path '${protocolPath} in the same context`
```
thehenrytsai, line 436: Nit
```suggestion
          `DID '${recipient}' is already recipient of a $contextRole record at protocol path '${protocolPath} in the same context`
```

tests/interfaces/protocols-configure.spec.ts
```diff
@@ -134,6 +168,62 @@ describe('ProtocolsConfigure', () => {
           .to.be.rejectedWith(DwnErrorCode.ProtocolsConfigureGlobalRoleAtProhibitedProtocolPath);
       });
 
+      it('rejects protocol definitions with $globalRole at records that are not root records', async () => {
```
thehenrytsai, line 171: C&P error

tests/handlers/records-read.spec.ts
```diff
@@ -604,6 +606,139 @@ export function testRecordsReadHandler(): void {
...
+              author       : alice,
+              protocol     : protocolDefinition.protocol,
+              protocolPath : 'thread/chat',
+              contextId    : threadRecord.message.contextId,
+              parentId     : threadRecord.message.recordId,
+            });
+            const chatRecordReply = await dwn.processMessage(alice.did, chatRecord.message, chatRecord.dataStream);
+            expect(chatRecordReply.status.code).to.equal(202);
+
+            // Bob invokes his 'participant' role to read the chat message
```
thehenrytsai, line 659: Can we add the following checks:

1. Bob can also read the `participants` records, most importantly his own (but also others I guess) this is how I am imagining encryption keys can be hand off to and persisted for Bob.
2. Bob can also read the `thread` record, this is needed for bob to keep a copy of the thread in his own DWN.
2.

src/types/protocols-types.ts
```diff
@@ -111,9 +111,16 @@ export type ProtocolRuleSet = {
...
    * may set $globalRole: true.
    * The recipient of a $globalRole record may invoke their role in RecordsRead or RecordsWrites
-   * by setting `asRole` property to the protocol path of the $globalRole record.
+   * by setting `protocolRole` property to the protocol path of the $globalRole record.
    */
   $globalRole?: boolean;
+  /**
+   * If true, this marks a record as a `role` that may used within a single context. Only
+   * second-level records may set $contextRole: true.
+   * They recipient of a $contextRole record may invoke their role in RecordsReads or RecordsWrites
```
thehenrytsai, line 120:
```suggestion
   * The recipient of a $contextRole record may invoke their role in RecordsReads or RecordsWrites
```

tests/handlers/records-read.spec.ts
diehuxx: Updated. Though for #1, Bob does not need to invoke his role to read his own role record. He is the recipient of the role record, and recipients are always authorized to read.

src/interfaces/protocols-configure.ts
```diff
@@ -79,24 +79,42 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
...
+        if (childRecordType.startsWith('$')) {
+          continue;
+        }
+        const childRuleSet: ProtocolRuleSet = rootRuleSet[childRecordType];
+        if (childRuleSet.$contextRole) {
+          contextRoles.push(`${rootRecordPath}/${childRecordType}`);
+        }
+      }
+
+      ProtocolsConfigure.validateRuleSet(rootRuleSet, rootRecordPath, [...globalRoles, ...contextRoles]);
```
thehenrytsai, line 95: Sanity: the placement of this line and the repeated concatenation of global and context roles threw me off: doesn't this mean that the `contextRoles` won't be built fully until the end of outer for-loop, leading to bugs in `validateRuleSet()`?

Consider moving this line into its own loop after the full `contextRoles` is built. I know it's literally repeating the same looping boilerplate 3 times, but I really liked it when you did it 2 times originally to reduce cognitive load, making it easy to comprehend and less error-prone!

src/interfaces/protocols-configure.ts
```diff
@@ -79,24 +79,42 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
...
+        `$contextRole is not allowed at protocol path (${protocolPath}). Only second-level records may set $contextRole true.`
+      );
     }
 
     // Validate $actions in the ruleset
     const actions = ruleSet.$actions ?? [];
     for (const action of actions) {
       // Validate that all `role` properties contain protocol paths $globalRole records
-      if (action.role !== undefined && !globalRoles.includes(action.role)) {
+      if (action.role !== undefined && !roles.includes(action.role)) {
```
thehenrytsai, line 117: Comment above (line 116) needs updating.

src/interfaces/protocols-configure.ts
```diff
@@ -79,24 +79,42 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
...
+          contextRoles.push(`${rootRecordPath}/${childRecordType}`);
+        }
+      }
+
+      ProtocolsConfigure.validateRuleSet(rootRuleSet, rootRecordPath, [...globalRoles, ...contextRoles]);
     }
   }
 
-  private static validateRuleSet(ruleSet: ProtocolRuleSet, protocolPath: string, globalRoles: string[]): void {
+  private static validateRuleSet(ruleSet: ProtocolRuleSet, protocolPath: string, roles: string[]): void {
```
thehenrytsai, line 99: More for self-reminder, 'coz I always forget, the method feels every time I review it! LOL.

```suggestion
  /**
   * Validates the given rule set structure then recursively validates the structure all its nested child rule sets.
   */
  private static validateRuleSet(ruleSet: ProtocolRuleSet, protocolPath: string, roles: string[]): void {
```

src/interfaces/protocols-configure.ts
diehuxx: We need to construct the `contextRoles` separately for each root record type. The list of `contextRoles` isn't unfinished, it's deliberately different for each iteration of the `for` loop.

Consider this protocol definition structure that has both `threads` and `blogs`, with context roles `thread/participant` and `blog/subscriber`. In the `thread` tree protocol rules, there should never be an `$action` rule containing `blog/subscriber` because there can never be a `blog/subscriber` in the same context as a `thread/chat`.

```javascript
{
  thread: {
     participant: {
       $contextRole: true,
     },
     chat: {
       $actions: [{
         role: 'thread/participant',
         can: 'read'
       },
       {
         // should not be allowed since 'blog/subscriber' roles can never exist in this context
         role: 'blog/subscriber',
         can: 'read'
       }]
     }
  },
  blog: {
    $actions: [{
      role: "blog/subscriber",
      can: "read"
    }]
    subscriber: {
      $contextRole: true
    }
  }
}
```

thehenrytsai: I see! Thanks, makes sense now!

PR #547, opened 2023-10-06
MessageStore Index tenant partitioning

LiranCohen:
The `MessageStores`'s `IndexLevel` doesn't take into account partitioning for the tenant.

This causes the following issue:
if `alice` + `bob` both put the same message in their respective tenant, the index updates accordingly and both can query.

But if `bob` deletes the message, `alice` can no longer query for the message, but she CAN do a get from the message store with the messageCid since that's still partitioned.

This PR adds a partition to the `IndexLevel` to index for each tenant separately. There are more storage-efficient ways of handing this, but puling in this low-overhead fix to handle the bug.

As a side note, we also allow "empty" querying, currently this is done by injecting the `tenant` into the index, as well as into the query. If an empty filter is provided, it will ALWAYS filter on the `tenant` property, therefore returning all of the results.

I left this behavior in tact as is for now with a note.

Also a note, this is partitioned properly on the `dwn-sql-store` repo already, the updated tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Here's to a thrilling Hacktoberfest voyage with us! 🎉
 # Decentralized Web Node (DWN) SDK <!-- omit in toc -->
 
 Code Coverage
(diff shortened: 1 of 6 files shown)


diehuxx (approved):
slickkkkk. thanks @LiranCohen

PR #549, opened 2023-10-07
Add Dwn.handleRecordsDelete function

shobitb:
This commit introduces a handleRecordsDelete function meant to be used in place of the existing processMessage (which is eventually meant to be replaced, as described in issue #289).

This commit also updates the relevant tests in records-(write|delete|read).ts files, which were using processMessage and replaces its usage with the new handleRecordsDelete.

It also introduces a test in dwn.spec.ts, enough to bring uncovered lines of this new function to 0 (more specifically, the new test is meant to cover the exception path).

### DIFF
--- a/src/types/records-types.ts
+++ b/src/types/records-types.ts
@@ -27,6 +27,8 @@ export type RecordsWriteDescriptor = {
 
 export type RecordsWriteReply = GenericMessageReply;
 
+export type RecordsDeleteReply = GenericMessageReply;
+
 /**
  * Internal RecordsWrite message representation that can be in an incomplete state.
  */
--- a/src/dwn.ts
+++ b/src/dwn.ts
@@ -8,7 +8,8 @@ import type { RecordsWriteHandlerOptions } from './handlers/records-write.js';
 import type { TenantGate } from './core/tenant-gate.js';
 import type { GenericMessageReply, UnionMessageReply } from './core/message-reply.js';
 import type { MessagesGetMessage, MessagesGetReply } from './types/messages-types.js';
-import type { RecordsQueryMessage, RecordsQueryReply, RecordsReadMessage, RecordsReadReply, RecordsWriteMessage, RecordsWriteReply } from './types/records-types.js';
+import type { RecordsDeleteMessage, RecordsDeleteReply, RecordsQueryMessage, RecordsQueryReply,
+  RecordsReadMessage, RecordsReadReply, RecordsWriteMessage, RecordsWriteReply } from './types/records-types.js';
 
 import { AllowAllTenantGate } from './core/tenant-gate.js';
 import { DidResolver } from './did/did-resolver.js';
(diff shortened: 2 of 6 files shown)


src/types/records-types.ts
```diff
@@ -27,6 +27,8 @@ export type RecordsWriteDescriptor = {
 
 export type RecordsWriteReply = GenericMessageReply;
 
+export type RecordsDeleteReply = GenericMessageReply;
```
thehenrytsai, line 30: One of the main goals for #289 is to ensure the return type is not generic and very specific to the message being handled.

See type `RecordsReadReply` that demonstrates this intent. Line 28 in this PR ought to be updated accordingly also, but that can be optional it being outside the context of this particular PR.

shobitb: Thanks @thehenrytsai. `RecordsReadReply` seems to be an extension of type `GenericMessageReply`:

```
export type RecordsReadReply = GenericMessageReply & {
  record?: RecordsWriteMessage & {
    data: Readable;
  }
};
````

If I understand your comment correctly,`RecordsWriteReply` and `RecordsDeleteReply` also similarly need to be extended types (so that they're more Write- and Delete-specific).  If yes, where can I find the specific data models for these types so I know what fields they need to include?  I didn't find them on [decentralized-web-node/spec/#response-objects](https://identity.foundation/decentralized-web-node/spec/#response-objects).

thehenrytsai: @shobitb, you are correct that the existing DWN spec is behind. The best way I can think of is to just look at the handler method, which I just realized after checking that `GenericMessageReply` is sufficient. I'd rather for us not introduce new types (`RecordsWriteReply` and `RecordsDeleteReply`)  unnecessarily in these two particular cases. Let's remove those two types and use `GenericMessageReply` directly and we can get this approved and merged!

thehenrytsai: Actually, let's just merge this in. I can make those changes if needed! Really appreciate your contributions!

thehenrytsai (approved):
🙏 🎖️ 💯

shobitb:
@thehenrytsai thanks for reviewing and approving!  I merged it, but had to resolve a conflict. I think it needs another review.

thehenrytsai (approved):
Approved! So FYI, @diehuxx made a change (#554) which uses a clever alternate scheme to provide TS type info.

So we can put a hold on the rest of the methods if it turns out to be sufficient! The work might then end up being to remove all dedicated methods such as this one instead. But for now, the PR is good to merge!

shobitb:
Wow, nice!  I learned something new about method overloading in TS today.  That's indeed very clever @diehuxx!

@thehenrytsai I don't have permissions to merge this above PR, but feel free to discard if you think that's the right direction anyway.

Thanks.

thehenrytsai:
@shobitb thanks so much for your understanding! In which case, I am going to close the PR. I have created #563 to track the new work, which would be perfect for you to work on (only if you are interested of course) now that you have the full context.

PR #552, opened 2023-10-10
Make protocol record validation and authorization distinct actions

diehuxx:
Currently, protocol record validation and authorization are coupled. This PR attempts to separate the two actions into distinct calls `ProtocolAuthorization.validate()` for RecordsWrites and `ProtocolAuthorization.authorize()` for any incoming message. There is some duplication of code across `validate` and `authorize`. The most significant of which is that we iterate up the ancestor chain with `constructAncestorMessageChain()`. If we think this is an unacceptable inefficiency, we can fix by updating `verifyProtocolPath` to look at only the parent instead of the entire ancestor chain.

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -25,10 +25,6 @@ export class ProtocolAuthorization {
     incomingMessage: RecordsWrite,
     messageStore: MessageStore,
   ): Promise<void> {
-    // fetch ancestor message chain
-    const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, incomingMessage, messageStore);
-
     // fetch the protocol definition
     const protocolDefinition = await ProtocolAuthorization.fetchProtocolDefinition(
       tenant,
@@ -43,9 +39,10 @@ export class ProtocolAuthorization {
     );
 
     // validate `protocolPath`
-    ProtocolAuthorization.verifyProtocolPath(
+    await ProtocolAuthorization.verifyProtocolPath(
+      tenant,
       incomingMessage,
-      ancestorMessageChain,
+      messageStore,
     );
 
     // get the rule set for the inbound message
@@ -172,6 +169,8 @@ export class ProtocolAuthorization {
       };
       const { messages: parentMessages } = await messageStore.query(tenant, [ query ]);
 
+      // We already check the immediate parent in `verifyProtocolPath`, so if it triggers,
+      // it means a bug that caused an invalid message to be saved to the DWN.
       if (parentMessages.length === 0) {
         throw new Error(`no parent found with ID ${currentParentId}`);
       }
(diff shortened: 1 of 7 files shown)


src/core/protocol-authorization.ts
```diff
@@ -18,25 +17,22 @@ const methodToAllowedActionMap: Record<string, ProtocolAction> = {
...
   /**
-   * Performs protocol-based authorization against the given message.
-   * @param recordsWrite Either the incomingMessage itself if the incoming is a RecordsWrite,
-   *                     or the latest RecordsWrite associated with the recordId being read.
-   * @throws {Error} if authorization fails.
+   * Performs validation on the structure of RecordsWrite messages that use a protocol.
+   * @throws {Error} if validation fails.
    */
-  public static async authorize(
+  public static async validate(
```
thehenrytsai, line 23: Can we rename to more precisely show the intent? I feel we have started to use "validate structure" in multiple context, e.g. there are "structure validations" that happens in `parse()` which begs the question why this couldn't be done there for people not familiar. Names below may make good alternate candidates:

`validateReferentialIntegrity()`, `validateReferentialConsistency()`

src/core/protocol-authorization.ts
```diff
@@ -52,6 +48,47 @@ export class ProtocolAuthorization {
...
+    // If the incoming message is writing a $globalRole record, validate that the recipient is unique
+    await ProtocolAuthorization.verifyUniqueRoleRecipient(
+      tenant,
+      incomingMessage,
+      inboundMessageRuleSet,
+      messageStore,
+    );
+
+    // verify allowed condition of incoming message
+    await ProtocolAuthorization.verifyActionCondition(tenant, incomingMessage, messageStore);
```
thehenrytsai, line 66: Could be arguable, but this feels more like actual authorization than "structure validation", I understand the code is "less clean" if it is `authorization` path because of the `RecordsRead` check that you removed. But I am okay with that tradeoff.

src/core/protocol-authorization.ts
```diff
@@ -18,25 +17,22 @@ const methodToAllowedActionMap: Record<string, ProtocolAction> = {
...
     tenant: string,
-    incomingMessage: RecordsRead | RecordsWrite,
-    recordsWrite: RecordsWrite,
+    incomingMessage: RecordsWrite,
     messageStore: MessageStore,
   ): Promise<void> {
     // fetch ancestor message chain
     const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, recordsWrite, messageStore);
+      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, incomingMessage, messageStore);
```
thehenrytsai, line 30: I am down with your suggestion on replacing `verifyProtocolPath()` with say `verifyParent()`

src/core/protocol-authorization.ts
diehuxx: Agreed on renaming -- I like `validateReferentialIntegrity`. I considered putting this inside `RecordsWrite.parse()` but decided against it because this method requires `tenant` and `messageStore`, which feel outside of the scope of what `parse()` should do/have.

src/core/protocol-authorization.ts
```diff
@@ -430,21 +443,25 @@ export class ProtocolAuthorization {
...
-      const { messages: result } = await messageStore.query(tenant, [ query ]);
-
-      // check the author of the initial write matches the author of the incoming message
-      const initialWrite = result[0] as RecordsWriteMessage;
-      const authorOfInitialWrite = Message.getAuthor(initialWrite);
-      if (recordsWrite.author !== authorOfInitialWrite) {
-        throw new Error(`author of incoming message '${recordsWrite.author}' must match to author of initial write '${authorOfInitialWrite}'`);
+  private static async verifyActionCondition(tenant: string, incomingMessage: RecordsRead | RecordsWrite, messageStore: MessageStore): Promise<void> {
+    if (incomingMessage.message.descriptor.method === DwnMethodName.Read) {
+      // Currently no conditions for reads
```
thehenrytsai, line 448: Optional nit: In preparation of other methods being supported in near future, probably just remove this `if` and perhaps say in comment after write handling:

`// Currently no conditions for other methods`

src/interfaces/records-write.ts
```diff
@@ -451,12 +451,6 @@ export class RecordsWrite {
       );
     }
 
-    // All protocol RecordsWrites must go through protocol auth, because protocolPath, contextId, and record type must be validated
-    if (this.message.descriptor.protocol !== undefined) {
-      await ProtocolAuthorization.authorize(tenant, this, this, messageStore);
-      return;
-    }
-
     // Remainder of the code is for flat-space writes
```
thehenrytsai, line 454: Remove.

thehenrytsai (dismissed):
Approved with a couple of small comments that could be addressed in a follow up PR.

src/interfaces/records-write.ts
diehuxx: Removed.

src/core/protocol-authorization.ts

thehenrytsai (approved):

Issue #525, opened 2023-09-29
Add range filter support for `dataSize`


thehenrytsai:
### Background:
Currently the supported records filter does not have `dataSize`

```ts
export type RecordsFilter = {
  attester?: string;
  recipient?: string;
  protocol?: string;
  protocolPath?: string;
  contextId?: string;
  schema?: string;
  recordId?: string;
  parentId?: string;
  dataFormat?: string;
  dateCreated?: RangeCriterion;
};
```

### Task Details:
Add range filter support for `dataSize`.

We can follow the pattern established by the `dateCreated` filter. But the existing `RangeCriterion` type only supports string, thus some creativity is required to either extend the type, or come up with a new type.

## Picking Up This Issue:
- If you'd like to work on this, please comment "picking this up" below, and I'll assign the issue to you 

## Questions:
- If you need assistance or clarification, feel free to comment below.
- Feeling overwhelmed? Join our [`#hack-together` channel](https://discord.gg/tbd) and collaborate with a buddy.
- New to open-source? Check out our [Hacktober `#getting-started` channel](https://discord.gg/tbd).
- [Leaderboard](https://github.com/TBD54566975/developer.tbd.website/issues/721): Check your ranking after finishing this task.

### Resources:
- **Creating a Pull Request**: If you're new to GitHub and unsure how to create a pull request, follow this [step-by-step guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Remember, communication is key! If you have any questions or face any challenges, we're here to help so please don't hesitate to reach out.

Good Luck! 🍁

captain-Akshay:
i would like to work on this!

thehenrytsai:
Hi @LynxSumit, thanks for your interest in contributing. Added the description above, let us know if you need further info.

shobitb:
@thehenrytsai There's also `RangeFilter` type which takes both string and number `RangeValue`:

https://github.com/TBD54566975/dwn-sdk-js/blob/11caeb4500d1edccff27e78fde96dc9ea4687106/src/types/message-types.ts#L70

Do you still need `RangeCriterion` to support number? Or was there a plan to converge to `RangeFilter` everywhere?

shobitb:
[...]

PR #553, opened 2023-10-11
Replace RangeCriterion with RangeFilter for `dateCreated` field

shobitb:
As part of the conversation in issue #525, it was decided to "converge on RangeFilter and remove RangeCriterion". This commit replaces the usage of RangeCriterion for the `dateCreated` field in RecordsFilter with the RangeFilter type. It also introduces a new json-schema for RangeFilter.

This commit also fixes the tests that were using RangeCriterion for the dateCreated filter.

Finally, this commit also removes the Records.convertFilter function as it's no longer needed to convert the input dateCreated RangeCriterion to RangeFilter type. Note that the convertFilter function was using the `lt` (i.e., exclusive) field for converting `dateCreated.to`, even though the RangeCriterion type said the `to` was inclusive. This logic has been preserved in the updated tests, i.e., the `to` fields are replaced with `lt` rather than `lte`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Here's to a thrilling Hacktoberfest voyage with us! 🎉
 # Decentralized Web Node (DWN) SDK <!-- omit in toc -->
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-98.37%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-95.11%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-95.7%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-98.37%25-brightgreen.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-98.38%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-95.12%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-95.7%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-98.38%25-brightgreen.svg?style=flat)
 
 - [Introduction](#introduction)
 - [Installation](#installation)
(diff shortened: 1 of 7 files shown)


json-schemas/interface-methods/range-filter.json
thehenrytsai, line 1: Probably easiest for us to rename this to `date-range-filter.json` or `string-range-filter.json` and modify the JSON schema as such. The problem is that this is used by the `dateCreated` filter right now, and if we allow number in this schema, then the DWN is going to allow numbers to be given for `dateCreated` filter which will result in unexpected error.

json-schemas/interface-methods/range-filter.json
```diff
@@ -0,0 +1,55 @@
...
+        { "type": "number" }
+      ]
+    },
+    "lte": {
+      "oneOf": [
+        { "type": "string" },
+        { "type": "number" }
+      ]
+    }
+  },
```
thehenrytsai, line 32: I am unsure that the newly allowed keywords `gt` and `lte` actually works, since we don't have test for them...

So either we:
- add tests to make sure they work; or 
- I'm okay if we just remove support for `gt` and `lte` at in this schema, calling this schema `date-range-filter.json`.

Obviously the latter option is a quicker way out, but I am okay with that. Though keep in mind when/if you work on the `dataSize` filter, we'd want to support all 4 keywords since it is "just" a number.

thehenrytsai (commented):
@shobitb, thanks again for this quality PR and picking up on the incorrect "inclusive" documentation! This PR is almost good for merge, with just a couple of comments regarding the `range-filter.json`

json-schemas/interface-methods/range-filter.json
shobitb: @thehenrytsai that makes sense, thanks.  However, if string-range-filter is used for `dateCreated`, wouldn't this still be a problem?  Because any string format other than ISO-8601 would still cause the query to not work correctly, and ISO-8601 is not enforced anywhere, as far as I understand from reading the code.  It wasn't immediately clear to me whether ISO-8601 is enforced during record storage.

shobitb: Another option is to introduce a date-time-range-filter with `{ "type": "string", "format": "date-time" }` https://json-schema.org/understanding-json-schema/reference/string#built-in-formats 

But this would require a new dependency on ajv-formats in package.json https://ajv.js.org/packages/ajv-formats.html#ajv-formats

I can create a separate issue for it, and only add `dataSize` with number-range-filter in this PR.

Let me know your thoughts!

Thoughts?

PR #559, opened 2023-10-13
EventsQuery interface and LevelIndex enhancement

LiranCohen:
This PR contains a new interface method, `EventsQuery`, which is primarily used to query for filtered events in order to enable SelectiveSync.

### **EventLogLevel:**
EventLogLevel now takes an additional property `indexes` to it's `append()` function. These indexes are key-value indexes in the same format that we use withe the MessageStore index.

In order to enable a filtered query for items like `RecordsDelete`, `PermissionsGrant` and `PermissionsRevoke` by additional indexes not currently needed in the `MessageStore` we added `constructAdditionalIndexes` methods to the proper handlers. This will allow, for example, to filter for both `RecordsWrite` and `RecordsDelete` events based on a `protocolPath` or `schema`

### **IndexLevel:**
To enable this functionality `IndexLevel` has been enhanced to become a more general purpose LevelDB Index that supports sorting as well as a cursor pointer property.

When appending data to the IndexLevel store, we save the sorted index values in a reverse lookup. During a query when a cursor value is passed, this lookup is used to re create a starting index for the individual filtered properties.

**note**
Although IndexLevel has been enhanced to support and a cursor pointer generically, for each sortable index the entire index grows in size. For that reason MessageStore is only indexing by the `messageTimestamp` field and is still sorting/paginating outside of the `IndexLevel` query as it has been doing.

Additional tests were added to IndexLevel as well as some changes to RecordsQuery tests and MessageStore tests to catch some missing edge cases.


json-schemas/interface-methods/events-query.json
```diff
@@ -0,0 +1,48 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://identity.foundation/dwn/json-schemas/events-query.json",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "descriptor"
+  ],
```
diehuxx, line 8: `authorizaton` not required?

src/handlers/permissions-grant.ts
```diff
@@ -29,22 +30,57 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+      author: permissionsGrant.author!,
+      ...propertiesToIndex,
+    };
+    return indexes;
+  }
+
+  /**
+   * Additional indexes that are not needed within the MessageStore but are necessary within the EventLog.
+   */
+  static constructAdditionalIndexes(
```
diehuxx, line 68: nit: rename to `constructIndexes` -> `constructMessageStoreIndexes` and `constructAdditionalIndexes` -> `constructAdditionalEventLogIndexes`.

src/handlers/events-query.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    try {
+      eventsQuery = await EventsQuery.parse(message);
+    } catch (e) {
+      return messageReplyFromError(e, 400);
+    }
+
+    try {
+      await authenticate(message.authorization, this.didResolver);
+      await authorize(tenant, eventsQuery);
```
diehuxx, line 29: I realized I never updated the error message in `authorize()`. Though it's tangential to this PR, do you mind updating that error message to something other than the current "permissions check not yet implemented"? Maybe, "Message failed authorization. Only the tenant is authorized."

src/types/event-types.ts
```diff
@@ -1,8 +1,20 @@
...
 import type { DwnInterfaceName, DwnMethodName } from '../core/message.js';
 
+export type EventsFilter = RecordsFilter & {
+  method?: string[];
+  // Default to all methods. Otherwise, explicitly subscribe to subset of methods.
+  interface?: string[];
+  // Default to all interfaces. Can be subset.
+  author?: string;
+
+  // optional watermark of the last message received from this filter.
```
diehuxx, line 14: Might as well use `/** */` comments and get benefit of some documentation

src/event-log/event-log-level.ts
```diff
@@ -76,46 +102,40 @@ export class EventLogLevel implements EventLog {
...
-        numEventsDeleted += 1;
+    const indexDeletePromises: Promise<void>[] = [];
+    for (const messageCid of messageCids) {
+      const watermark = await cidLog.get(messageCid);
+      if (watermark === undefined) {
+        continue;
       }
+      ops.push({ type: 'del', key: watermark });
+      cidOps.push({ type: 'del', key: messageCid });
+      indexDeletePromises.push(this.index.delete(tenant, messageCid));
```
diehuxx, line 132: Am I reading it correctly that we do not batch the deletion of indexes, but we delete the indexes one by one?

json-schemas/interface-methods/events-query.json
LiranCohen: 😱 it certainly is!

src/event-log/event-log-level.ts
LiranCohen: The individual indexes within the index log for each message is batched, but the index.delete() method is what handles the lookup/deleting of the individual indexes.

So for each message we add a promise that will delete the index for the message.

src/handlers/events-query.ts
diehuxx, line 29: Is selective sync only meant to allow one DID to sync data between their own DWNs? Or is it meant to allow two totally separate DWNs to sync e.g. a protocol context. If Alice has a protocol context, and Bob is trying to sync it, does he send an `EventsQuery` to Alice's DWN?

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
+    value: T,
+    indexes: { [key:string]: unknown },
+    sortIndexes: { [key:string]: unknown },
     options?: IndexLevelOptions
   ): Promise<void> {
-    const partition = await executeUnlessAborted(this.db.partition(tenant), options?.signal);
+    // ensure sorted indexes are flat and exist
+    sortIndexes = flatten(sortIndexes);
+    if (!sortIndexes || Object.keys(sortIndexes).length === 0) {
+      throw new Error('must include at least one sorted index');
```
diehuxx, line 48: Why must we include at least one sort index? If I understand correctly, since we didn't have sort indexes before, we effectively had zero sort indexes.

tests/store/index-level.spec.ts
```diff
@@ -1,136 +1,191 @@
...
+        `"${id}"`,
+        id,
+      );
+      const key = await partitionedDB.get(indexKey);
+      expect(key).to.not.be.undefined;
+      expect(JSON.parse(key!)).to.equal(id);
     });
 
-    it('removes empty objects', async () => {
+    it('adds 1 key per property, per sorted property, aside from id', async () => {
```
diehuxx, line 122: This test made me realize that I don't know the expected behavior of `put`ing the same key-value pair twice. I would have thought `put`ting to the same key without deleting would either:
1. Overwrite the existing value and indexes.
2. Throw error

But it looks like we actually have a third behavior:
3. Value is overwritten, keep existing indexes, AND add new indexes.

@thehenrytsai Curious your thoughts here. What do we expect to happen to existing indexes if we `IndexLeve#put` to the same key twice without deletion? How about if we `MessageStore#put` to the same `messageCid` twice? Or should it never happen that we `put` to the same key twice?

diehuxx: Alternate implementations of the `MessageStore` and `EventLog` should have the same behavior as our `level` implementation`. I haven't yet investigated how `dwn-sql-store` handles double-`put`s, but it's probably subtly different than `IndexLevel`.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
+    const matchPrefix = this.constructIndexedKey(sortProperty, propertyName);
 
     for (const comparator in rangeFilter) {
       const comparatorName = comparator as keyof RangeFilter;
-      iteratorOptions[comparatorName] = this.join(propertyName, this.encodeValue(rangeFilter[comparatorName]));
+      iteratorOptions[comparatorName] = this.constructIndexedKey(sortProperty, propertyName, this.encodeValue(rangeFilter[comparatorName]));
+    }
+
+    // if a cursor exists, it will be the starting point for the range query but not equal to.
+    if (cursor !== undefined) {
```
diehuxx, line 279: Move this check to top of function

src/handlers/permissions-grant.ts
```diff
@@ -29,22 +30,57 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+   */
+  static constructAdditionalEventLogIndexes(
+    permissionsGrant: PermissionsGrant
+  ): Record<string, string> {
+    let indexes: Record<string,any> = {};
+    const { scope } = permissionsGrant.message.descriptor;
+    if (scope.interface === DwnInterfaceName.Records) {
+      const { protocol, protocolPath, schema, contextId } = scope as RecordsPermissionScope;
+      indexes = {
+        ...indexes,
```
thehenrytsai, line 76: Sanity: what is this for? Not needed?

src/handlers/permissions-grant.ts
```diff
@@ -29,22 +30,57 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+    };
+    return indexes;
+  }
+
+  /**
+   * Additional indexes that are not needed within the MessageStore but are necessary within the EventLog.
+   */
+  static constructAdditionalEventLogIndexes(
+    permissionsGrant: PermissionsGrant
+  ): Record<string, string> {
```
thehenrytsai, line 70: Please avoid using `Record<K, V>` in this repo and use `[key: K]: V` instead. Normally it will be fine (desirable even) in other repo, but "Record" is a first-class term in DWN, so we agreed to avoid using it for/from other things to reduce cognitive load.

src/handlers/permissions-grant.ts
```diff
@@ -29,22 +30,57 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+      author: permissionsGrant.author!,
+      ...propertiesToIndex,
+    };
+    return indexes;
+  }
+
+  /**
+   * Additional indexes that are not needed within the MessageStore but are necessary within the EventLog.
+   */
+  static constructAdditionalEventLogIndexes(
```
thehenrytsai, line 68: Sanity: I am sure if you have thought this through already, but just in case: are you sure these addition indexesea

src/handlers/permissions-grant.ts
```diff
@@ -29,22 +30,57 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+    const indexes = PermissionsGrantHandler.constructMessageStoreIndexes(permissionsGrant);
 
     // If we have not seen this message before, store it
     const messageCid = await Message.getCid(message);
     const existingMessage = await this.messageStore.get(tenant, messageCid);
     if (existingMessage === undefined) {
       await this.messageStore.put(tenant, message, indexes);
-      await this.eventLog.append(tenant, messageCid);
+      const additionalIndexes = PermissionsGrantHandler.constructAdditionalEventLogIndexes(permissionsGrant);
+      await this.eventLog.append(tenant, messageCid, { ...indexes, ...additionalIndexes });
```
thehenrytsai, line 41: Sanity: I am sure if you have thought this through already, but just in case: are you sure we don't need to worry about addition indexes overwriting the indexes? Is it either not possible or desirable?

src/handlers/permissions-revoke.ts
```diff
@@ -83,13 +84,12 @@ export class PermissionsRevokeHandler implements MethodHandler {
...
-      method             : DwnMethodName.Revoke,
-      permissionsGrantId : message.descriptor.permissionsGrantId,
-    };
+    const indexes = PermissionsRevokeHandler.constructMessageStoreIndexes(permissionsRevoke);
     await this.messageStore.put(tenant, message, indexes);
-    await this.eventLog.append(tenant, await Message.getCid(message));
+
+    // get additional indexes for the associated grant message.
+    const additionalIndexes = await PermissionsRevokeHandler.constructAdditionalEventLogIndexes(permissionsGrantMessage);
+    await this.eventLog.append(tenant, await Message.getCid(message), { ...indexes, ...additionalIndexes });
```
thehenrytsai, line 92: I think I follow the pattern now, question:

Do you think if there is harm in the message store having the exact same indexes as the Events Store apart from the "additional" storage space used?

You'll probably counter with some questions also, but let's start with this question!

src/types/event-log.ts
```diff
@@ -1,13 +1,22 @@
...
 };
 
-
 export type GetEventsOptions = {
   gt: string
 };
 
+
+export type FilteredQuery = {
+  filter: Filter;
```
thehenrytsai, line 14: I've gotten confused by the ambiguity a few times in the past. With your permission, I'd like to propose that we should rename this to `AndFilter`, `Filters`, or `CompoundFilter`.

```suggestion
  andFilter: AndFilter;
```

src/store/index-level.ts
```diff
@@ -216,46 +287,203 @@ export class IndexLevel {
...
+      keyConstruction.push(sortValue);
     }
+    if (key !== undefined) {
+      keyConstruction.push(key);
+    }
+
+    return this.join(...keyConstruction);
+  }
+
+  async getFilterCursors(
```
thehenrytsai, line 367: This method is kinda vague to me upon encountering it, a few things I am not clear about:

1. Feels a summary description is useful given the complexity of the method.
1. Unsure why you'd need "multiple cursors".
1. Return type is.... unclear. Further glancing doesn't help, sometimes it returns `undefined`, sometimes it returns an empty object, and sometimes it returns an objects to Maps...

Would be useful to give some context in code here.

thehenrytsai (commented):
`getFilterCursors()` is where I got to today and still yet to wrap my head around it, will need to resume again tomorrow...

src/handlers/events-query.ts
LiranCohen: Yes currently EventsQuery and EventsGet are scoped to only work for the DID's own DWN.

src/store/index-level.ts
LiranCohen: This is part of generalizing IndexLevel to allow different sorting properties. Before this change all of the sorting was done after retrieving all of the records from the query() method.

Since we want to allow an efficient query for a cursor/watermark (more specifically the watermark for EventsQuery in this PR) we have to include a sorting property, that way we can have a specific key as a starting point in level, and know that everything coming after that key is sorted accordingly.

tests/store/index-level.spec.ts
LiranCohen: This is a good call out.

It *technically* shouldn't arrive at indexing the same record twice, as it should be met with a 409 error somewhere before it arrives here.

However maybe it's worth putting in some defensive programming in place for this and explicitly ignore.

Looking at it there is definitely a different in behavior between MessageStore and EventLog with the same data, and there is also a difference in behavior between the dwn-sql-store implementation (which just uses an incremented ID as a unique index in both cases).

@thehenrytsai mentioned that we might be able to just combine MessageStore and EventLog, which I want to think about today and see what the implications would be.

It makes sense in the dwn-sql-store as they are literally indexing the same things, the query is just slightly different.

src/store/index-level.ts
LiranCohen: Moving it up top would mean adding more complexity/less understood code where the iteratorOptions are built above it.

Since it's modifying the starting point of the iterator if a cursor exists, it makes more sense to leave it here.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
+   * @param value the value representing the data being indexed.
+   * @param indexes - (key-value pairs) to be included as part of indexing this item. Must include at least one indexing property.
+   * @param sortIndexes - (key-value pairs) to be used for sorting the index. Must include at least one sorting property.
+   * @param options IndexLevelOptions that include an AbortSignal.
    */
   async put(
     tenant: string,
-    dataId: string,
-    indexes: { [property: string]: unknown },
+    key: string,
```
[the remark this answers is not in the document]
LiranCohen: Yeah I can see how that might lead to confusion, `id` I think makes the most sense then.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
+      throw new Error('must include at least one indexable property');
+    }
 
-    const operations: LevelWrapperBatchOperation<string>[] = [ ];
+    const indexOps: LevelWrapperBatchOperation<string>[] = [];
+    // adding a reverse lookup to be able to delete index data as well as look up sorted indexes by a cursor
+    indexOps.push({ type: 'put', key: `__${key}__indexes`, value: JSON.stringify({ indexes, sortIndexes }) });
+
+    // for each indexable property we go through the different sorting indexes and construct a sorted index key.
+    // the sort property is the last property in the key before the tie-breaker.
```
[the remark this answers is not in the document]
LiranCohen: I'm not sure I understand the question.

However I will modify the comment, as I was reading the entire thing I realized that there is some confusion in the way that it is worded.

```suggestion
    // since LevelDB keys are sorted lexicographically, we need to create sortable keys for each indexable property.
    // for each indexable property we go through the different sorting indexes and construct a unique sorted index key.
    // the sort property is the last property in the key before the tie-breaker, the key itself is used as a tie-breaker.
```

src/store/index-level.ts
```diff
@@ -277,36 +505,24 @@ export class IndexLevel {
...
+    }
   }
 
   /**
    * Joins the given values using the `\x00` (\u0000) character.
    */
-  private static delimiter = `\x00`;
-  private join(...values: unknown[]): string {
+  protected static delimiter = `\x00`;
+  protected join(...values: unknown[]): string {
```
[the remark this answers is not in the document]
LiranCohen: Will fix, I originally had this as a base class.

src/store/index-level.ts
```diff
@@ -216,46 +287,203 @@ export class IndexLevel {
...
+    if (propertyValue !== undefined) {
+      keyConstruction.push(propertyValue);
+    }
+
+    if (sortValue !== undefined) {
+      keyConstruction.push(sortValue);
     }
+    if (key !== undefined) {
+      keyConstruction.push(key);
+    }
```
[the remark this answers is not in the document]
LiranCohen: Since you mentioned finding it confusing, I'll revert to just calling the `keySegmentJoin` when I'm building the prefixes where they are used. It will probably be more clear there.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
-   * Executes the given single filter query and appends the results without duplicate into `matchedIDs`.
-   */
-  private async executeSingleFilterQuery(tenant: string, filter: Filter, matchedIDs: Set<string>, options?: IndexLevelOptions): Promise<void> {
+  async query(tenant:string, queries: FilteredQuery[], options?: IndexLevelOptions): Promise<T[]> {
+    const matched:Map<string, T> = new Map();
+    await Promise.all(queries.map(query => this.executeSingleFilterQuery(tenant, query, matched, options)));
+    return [...matched.values()];
+  }
+
+  async executeSingleFilterQuery(tenant:string, query: FilteredQuery, matches: Map<string, T>, options?: IndexLevelOptions): Promise<void> {
```
thehenrytsai, line 142: Can we add light documentation on what `matches` is for code readers? Specifically on the `string` since `matches` acts as return type which is kinda important.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
     // We will find the union of these many individual queries later.
     for (const propertyName in filter) {
       const propertyFilter = filter[propertyName];
-
       if (typeof propertyFilter === 'object') {
         if (Array.isArray(propertyFilter)) {
-          // `propertyFilter` is a AnyOfFilter
+          // `propertyFilter` is a OneOfFilter
+          // if OneOfFilter, the cursor properties are a map of each individual EqualFilter and the associated cursor string
+          const cursorMap = propertyCursors[propertyName] as Map<EqualFilter, string>|undefined;
```
thehenrytsai, line 164: Sanity: is `undefined` necessary?

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
 
-import { executeUnlessAborted } from '../index.js';
 import { flatten } from '../utils/object.js';
-import { createLevelDatabase, LevelWrapper } from './level-wrapper.js';
+import { SortOrder } from '../types/message-types.js';
+
+
+type IndexedItem<T> = {
+  value: T;
+  key: string;
```
thehenrytsai, line 12: Assuming I understand correctly, please rename to `id`. Got confused the `key` being a index key.

src/store/index-level.ts
```diff
@@ -216,46 +287,203 @@ export class IndexLevel {
...
+      const lteCursor = cursor && this.extractValueFromKey(cursor) === this.encodeValue(rangeFilter.lte) ? cursor : undefined;
+      for (const item of await this.findExactMatches(tenant, propertyName, rangeFilter.lte, sortProperty, sortDirection, lteCursor, options)) {
+        lteMatches.push(item);
       }
     }
 
-    return matches;
+    // if the iterator is reversed and the results should be in Ascending order, we reverse the iterated matches.
+    // if the iterator is not reversed and the results should be in Descending order, we also reverse the iterated matches.
+    if ((iteratorOptions.reverse === true && sortDirection === SortOrder.Ascending) ||
```
thehenrytsai, line 325: Sanity: I am failing to understand the point of sorting here and in all other "match" methods. Don't they get put into `matches` later which is a `Map` thus all this work is throw-away?

src/store/index-level.ts
thehenrytsai: Yeah... the use of `cursor` itself is overloaded (thus confusing for newcomer), sometimes it means the item ID, sometimes it means something else (key value as you mentioned).

We need to rename them to differentiate somehow. But looking pass all that, I am still trying to see how "sort value" is been taken advantage of, 'coz that's gotta be where the magic happens, especially in range queries. I'll keep reading, I am getting it bit by bit.... SLOOOOOOOWLY.

thehenrytsai (approved):
🙏

thehenrytsai (changes requested):
Approved the wrong PR! Looking at multiple PR at the same time!

src/store/index-level.ts
thehenrytsai: #572 submitted for review.

src/store/index-level.ts
```diff
@@ -1,213 +1,284 @@
...
+    await Promise.all(queries.map(query => this.executeSingleFilterQuery(tenant, query, matched, options)));
+    return [...matched.values()];
   }
 
-  /**
-   * Executes the given single filter query and appends the results without duplicate into `matchedIDs`.
-   */
-  private async executeSingleFilterQuery(tenant: string, filter: Filter, matchedIDs: Set<string>, options?: IndexLevelOptions): Promise<void> {
+  async executeSingleFilterQuery(tenant:string, query: FilteredQuery, matches: Map<string, T>, options?: IndexLevelOptions): Promise<void> {
+    const { filter, sortProperty , sortDirection, cursor } = query;
```
thehenrytsai, line 143: I don't get how ts/js allows you declare `cursor` twice in the same method.

Anyways, submitted  PR #575 to remove the double meaning of cursor entirely.

src/handlers/permissions-grant.ts
LiranCohen: Comment isn't relevant after refactor.

LiranCohen:
@thehenrytsai This is ready for another look, there are still tests missing but I think it could use a review at this point in time.

As we discussed there are 2 strategies for querying:

1. A linear scan on the sorted property checking each message and returning when the requested limit is reached. If a cursor is passed, it's used as a starting point for the linear scan.

2. A filtered query that does a more scoped filter on each of the provided properties, this is not sorted and after the matches are all returned, it is then sorted and paginated according to the sort property and cursor/limit.

The "black box" function we talked about is `searchFilterSelector` within the `IndexLevel` class.
Currently this is untested (needs testing before this merges), and is only supporting equality and OneOf filters.
This function might be simplified for this initial push to merge, but this would be where improvements could be made along the way for filtering.

json-schemas/interface-methods/events-filter.json
thehenrytsai, line 1: For my understanding: what is the process/criteria/intent of picking the properties that goes into this filter?

src/types/index-types.ts
```diff
@@ -0,0 +1,41 @@
...
+export type RangeFilter = (GT | LT) & Partial<GT> & Partial<LT>;
+
+export type FilterValue = EqualFilter | OneOfFilter | RangeFilter;
+
+export type Filter = {
+  [property: string]: FilterValue;
+};
+
+export type Indexes = { [key: string]: unknown };
+export type IndexedItem<T> = { itemId: string, value: T, indexes: Indexes };
```
thehenrytsai, line 41: Nit: love the new grouping of types, but not seeing the need to pull these two types out from `index-level.ts`, consider:
1. leaving these two types with `index-leverl.ts`
2. renaming the file to `query-types.ts`

src/types/event-types.ts
```diff
@@ -1,8 +1,22 @@
...
+ * @param method optional array of methods to filter.
+ * @param interface optional array of interfaces to filter.
+ * @param author optional author to filter.
+ *
+ * @param watermark optional watermark of the last message received from this filter.
+ */
+export type EventsFilter = RecordsFilter & {
+  method?: string[];
+  interface?: string[];
+  author?: string;
```
thehenrytsai, line 17: The documentation should be above the actual properties for auto-complete/hint to work in IDEs.

```suggestion
  /** optional author to filter. */
  author?: string;
```

json-schemas/interface-methods/events-filter.json
LiranCohen: It's everything that `records-filters` has, and additionally filters for interface/method.

I would have liked to do schema inheritance, but couldn't figure out how to do that with the `additionalProperties` set to false. I think there might be a way but in a newer version of schema validation? I don't know enough about the intricacies of schema-validation world.

src/types/index-types.ts
LiranCohen: I was actually debating this exact thing as far as query-types vs index-types goes, but I'm fine with going with your suggestion and moving those two back into `index-level.ts`

src/types/event-types.ts
```diff
@@ -1,8 +1,22 @@
...
 import type { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
 
+/**
+ * @param method optional array of methods to filter.
+ * @param interface optional array of interfaces to filter.
+ * @param author optional author to filter.
+ *
+ * @param watermark optional watermark of the last message received from this filter.
+ */
+export type EventsFilter = RecordsFilter & {
```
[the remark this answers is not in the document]
LiranCohen: `EventsFilter` includes everything in `RecordsFilter` but in addition also has `interface` and `method` accepting OR filters.

I just looked and I'm missing the newer properties we added to record recently, will fix that.

src/types/event-log.ts
```diff
@@ -23,16 +24,25 @@ export interface EventLog {
...
    * retrieves all of a tenant's events that occurred after the watermark provided.
    * If no watermark is provided, all events for a given tenant will be returned.
    */
   getEvents(tenant: string, options?: GetEventsOptions): Promise<Array<Event>>
 
+  /**
+   * retrieves a filtered set of events that occurred after a the watermark provided.
+   * accepts multiple filters each with their own watermark.
+   *
+   * If no watermark is provided, all events for a given tenant and filter combo will be returned.
```
thehenrytsai, line 42: Sanity: I thought cursor can only be `messageCid`? Is it not the case here and watermark should be used instead?

LiranCohen: The cursor can only be `messageCid` for MessageStore, it's the `watermark` value for EventLog.

In IndexLevel, the cursor can be whatever is chosen as the `itemId`.

I have to think about it a bit more, but thinking about it now that this has been generalized more... I'm not sure that we need to return the watermark at all, it might be internal and the messageCid itself is the cursor for this as well.

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
-
+  /**
+   * Appends Events to the EventLog in the order they are appended using a ulid watermark.
+   * The indexes are used to query the event-log for specific events.
+   *
+   * @param tenant
+   * @param messageCid the messageCid of the message that is being appended to the log.
+   * @param indexes property and value indexes to use for querying potential events.
+   * @returns the ulid watermark generated during this operation.
+   */
```
thehenrytsai, line 57: Remove these comments as well as for methods below) or move to/consolidate with `EventLog`. Since you are implementing `EventLog`, there is no need to duplicate the documentation, code auto-completion will show the documentation from the interface directly, as it should be.

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
     const watermark = this.ulidFactory();
-
-    await watermarkLog.put(watermark, messageCid);
-    await cidLog.put(messageCid, watermark);
-
+    // A reverse lookup in order to delete by messageCid
+    // We are using the watermark as the key for this index.
+    // When deleting by messageCid we need to look up the watermark.
+    const cidLogIndex = await this.messageCidPartition(tenant);
+    await cidLogIndex.put(messageCid, watermark);
```
thehenrytsai, line 64: For my understanding, why can't the `messageCid` itself be in index as part of the `this.index.put(...)` call?

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
-
+    await cidLog.batch(cidOps);
+    await Promise.all(indexDeletePromises);
     return numEventsDeleted;
   }
+
+  /**
+   * Gets the LevelDB Sublevel for the messageCid reverse lookup.
+   */
+  private async messageCidPartition(tenant: string): Promise<LevelWrapper<string>> {
```
thehenrytsai, line 127: Please rename.

```suggestion
  private async getMessageCidToWatermarkTable(tenant: string): Promise<LevelWrapper<string>> {
```

Of if you think too long:

```suggestion
  private async getWatermarkTable(tenant: string): Promise<LevelWrapper<string>> {
```

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
-    const watermarkLog = await tenantEventLog.partition('watermarks');
-    await watermarkLog.batch(ops);
-
+    await cidLog.batch(cidOps);
+    await Promise.all(indexDeletePromises);
     return numEventsDeleted;
   }
+
+  /**
+   * Gets the LevelDB Sublevel for the messageCid reverse lookup.
```
thehenrytsai, line 125:
```suggestion
   * Gets the messageCid to watermark lookup table of the given tenant.
```

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
+  async append(tenant: string, messageCid: string, indexes: { [key:string]: unknown }): Promise<string> {
     const watermark = this.ulidFactory();
-
-    await watermarkLog.put(watermark, messageCid);
-    await cidLog.put(messageCid, watermark);
-
+    // A reverse lookup in order to delete by messageCid
+    // We are using the watermark as the key for this index.
+    // When deleting by messageCid we need to look up the watermark.
+    const cidLogIndex = await this.messageCidPartition(tenant);
```
thehenrytsai, line 63:
```suggestion
    const cidToWatermarkTable = await this.messageCidPartition(tenant);
```

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
+   * @returns the number of events deleted.
+   */
+  async deleteEventsByCid(tenant: string, messageCids: Array<string>): Promise<number> {
+    if (messageCids.length === 0) {
       return 0;
     }
 
-    const tenantEventLog = await this.db.partition(tenant);
-    const cidLog = await tenantEventLog.partition(CIDS_SUBLEVEL_NAME);
+    const cidLog = await this.messageCidPartition(tenant);
```
thehenrytsai, line 104:
```suggestion
    const cidToWatermarkTable = await this.messageCidPartition(tenant);
```

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
   createLevelDatabase?: typeof createLevelDatabase,
 };
 
-const WATERMARKS_SUBLEVEL_NAME = 'watermarks';
 const CIDS_SUBLEVEL_NAME = 'cids';
```
thehenrytsai, line 21: To reduce cognitive load, I think the DB name should be identified by the value it stores (watermarks) as opposed to the keys (cids).

```suggestion
const WATERMARK_SUBLEVEL_NAME = 'watermarks';
```

Also style consideration: consider just declaring this const within the method that uses it, no one else references/cares about it in this case.

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
-    await cidLog.put(messageCid, watermark);
-
+    // A reverse lookup in order to delete by messageCid
+    // We are using the watermark as the key for this index.
+    // When deleting by messageCid we need to look up the watermark.
+    const cidLogIndex = await this.messageCidPartition(tenant);
+    await cidLogIndex.put(messageCid, watermark);
+
+    const event:Event = { messageCid, watermark };
+    await this.index.put(tenant, watermark, event, { ...indexes, watermark });
```
thehenrytsai, line 67: Design: I am not 100% convinced that the motivation behind changing the `index.put(...)` method to incorporate "value" as input is really needed (though I have yet to go through the entire PR), especially given that we already have a CID -> watermark table, so seems like the same pattern used by the message store can be followed.

We should probably discuss this one. To be clear, the potential change here isn't much, a simplification even.

src/event-log/event-log-level.ts
```diff
@@ -15,101 +18,114 @@ type EventLogLevelConfig = {
...
-      events.push(event);
-    }
+  /**
+   * Queries Events for a given tenant using the filters provided.
+   *
+   * @param tenant
+   * @param filters an array of filters that designates which event properties are being queried.
+   * @returns an array of matching Events without duplicate entries between the filters.
+   */
+  async queryEvents(tenant: string, filters: Filter[], watermark?: string): Promise<Event[]> {
```
thehenrytsai, line 78: Sanity: no paging? Same question to `getEvents()`

json-schemas/interface-methods/events-filter.json
thehenrytsai: It's not everything records filter has though right? Like, a bunch date properties, `dataSize`, `published` are not there, hence the question. Or am I misreading something?

thehenrytsai: I see your comments below. Will wait for the update.

src/types/event-log.ts
thehenrytsai: This is great, I had a comment exactly on this later in the PR. Seems we are arriving at the same conclusion that `index.put()` signature might not need to be modified. Happy to jump on a call to confirm each other's understanding if needed.

src/event-log/event-log-level.ts
LiranCohen: Will add this in a separate PR along with ProtocolQuery pagination (that would be all the queries).

src/event-log/event-log-level.ts
LiranCohen: This is done and simplified!

src/event-log/event-log-level.ts
LiranCohen: No longer relevant.

thehenrytsai (commented):
Stopped at `FilterSelector`, since it's passed 1AM!

src/utils/records.ts
```diff
@@ -263,6 +265,11 @@ export class Records {
       filterCopy.datePublished = datePublishedFilter;
     }
 
+    //if sorting by published, results must filter for published
+    if (dateSort === DateSort.PublishedAscending || dateSort === DateSort.PublishedDescending) {
+      filterCopy.published = true;
+    }
```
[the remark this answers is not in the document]
LiranCohen: There's a check for something similar but not exactly. It's a check to make sure the user doesn't set a `datePublisehd` while also having  `publisehd:false` as that would return no results. But that is self contained within the filter schema.

The filter and pagination properties are different schema objects, so there isn't any dependency link there (as it's structured currently).

src/interfaces/events-query.ts
```diff
@@ -0,0 +1,76 @@
...
+    return new EventsQuery(message);
+  }
+
+  public static async create(options: EventsQueryOptions): Promise<EventsQuery> {
+    const descriptor: EventsQueryDescriptor = {
+      interface        : DwnInterfaceName.Events,
+      method           : DwnMethodName.Query,
+      filters          : this.normalizeFilters(options.filters),
+      messageTimestamp : options.messageTimestamp ?? Time.getCurrentTimestamp(),
+      watermark        : options.watermark,
```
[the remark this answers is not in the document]
LiranCohen: Yeah this should be named cursor... I can do that now as a part of this PR.

Subsequent PR will add pagination, as well as probably migrate `Query` to the `Messages` interface, getting rid of `EventsGet` completely (it's just a query without filters at that point).

src/types/event-types.ts
```diff
@@ -1,8 +1,17 @@
...
 import type { GenericMessageReply } from '../core/message-reply.js';
+import type { RecordsFilter } from './records-types.js';
 import type { AuthorizationModel, GenericMessage } from './message-types.js';
 import type { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
 
+export type EventsFilter = RecordsFilter & {
+  /** optional array of methods to filter */
+  method?: string[];
+  /** optional array of interfaces to filter */
+  interface?: string[];
```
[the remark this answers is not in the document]
LiranCohen: It can be used as a OneOf filter. We can discuss limiting the types of filters allowed by this interface (and it would be the same for the Subscription interface).

Original thought is someone could want to selectively filter for `Records` and `Permissions` interfaces by a particular `protocol` as an example.

src/types/event-log.ts
```diff
@@ -23,21 +19,35 @@ export interface EventLog {
...
-   * @returns {Promise<string>} watermark
+   * @param indexes - (key-value pairs) to be included as part of indexing this event.
+   */
+  append(tenant: string, messageCid: string, indexes: { [key:string]: string|boolean }): Promise<void>
+
+  /**
+   * Retrieves all of a tenant's events that occurred after the cursor provided.
+   * If no cursor is provided, all events for a given tenant will be returned.
+   *
+   * The cursor is a messageCid.
```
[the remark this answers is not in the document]
LiranCohen: I really want to get rid of this method completely in favor of Query, I can do that as a part of this PR or I think it would be better suited as a part of a separate PR to move `EventsQuery` to the `Messages` interface.

src/utils/filter.ts
```diff
@@ -0,0 +1,341 @@
...
+      const filterValue = filter[filterProperty];
+      const indexValue = indexedValues[filterProperty];
+      if (indexValue === undefined) {
+        return false;
+      }
+
+      if (typeof filterValue === 'object') {
+        if (Array.isArray(filterValue)) {
+          // `propertyFilter` is a OneOfFilter
+          // if OneOfFilter, the cursor properties are a map of each individual EqualFilter and the associated cursor string
```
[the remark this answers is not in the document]
LiranCohen: Sorry, this was copy-pasta leftover updated comments to remove reference to cursor.

src/utils/filter.ts
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  /**
+   * Encodes an indexed value to a string
+   *
+   * NOTE: we currently only use this for strings, numbers and booleans.
+   * Objects are returned as "[object Object]".
+   * Although this never happens maybe we should consider making this function, and those which call it, typed better.
+   */
+  static encodeValue(value: unknown): string {
```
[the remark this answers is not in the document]
LiranCohen: https://github.com/TBD54566975/dwn-sdk-js/pull/621/files

I addressed this and the discussion around it in this PR. Easier to go over and discuss.

src/utils/records.ts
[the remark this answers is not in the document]
thehenrytsai: > It's a check to make sure the user doesn't set a datePublisehd while also having publisehd:false

Would it make sense to do this check in the same place? Since it feels like an integrity check, it feels strange to be part of "convertFilter".

src/interfaces/events-query.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      return normalizedFilter;
+    });
+  }
+
+  /**
+ *  Converts an incoming EventsFilter into a Filter usable by EventLog.
+ *
+ * @param filters An EventsFilter
+ * @returns {EventsLogFilter} a generic Filter able to be used with EventLog query.
```
thehenrytsai: `EventsLogFilter` doesn't exist. The documentation seems to be wrong, an array is returned.

```suggestion
 * @returns {Filter} a generic Filter able to be used with EventLog query.
```

src/interfaces/events-query.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (filter.schema !== undefined) {
+        normalizedFilter.schema = normalizeSchemaUrl(filter.schema);
+      }
+
+      return normalizedFilter;
+    });
+  }
+
+  /**
+ *  Converts an incoming EventsFilter into a Filter usable by EventLog.
```
thehenrytsai: Update documentation, array is both input and out put. Also nit, documentation indentation is wrong here.

src/types/event-log.ts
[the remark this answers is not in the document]
thehenrytsai: Supportive, let's then just rename `gt` to `cursor` to get documentation consistency going for now.

src/store/index-level.ts
```diff
@@ -1,213 +1,393 @@
...
   }
 
   async close(): Promise<void> {
-    return this.db.close();
+    await this.db.close();
+  }
+
+  /**
+ * deletes everything in the underlying index db.
+ */
```
thehenrytsai, line 54: Nit: incorrect indentation, might be worthwhile checking if there is away to just have linter detect/fix this.

src/store/index-level.ts
```diff
@@ -1,213 +1,393 @@
...
-      // example keys (\u0000 is just shown for illustration purpose because it is the delimiter used to join the string segments below):
-      // 'interface\u0000"Records"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
-      // 'method\u0000"Write"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
-      // 'schema\u0000"http://ud4kyzon6ugxn64boz7v"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
-      // 'dataCid\u0000"bafkreic3ie3cxsblp46vn3ofumdnwiqqk4d5ah7uqgpcn6xps4skfvagze"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
-      // 'dateCreated\u0000"2023-05-25T18:23:29.425008Z"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
-      const key = this.join(propertyName, this.encodeValue(propertyValue), dataId);
-      operations.push({ type: 'put', key, value: dataId });
+      const key = IndexLevel.keySegmentJoin(IndexLevel.encodeValue(sortValue), itemId);
+      const item: IndexedItem = { itemId, indexes };
```
thehenrytsai, line 92: Guess I will find out as I read on why indexes  need to be included for every index write/push. Immediate reaction assuming I am reading it correctly is that this seems rather heavy. e.g. for every `RecordsWrite`, we could be looking at close to 20 key-value pairs stored repetitively ~20 times.

src/store/index-level.ts
```diff
@@ -1,213 +1,393 @@
...
-      const propertyValue = indexes[propertyName];
+    const tenantPartition = await this.db.partition(tenant);
+    const indexOps: LevelWrapperBatchOperation<string>[] = [];
 
-      // NOTE: appending data ID after (property + value) serves two purposes:
-      // 1. creates a unique entry of the property-value pair per data/object
-      // 2. when we need to delete all indexes of a given data ID (`delete()`), we can reconstruct the index keys and remove the indexes efficiently
+    // create an index entry for each property
+    // these indexes are all sortable lexicographically.
+    for (const sortProperty in indexes) {
```
thehenrytsai, line 85: Nit: can we rename `sortProperty` -> `indexName`, and `sortValue` -> `indexValue`? 'coz the word "sort" is implied for an index.

src/store/level-wrapper.ts
```diff
@@ -205,6 +205,16 @@ export class LevelWrapper<V> {
     return executeUnlessAborted(this.db.batch(operations), options?.signal);
   }
 
+  /**
+   * @returns a LevelWrapperBatchOperation for the specified partition name
+   */
+  partitionOperation(name: string, operation: LevelWrapperBatchOperation<V>): LevelWrapperBatchOperation<V> {
```
thehenrytsai, line 211: Nit: start methods with a verb:

```suggestion
  createPartitionOperation(partitionName: string, operation: LevelWrapperBatchOperation<V>): LevelWrapperBatchOperation<V> {
```

LiranCohen:
Closing this PR in favor of a newer one.

PR #560, opened 2023-10-14
Use protocolRoles to authorize RecordsQuery

diehuxx:
Add the role-authorized `query` action to protocol definitions


src/handlers/records-query.ts
```diff
@@ -38,6 +39,11 @@ export class RecordsQueryHandler implements MethodHandler {
       // authentication
       try {
         await authenticate(message.authorization!, this.didResolver);
+
+        // Only run protocol authz if message deliberately invokes it
+        if (RecordsQueryHandler.shouldProtocolAuthorizeQuery(recordsQuery)) {
+          await ProtocolAuthorization.authorizeQuery(tenant, recordsQuery, this.messageStore);
```
diehuxx, line 45: To preserve existing functionality, we only run authz and throw errors if the message invokes a `protocolRole`. If the message does not invoke a `protocolRole`, we simply skip querying of protocol records.

src/core/dwn-error.ts
```diff
@@ -41,16 +41,20 @@ export enum DwnErrorCode {
   ProtocolAuthorizationIncorrectProtocolPath = 'ProtocolAuthorizationIncorrectProtocolPath',
   ProtocolAuthorizationInvalidSchema = 'ProtocolAuthorizationInvalidSchema',
   ProtocolAuthorizationInvalidType = 'ProtocolAuthorizationInvalidType',
+  ProtocolAuthorizationMissingContextId = 'ProtocolAuthorizationMissingContextId',
   ProtocolAuthorizationMissingRole = 'ProtocolAuthorizationMissingRole',
   ProtocolAuthorizationMissingRuleSet = 'ProtocolAuthorizationMissingRuleSet',
   ProtocolAuthorizationParentlessIncorrectProtocolPath = 'ProtocolAuthorizationParentlessIncorrectProtocolPath',
   ProtocolAuthorizationNotARole = 'ProtocolAuthorizationNotARole',
+  ProtocolAuthorizationQueryFilterMissingRequiredProperties = 'ProtocolAuthorizationQueryFilterMissingProtocolPath',
```
thehenrytsai, line 49:
```suggestion
  ProtocolAuthorizationQueryFilterMissingRequiredProperties = 'ProtocolAuthorizationQueryFilterMissingRequiredProperties',
```

src/handlers/records-query.ts
```diff
@@ -165,6 +175,19 @@ export class RecordsQueryHandler implements MethodHandler {
     };
   }
 
+  /**
+   * Creates a filter for unpublished records that are intended for the query author (where `recipient` is the author).
```
thehenrytsai, line 179: Need to change.

src/handlers/records-query.ts
thehenrytsai: > we simply skip querying of protocol records.

Sanity: Not sure if I understand what this means exactly. You don't mean we "explicitly exclude" protocol records if message doesn't invoke `protocolRole` right?

tests/interfaces/protocols-configure.spec.ts
```diff
@@ -314,6 +314,34 @@ describe('ProtocolsConfigure', () => {
           .to.be.rejectedWith(DwnErrorCode.ProtocolsConfigureInvalidActionMissingOf);
       });
 
+      it('rejects protocol definitions with `can: query` in non-role rules', async () => {
```
thehenrytsai, line 317: Sanity: is this just a temporary measure, or we actually think that query only makes sense for protocol actors defined by roles?

src/core/protocol-authorization.ts
```diff
@@ -109,16 +112,63 @@ export class ProtocolAuthorization {
...
+    messageStore: MessageStore,
+  ): Promise<void> {
+    // validate that required properties exist in query filter
+    const { protocol, protocolPath, contextId } = incomingMessage.message.descriptor.filter;
+    if (protocolPath === undefined) {
+      throw new DwnError(
+        DwnErrorCode.ProtocolAuthorizationQueryFilterMissingRequiredProperties,
+        'Role-authorized queries must include `protocolPath` in the filter'
+      );
+    }
```
thehenrytsai, line 131: Should we front load this check which seems to be structural to `Records.Query.parse()`?

src/handlers/records-query.ts
```diff
@@ -38,6 +39,11 @@ export class RecordsQueryHandler implements MethodHandler {
       // authentication
```
thehenrytsai, line 39: Nit: 
```suggestion
      // authentication & authorization
```

src/core/protocol-authorization.ts
```diff
@@ -109,16 +112,63 @@ export class ProtocolAuthorization {
...
+      contextId,
+      protocolDefinition,
+      messageStore,
+    );
+
+    // verify method invoked against the allowed actions
+    await ProtocolAuthorization.verifyAllowedActions(
+      incomingMessage,
+      inboundMessageRuleSet,
+      [], // ancestor chain is not relevant to queries
```
thehenrytsai, line 160: Related to the other comment regarding if we will ever allow query from path-based actors:

Especially if that's still in consideration and assuming you are in agreement regarding moving out the above structure check, then it seems that it would be a trade off regarding efficiency (creating a new method) vs code reuse (using existing `authorize()`)?

I am not necessarily advocating one over the other, but wanted to first make sure this is correct understanding.

src/core/protocol-authorization.ts
```diff
@@ -315,14 +364,22 @@ export class ProtocolAuthorization {
...
       interface         : DwnInterfaceName.Records,
       method            : DwnMethodName.Write,
-      protocol          : recordsWrite.message.descriptor.protocol!,
+      protocol          : protocolUri,
       protocolPath      : protocolRole,
       recipient         : incomingMessage.author!,
       isLatestBaseState : true,
     };
+
     if (roleRuleSet?.$contextRole) {
```
thehenrytsai, line 373: Gentle reminder there are still a few places with annoying uncovered branch coverage.

tests/interfaces/protocols-configure.spec.ts
diehuxx: This is a temporary measure

src/core/protocol-authorization.ts
diehuxx: @thehenrytsai and I discussed a bit offline. We decided that the code duplication in `authorizeQuery` and `authorize` is acceptable because combining them into one function is harder to read/reason about. In fact, we'll take it a step further by separating `authorize` into `authorizeRead` and `authorizeWrite`.

src/core/protocol-authorization.ts
diehuxx:
```
protocol-authorization.ts      |   99.36 |    95.14 |     100 |   99.36 | 272-273,609-610
```
I don't see this line in the uncovered lines when I run `npm run test:node`. Lines `272-273,609-610` (in commit [49d0cdd](https://github.com/TBD54566975/dwn-sdk-js/pull/560/commits/49d0cdd78de9e281054effed3ee9eda488216645)) have comments explaining why they are uncovered. Are you getting different test coverage results than me?

src/core/protocol-authorization.ts
diehuxx: I'm on the fence about this one. This check is only run if `recordsQuery.authorizationPayload.protocolRole` is present, and we haven't been validating `authorization` payload fields within `parse()` methods. I'd prefer to keep the check here for now since it's related to `authorization`.

src/handlers/records-query.ts
diehuxx: Updated

src/handlers/records-query.ts
diehuxx: Updated

src/core/protocol-authorization.ts
thehenrytsai: https://app.codecov.io/github/TBD54566975/dwn-sdk-js/blob/main/src%2Fcore%2Fprotocol-authorization.ts shows what I am seeing.

tests/interfaces/protocols-configure.spec.ts
thehenrytsai: Got it!

src/handlers/records-query.ts
diehuxx: We don't exclude protocol records when `protocolRole` is absent. It's that `protocolRole` must be present for the request to get records beyond published records, records they authored, and records they received.

src/core/protocol-authorization.ts
diehuxx: Pushed an additional test that should catch this.

thehenrytsai: 🙏

src/core/protocol-authorization.ts
thehenrytsai: We do validate `authorization` payload via `validateMessageSignatureIntegrity()` which is called in `parse()` right? 

I do agree it is debatable where the best place is, though given the pattern of structural integrity check originally using JSON schema and later via code, I do fee `validateMessageSignatureIntegrity()` or maybe a new method called within `parse()` is worth considering.

We can punt this but keep track the consideration by adding a comment such as :
```ts
// consider moving this check to structural check to `parse()` or `validateMessageSignatureIntegrity()`
```

README.md
```diff
@@ -48,7 +48,13 @@ Here's to a thrilling Hacktoberfest voyage with us! 🎉
 # Decentralized Web Node (DWN) SDK <!-- omit in toc -->
 
 Code Coverage
+<<<<<<< HEAD
```
thehenrytsai, line 51: Merge conflict slipped through

src/handlers/records-query.ts
```diff
@@ -179,4 +203,11 @@ export class RecordsQueryHandler implements MethodHandler {
       published         : false
     };
   }
+
+  /**
+   * Determines if ProtocolAuthorization.authorizeQuery should be run and if the corresponding filter should be used.
+   */
+  private static shouldProtocolAuthorizeQuery(recordsQuery: RecordsQuery): boolean {
+    return recordsQuery.message.descriptor.filter.protocol !== undefined && recordsQuery.authorSignaturePayload?.protocolRole !== undefined;
```
thehenrytsai, line 211: Nit: Missed a F'ing annoy branch coverage here... on `authorSignaturePayload?.protocolRole`

README.md
diehuxx: ope, thanks

src/core/protocol-authorization.ts
thehenrytsai: 🙏

README.md
```diff
@@ -48,7 +48,7 @@ Here's to a thrilling Hacktoberfest voyage with us! 🎉
 # Decentralized Web Node (DWN) SDK <!-- omit in toc -->
 
 Code Coverage
-![Statements](https://img.shields.io/badge/statements-98.48%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-95.56%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-95.71%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-98.48%25-brightgreen.svg?style=flat)
+![Statements](https://img.shields.io/badge/statements-98.5%25-brightgreen.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-95.82%25-brightgreen.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-95.76%25-brightgreen.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-98.5%25-brightgreen.svg?style=flat)
```
thehenrytsai, line 51: Sanity: where did the `codecov` badge go?

We should just use that one going forward.

diehuxx: Updated -- I accidentally switched it back to the old badges in another PR

thehenrytsai (approved):
🙏

Issue #136, opened 2022-11-04
Increase code coverage (even if just by a little bit)


thehenrytsai:
### Background:
The code coverage goal is 100% for this repo, so any PR adding any test to increase the overall code coverage would be super useful!

### Task Details:
To know what code hasn't been covered:
1. Run `npm run test:node`
2. Ignore the code coverage summary at the end of the run, there is a better way:
3. Use MacOS Finder or Windows Explorer to open the `coverage/index.html` page, it will show you precisely what line of code is not covered

Files that can be ignored as they will likely be undergoing drastic changes:
- `permissions-request.ts`

## Picking Up This Issue:
- If you'd like to work on this, please comment "picking this up" below, and I'll assign the issue to you 

## Questions:
- If you need assistance or clarification, feel free to comment below.
- Feeling overwhelmed? Join our [`#hack-together` channel](https://discord.gg/tbd) and collaborate with a buddy.
- New to open-source? Check out our [Hacktober `#getting-started` channel](https://discord.gg/tbd).
- [Leaderboard](https://github.com/TBD54566975/developer.tbd.website/issues/721): Check your ranking after finishing this task.

### Resources:
- **Creating a Pull Request**: If you're new to GitHub and unsure how to create a pull request, follow this [step-by-step guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Remember, communication is key! If you have any questions or face any challenges, we're here to help so please don't hesitate to reach out.

Good Luck! 🍁

tux7P:
picking this up

EbonyLouis:
@tux7P Yay! I'll assign it to you now.

 I'd also like to note that Hacktoberfest officially kicks off on **October 1st**. Contributions made before the event may not be eligible for the challenge. **However, feel free to start working on it, but I'd recommend holding off on submitting your PR until the event begins.** 

> to be eligible for Hacktoberfest's official prizes make sure you're registered, registration [is now open.](https://hacktoberfest.com/auth/) Reach out if you need any help 🥳 

Good Luck 🍁

EbonyLouis:
More than one person can work on this issue, feel free to comment "picking this up" if it interest you! 🍁

shobitb:
@EbonyLouis, picking this up!

EbonyLouis:
Heyy @tux7P how are things going with this issue? Is this something you still want to work on?

[...]

PR #561, opened 2023-10-16
Add tests for error cases in did-resolver.spec.ts

shobitb:
This commit adds new tests in did-resolver.spec.ts to cover some of the previously uncovered error-related branching paths.

This commit contributes partly to issue #136.

Based on the output of `npm run test:node` before vs. after this commit, the `% Branch` is now increased by 9.66 percentage points, and the `% Stmt` and `% Lines` by 3.29 percentage points.

Before:
```
did-resolver.ts |   87.91 |    78.57 |   66.66 |   87.91
```

After:
```
did-resolver.ts |    91.2 |    88.23 |   66.66 |    91.2 |
```


thehenrytsai (commented):
@shobitb thanks for helping us increase CC! Could you opportunistically convert the `Error` used in DidResolver into `DwnError` and introduce new error codes?  Search for `DwnError` and you'll find pattern used in code and tests.

shobitb:
> @shobitb thanks for helping us increase CC! Could you opportunistically convert the `Error` used in DidResolver into `DwnError` and introduce new error codes? Search for `DwnError` and you'll find pattern used in code and tests.

Yes, most def. Let me take a look and will send an updated PR.

src/did/did.ts
```diff
@@ -16,13 +18,13 @@ export class Did {
    */
   public static validate(did: unknown): void {
     if (typeof did !== 'string') {
-      throw new Error(`DID is not string: ${did}`);
+      throw new DwnError(DwnErrorCode.DidNotValid, `DID is not string: ${did}`);
```
thehenrytsai, line 21: Nit: As a coding guideline, we never reuse the same error code. So introduce a new one, perhaps:

```suggestion
      throw new DwnError(DwnErrorCode.DidNotString, `DID is not string: ${did}`);
```

thehenrytsai (commented):
Last nit then good to go!

shobitb:
> Last nit then good to go!

Gotcha!  I originally thought the codes were like HTTP codes, where a generic code could be used with different specific error messages.

Added a new code for `DidNotString` in the latest commit

thehenrytsai (approved):
Good to merge! Thanks again!!

PR #569, opened 2023-10-20
Add protocol update action

diehuxx:
The `update` action allows an actor to update an existing record. This contrasts with the `write` action which allows only the creator of a record to update it. In other words:
1. `write` allows a DID to create a record and update the record they have created
2. `update` allows a DID to update a record, regardless of initial author

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -328,7 +328,10 @@ export class ProtocolAuthorization {
     }
   }
 
-  private static async getInboundMessageActions(
+  /**
+   * Returns a list of ProtocolAction(s) based on the incoming message, one of which must be allowed for the message to be authorized.
+   */
+  private static async getActionsSeekingARuleMatch(
     tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
     messageStore: MessageStore,
@@ -342,7 +345,7 @@ export class ProtocolAuthorization {
     if (await incomingRecordsWrite.isInitialWrite()) {
       // only 'write' allows initial RecordsWrites; 'update' only applies to subsequent RecordsWrites
       return [ProtocolAction.Write];
-    } else if (await incomingRecordsWrite.authoredByInitialRecordAuthor(tenant, messageStore)) {
+    } else if (await incomingRecordsWrite.isAuthoredByInitialRecordAuthor(tenant, messageStore)) {
       // Both 'update' and 'write' authorize the incoming message
       return [ProtocolAction.Write, ProtocolAction.Update];
     } else {
@@ -363,7 +366,7 @@ export class ProtocolAuthorization {
     messageStore: MessageStore,
   ): Promise<void> {
     const incomingMessageMethod = incomingMessage.message.descriptor.method;
-    const inboundMessageActions = await ProtocolAuthorization.getInboundMessageActions(tenant, incomingMessage, messageStore);
+    const inboundMessageActions = await ProtocolAuthorization.getActionsSeekingARuleMatch(tenant, incomingMessage, messageStore);
     const author = incomingMessage.author;
     const actionRules = inboundMessageRuleSet.$actions;
 
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -6,9 +6,8 @@ import type { ProtocolActionRule, ProtocolDefinition, ProtocolRuleSet, Protocols
 
 import { RecordsWrite } from '../interfaces/records-write.js';
 import { DwnError, DwnErrorCode } from './dwn-error.js';
-import { DwnInterfaceName, DwnMethodName, Message } from './message.js';
+import { DwnInterfaceName, DwnMethodName } from './message.js';
 import { ProtocolAction, ProtocolActor } from '../types/protocols-types.js';
-import type { RecordsDelete } from '../index.js';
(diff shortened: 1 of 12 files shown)


alanhkarp:
Is this distinction necessary?  Can't you just use capabilities for this?

Say that Alice creates some record.  As I understand it, the DWN node can't create a capability to return to her, but it does know she is the owner.  That means Alice can create a "write" capability that she can use.  She can also delegate that capability to someone else.

The upshot is that you only need one "write" method at the expense of the owner having to create a root capability.  An added advantage is that Alice won't be vulnerable to a confused deputy attack as she would be if she used her ambient authority.

tests/vectors/protocol-definitions/author-update.json
```diff
@@ -0,0 +1,27 @@
+{
+  "protocol": "http://recipient-update-protocol.xyz",
```
thehenrytsai, line 2: Super nit:
```suggestion
  "protocol": "http://author-update-protocol.xyz",
```
Also the protocol is a bit weird in that it only allows the owner to comment, but fair enough given this is all superficial anyways.

src/core/protocol-authorization.ts
```diff
@@ -333,17 +328,42 @@ export class ProtocolAuthorization {
...
   private static async verifyAllowedActions(
+    tenant: string,
     incomingMessage: RecordsRead | RecordsWrite,
     inboundMessageRuleSet: ProtocolRuleSet,
     ancestorMessageChain: RecordsWriteMessage[],
+    messageStore: MessageStore,
   ): Promise<void> {
     const incomingMessageMethod = incomingMessage.message.descriptor.method;
-    const inboundMessageAction = methodToAllowedActionMap[incomingMessageMethod];
+    const inboundMessageActions = await ProtocolAuthorization.getInboundMessageActions(tenant, incomingMessage, messageStore);
```
thehenrytsai, line 366: Naming is a bit vague to me, consider renaming to perhaps `getActionsSeekingARuleMatch()` or `getPotentiallyAllowedActions()` or `getActionCandidates()`

```suggestion
    const actionsSeekingARuleMatch = await ProtocolAuthorization.getActionsSeekingARuleMatch(tenant, incomingMessage, messageStore);
```

thehenrytsai (dismissed):
LGTM and test coverage is super thorough!! Just a few superficial comment.

diehuxx:
> Is this distinction necessary? Can't you just use capabilities for this?
> 
> Say that Alice creates some record. As I understand it, the DWN node can't create a capability to return to her, but it does know she is the owner. That means Alice can create a "write" capability that she can use. She can also delegate that capability to someone else.
> 
> The upshot is that you only need one "write" method at the expense of the owner having to create a root capability. An added advantage is that Alice won't be vulnerable to a confused deputy attack as she would be if she used her ambient authority.

@csuwildcat Any thoughts?

src/interfaces/records-write.ts
```diff
@@ -602,6 +604,17 @@ export class RecordsWrite {
     return (entryId === this.message.recordId);
   }
 
+  public async authoredByInitialRecordAuthor(tenant: string, messageStore: MessageStore): Promise<boolean> {
```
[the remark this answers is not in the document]
diehuxx: Added summary doc and updated method name

tests/vectors/protocol-definitions/author-update.json
diehuxx: Updated string. You're right, it's not the most realistic example. This is mostly a test fixture than an insightful example to read.

src/core/protocol-authorization.ts

LiranCohen:
> That means Alice can create a "write" capability that she can use. She can also delegate that capability to someone else.

Yes, you could technically build the same type of behavior with delegation for other aspects of the protocol rules, but it would be difficult to enforce these as a 'protocol' with other implementations.

Say you have a protocol with the following rules:
1) anyone can create a `post`.
2) anyone can add a `post/tag` to existing posts.
3) the `did` who created the `post` can edit any `post/tag` relating to the `post` they created.

Instead of each implementation of this social protocol front-end app explicitly delegating the write capability of the `post/tag` they created to the author of the `post`, the protocol rules handle this.

LiranCohen (approved):
LGTM!

alanhkarp:
@LiranCohen 

> @alanhkarp
> 
> > That means Alice can create a "write" capability that she can use. She can also delegate that capability to someone else.
> 
> Yes, you could technically build the same type of behavior with delegation for other aspects of the protocol rules, but it would be difficult to enforce these as a 'protocol' with other implementations.
> 
> Say you have a protocol with the following rules:
> 
> 1. anyone can create a `post`.
> 2. anyone can add a `post/tag` to existing posts.

In that case, allow the creator of the `post` to specify that no capability is needed to add a `post/tag` to the post. 

> 3. the `did` who created the `post` can edit any `post/tag` relating to the `post` they created.
> 

Using the `did` of the creator of the `post` is effectively using an access control list.  (The same goes for using Roles as I pointed out in another issue.)  It's not as bad a a pure ACL system, since Alice can create a capability when she wants to delegate some of her rights.  The risk, though, is a confused deputy vulnerability.

csuwildcat:
@nidhinahar looks good to me

thehenrytsai (approved):

PR #574, opened 2023-10-24
#483 Fix level wrapper close implementation

diehuxx:
follow up to #501


thehenrytsai (approved):
🙏 , though coverage went down...

PR #575, opened 2023-10-25
Disambiguating meaning of cursor

thehenrytsai:
Currently `cursor` can be the message CID, or be the ugly-looking string that represents an index key. This creates confusion when for newbies like me, especially there are literally variables named `cursor` in the SAME function that mean different things.


src/store/index-level.ts
```diff
@@ -314,8 +326,9 @@ export class IndexLevel<T> {
       // thus would not be included in the iterator even though we'd like it to be.
       //
       // we also only include the cursor ONLY if it is relevant to the exact property in the 'lte' filter.
```
diehuxx, line 328:
```suggestion
      // we also only include the index key ONLY if it is relevant to the exact property in the 'lte' filter.
```

Issue #557, opened 2023-10-13
Support delete in protocols


thehenrytsai:

thehenrytsai:
This is done.

Issue #387, opened 2023-05-31
Replace all `new Error(...)` with `new DwnError(...)`


thehenrytsai:
### Background:
currently we have `new DwnError(...)` and `new Error(...)` throughout the code

### Objective:
- Replace all `new Error(...)` with `new DwnError(...)`
- List of DwnError codes can be found here: `src/core/dwn-error.ts`

## Picking Up This Issue:
- If you'd like to work on this, please comment "picking this up" below, and I'll assign the issue to you 

## Questions:
- If you need assistance or clarification, feel free to comment below.
- Feeling overwhelmed? Join our [`#hack-together` channel](https://discord.gg/tbd) and collaborate with a buddy.
- New to open-source? Check out our [Hacktober `#getting-started` channel](https://discord.gg/tbd).
- [Leaderboard](https://github.com/TBD54566975/developer.tbd.website/issues/721): Check your ranking after finishing this task.

### Resources:
- **Creating a Pull Request**: If you're new to GitHub and unsure how to create a pull request, follow this [step-by-step guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Remember, communication is key! If you have any questions or face any challenges, we're here to help so please don't hesitate to reach out.

Good Luck! 🍁

tux7P:
picking this up🙂

EbonyLouis:
@tux7P Yay! I'll assign it to you now.

I'd also like to note that Hacktoberfest officially kicks off on **October 1st**. Contributions made before the event may not be eligible for the challenge. **However, feel free to start working on it, but I'd recommend holding off on submitting your PR until the event begins.** 

> to be eligible for Hacktoberfest's official prizes make sure you're registered, registration [is now open.](https://hacktoberfest.com/auth/) Reach out if you need any help 🥳

FroostyCodes:
Hi @EbonyLouis @thehenrytsai 
If this is not being worked on, I would like to work on it.

EbonyLouis:
@tux7P Hey do you still want to work on this issue?

EbonyLouis:
@FroostyCodes it's all yours, assigning it to you now, thanks! 🥳

Tushar98644:
@EbonyLouis can i work on this if the current assignee is inactive ? Thanks

flothjl:
[...]

PR #576, opened 2023-10-25
Support protocol authorized RecordsDelete

diehuxx:
Addresses #557. Support protocol authorized deletes for both actors (`author`, `recipient`, `anyone`) and roles (`$globalRole`, `$contextRole`).

### DIFF
--- a/src/interfaces/records-delete.ts
+++ b/src/interfaces/records-delete.ts
@@ -8,6 +8,7 @@ import { Message } from '../core/message.js';
 
 import { ProtocolAuthorization } from '../core/protocol-authorization.js';
 import { validateMessageSignatureIntegrity } from '../core/auth.js';
+import { DwnError, DwnErrorCode } from '../index.js';
 import { DwnInterfaceName, DwnMethodName } from '../core/message.js';
 
 export type RecordsDeleteOptions = {
@@ -60,7 +61,10 @@ export class RecordsDelete extends Message<RecordsDeleteMessage> {
     } else if (newestRecordsWrite.message.descriptor.protocol !== undefined) {
       await ProtocolAuthorization.authorizeDelete(tenant, this, newestRecordsWrite, messageStore);
     } else {
-      throw new Error('message failed authorization');
+      throw new DwnError(
+        DwnErrorCode.RecordsDeleteAuthorizationFailed,
+        'RecordsDelete message failed authorization'
+      );
     }
   }
 }
--- a/src/interfaces/records-delete.ts
+++ b/src/interfaces/records-delete.ts
@@ -1,15 +1,19 @@
+import type { MessageStore } from '../index.js';
+import type { RecordsWrite } from './records-write.js';
+import type { Signer } from '../types/signer.js';
 import type { RecordsDeleteDescriptor, RecordsDeleteMessage } from '../types/records-types.js';
 
 import { getCurrentTimeInHighPrecision } from '../utils/time.js';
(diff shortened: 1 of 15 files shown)


tests/handlers/records-delete.spec.ts
```diff
@@ -282,6 +288,229 @@ export function testRecordsDeleteHandler(): void {
         expect(aliceQueryWriteAfterAliceRewriteReply.entries![0].encodedData).to.equal(encodedData);
       });
 
+      describe('protocol based deletes', () => {
```
thehenrytsai, line 291: General feed back on new tests:

The test suite seems to lack "negative" tests cases which IMO is equally important for protocol-authorization paths (minus the last else-catch test which doesn't go into the protocol-authorization path): 

ie. if your are not author/recipient, can't delete. Or if you don't belong in $global/$contextRole, can't delete.

While I agree the negative cases "should" work just like other Records methods, since delete has its own implementation... consider rolling the negative cases into the tests added. Let me know if I misread the test or if you think this is an overkill.

src/interfaces/records-delete.ts
```diff
@@ -38,16 +42,25 @@ export class RecordsDelete extends Message<RecordsDeleteMessage> {
...
-  public async authorize(tenant: string): Promise<void> {
-    // TODO: #203 - implement protocol-based authorization for RecordsDelete (https://github.com/TBD54566975/dwn-sdk-js/issues/203)
-    await authorize(tenant, this);
+  public async authorize(tenant: string, newestRecordsWrite: RecordsWrite, messageStore: MessageStore): Promise<void> {
+    if (this.author === tenant) {
+      return;
+    } else if (newestRecordsWrite.message.descriptor.protocol !== undefined) {
+      await ProtocolAuthorization.authorizeDelete(tenant, this, newestRecordsWrite, messageStore);
+    } else {
+      throw new Error('message failed authorization');
```
thehenrytsai, line 63: Nit: consider turning this into `DwnError` since a community is picking up #387 with a PR likely today or tomorrow, don't want this to be left out by accident.

tests/handlers/records-delete.spec.ts
diehuxx: Updated each test to add a failed attempt by `carol` to delete a record.

tests/handlers/records-delete.spec.ts
```diff
@@ -282,6 +288,229 @@ export function testRecordsDeleteHandler(): void {
...
+            recordId : recordsWrite.message.recordId,
+          });
+          const recordsDeleteReply = await dwn.processMessage(alice.did, recordsDelete.message);
+          expect(recordsDeleteReply.status.code).to.eq(202);
+        });
+
+        describe('recipient rules', () => {
+          it('should allow delete with ancestor recipient rule', async () => {
+            // scenario: Alice creates a record
+          });
```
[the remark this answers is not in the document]
diehuxx: Woops. Updated

src/interfaces/records-delete.ts
diehuxx: Updated

LiranCohen (approved):

thehenrytsai (approved):

tests/handlers/records-delete.spec.ts
thehenrytsai: Kinda hoped the negative role test will involve Carol invoking a role somewhere, but I'll leave that to your discretion!

Issue #535, opened 2023-10-03
Add range filter support for `published` & `datePublished`


thehenrytsai:
### Background:
Currently the supported records filter is missing `datePublished`

```ts
export type RecordsFilter = {
  attester?: string;
  recipient?: string;
  protocol?: string;
  protocolPath?: string;
  contextId?: string;
  schema?: string;
  recordId?: string;
  parentId?: string;
  dataFormat?: string;
  dateCreated?: RangeCriterion;
};
```

### Task Details:
Add filter support for `datePublished`.

We can for the most part follow the pattern established by the `dateCreated` filter. But there is one significant difference:

`datePublished` is an optional field, so we need to:
1. Also introduce `published` filter and require it when `datePublished` range filter is specified.
2. We must prevent unauthorized access of reading private date even when `published` filter is set `false`. e.g. Throw error if a query/read is unauthenticated.

alanhkarp:
"We must prevent unauthorized access of reading private date even when published filter is set false. e.g. Throw error if a query/read is unauthenticated."

Do you mean, "Throw an error if a query/read is unauthorized"?

thehenrytsai:
This is now done.

PR #578, opened 2023-10-26
Add missing filters to RecordsQuery

LiranCohen:
Added the following filters to RecordsQuery:
`published`
`dataCid`
`datePublished`
`dateUpdated` (messageTimestamp)

### DIFF
--- a/json-schemas/interface-methods/records-filter.json
+++ b/json-schemas/interface-methods/records-filter.json
@@ -66,6 +66,19 @@
           "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
         }
       }
+    },
+    "dateUpdated": {
+      "type": "object",
+      "minProperties": 1,
+      "additionalProperties": false,
+      "properties": {
+        "from": {
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        },
+        "to": {
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        }
+      }
     }
   }
 }
--- a/json-schemas/interface-methods/records-filter.json
+++ b/json-schemas/interface-methods/records-filter.json
@@ -29,11 +29,17 @@
     "parentId": {
       "type": "string"
     },
+    "published": {
+      "type": "boolean"
+    },
     "dataFormat": {
       "type": "string"
     },
     "dataSize": {
       "$ref": "https://identity.foundation/dwn/json-schemas/number-range-filter.json"
+    },
+    "dataCid": {
+      "type": "string"
     },
     "dateCreated": {
       "type": "object",
--- a/json-schemas/interface-methods/records-filter.json
+++ b/json-schemas/interface-methods/records-filter.json
@@ -80,5 +80,24 @@
         }
       }
     }
+  },
+  "dependencies": {
+    "datePublished": {
+      "oneOf": [
+        {
+          "properties": {
+            "published": {
+              "enum": [true]
+            }
+          },
+          "required": ["published"]
+        },
+        {
+          "not": {
+            "required": ["published"]
+          }
+        }
+      ]
+    }
   }
 }
--- a/json-schemas/interface-methods/records-filter.json
+++ b/json-schemas/interface-methods/records-filter.json
@@ -53,6 +53,19 @@
           "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
         }
       }
+    },
+    "datePublished": {
+      "type": "object",
+      "minProperties": 1,
+      "additionalProperties": false,
+      "properties": {
+        "from": {
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        },
+        "to": {
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        }
+      }
     }
   }
 }
--- a/src/handlers/records-query.ts
+++ b/src/handlers/records-query.ts
@@ -30,8 +30,8 @@ export class RecordsQueryHandler implements MethodHandler {
 
     let recordsWrites: RecordsWriteMessageWithOptionalEncodedData[];
     let paginationMessageCid: string|undefined;
-    // if this is an anonymous query and the filters support published records, query only published records
-    if (RecordsQueryHandler.filtersForPublishedRecords(recordsQuery) && recordsQuery.author === undefined) {
+    // if this is an anonymous query and the filter supports published records, query only published records
+    if (RecordsQueryHandler.filterIncludesPublishedRecords(recordsQuery) && recordsQuery.author === undefined) {
       const results = await this.fetchPublishedRecords(tenant, recordsQuery);
       recordsWrites = results.messages as RecordsWriteMessageWithOptionalEncodedData[];
       paginationMessageCid = results.paginationMessageCid;
@@ -125,7 +125,7 @@ export class RecordsQueryHandler implements MethodHandler {
    *
    * C) UNPUBLISHED:
    *    1. unpublished records intended for the query author (where `recipient` is the query author); and
-   *    3. unpublished records authorized by a protocol rule.
+   *    2. unpublished records authorized by a protocol rule.
    *
    */
   private async fetchRecordsAsNonOwner(
@@ -134,11 +134,11 @@ export class RecordsQueryHandler implements MethodHandler {
     const { dateSort, pagination } = recordsQuery.message.descriptor;
     const filters = [];
 
-    if (RecordsQueryHandler.filtersForPublishedRecords(recordsQuery)) {
+    if (RecordsQueryHandler.filterIncludesPublishedRecords(recordsQuery)) {
       filters.push(RecordsQueryHandler.buildPublishedRecordsFilter(recordsQuery));
     }
 
-    if (RecordsQueryHandler.filtersForUnpublishedRecords(recordsQuery)) {
+    if (RecordsQueryHandler.filterIncludesUnpublishedRecords(recordsQuery)) {
       filters.push(RecordsQueryHandler.buildUnpublishedRecordsByQueryAuthorFilter(recordsQuery));
 
       const recipientFilter = recordsQuery.message.descriptor.filter.recipient;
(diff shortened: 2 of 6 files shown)


tests/handlers/records-query.spec.ts
```diff
@@ -202,6 +202,62 @@ export function testRecordsQueryHandler(): void {
         expect(reply3.entries?.length).to.equal(0);
       });
 
+      it('should be able to query for published records', async () => {
```
thehenrytsai, line 205: Please add tests for getting only unpublished records. For authorized and anonymous/unauthorized access.

Daniel is interested in the authorized case, I am more interested in the later as it would be a security breach.

LiranCohen: @thehenrytsai I added the additional test cases and logic, I think they're covering the various edge cases that exist.

I also added some schema validation that will disallow `published` to be set to false as well as a `datePublished` range to be specified within a filter.

thehenrytsai:
This PR addresses #535. Will jump back on this PR today.

src/handlers/records-query.ts
```diff
@@ -210,4 +214,22 @@ export class RecordsQueryHandler implements MethodHandler {
   private static shouldProtocolAuthorizeQuery(recordsQuery: RecordsQuery): boolean {
     return recordsQuery.authorSignaturePayload!.protocolRole !== undefined;
   }
+
+  /**
+   * Does the recordQuery filter explicitly for published records.
+   * Checks that there is either an explicit datePublished filter, or published is explicitly not set to false.
```
thehenrytsai, line 220: Sanity:

> published is explicitly not set to false.
 
Is this different to explicitly set to true? What is the expected return value when `filter.published` is `undefined`? Assuming `true`, then the method name should be renamed to `filterIncludesPublishedRecords()`

src/handlers/records-query.ts
```diff
@@ -210,4 +214,22 @@ export class RecordsQueryHandler implements MethodHandler {
...
+    return filter.datePublished !== undefined || filter.published !== false;
+  }
+
+  /**
+   * Does the recordQuery filter explicitly for unpublished records.
+   * Checks that published is not explicitly set to true, and datePublished is explicitly undefined.
+   */
+  private static filtersForUnpublishedRecords(recordsQuery: RecordsQuery): boolean {
+    const { filter } = recordsQuery.message.descriptor;
+    return filter.published !== true && filter.datePublished === undefined;
```
thehenrytsai, line 233: Sanity: I have similar confusion as above. The method seems to imply if `published` is `undefined` and `datePublished` is also `undefined`, then it is a filter for unpublished records? Maybe a rename to `filterIncludesUnpublishedRecords()` better matches the intent?

json-schemas/interface-methods/records-filter.json
```diff
@@ -47,6 +53,51 @@
...
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        },
+        "to": {
+          "$ref": "https://identity.foundation/dwn/json-schemas/defs.json#/definitions/date-time"
+        }
+      }
+    }
+  },
+  "dependencies": {
+    "datePublished": {
```
thehenrytsai, line 85: For my understanding: is this just a perf optimization or is there anything broken if we don't have this dependency check? Wouldn't the outcome just that there are zero entries returned?

Fair enough if this is just a silly query prevention mechanism, just want to make sure I understand the intent.

tests/handlers/records-query.spec.ts
```diff
@@ -735,6 +1185,20 @@ export function testRecordsQueryHandler(): void {
...
+
+        // explicitly for published records
+        const anonymousQueryUnpublished = await TestDataGenerator.generateRecordsQuery({
+          anonymous : true,
+          filter    : { dateCreated: { from: '2000-01-01T10:20:30.123456Z' }, published: true }
+        });
+        // sanity check
+        expect(anonymousQueryUnpublished.message.authorization).to.not.exist;
+
+        // should not return any records
```
thehenrytsai, line 1197: Sanity: says should not return any records, but checks for entry count = 1.

tests/handlers/records-query.spec.ts
```diff
@@ -202,6 +202,133 @@ export function testRecordsQueryHandler(): void {
...
+
+        // anonymous query after publishing
+        anonymousPublishedPostReply = await dwn.processMessage(alice.did, anonymousPostQuery.message);
+        expect(anonymousPublishedPostReply.status.code).to.equal(200);
+        expect(anonymousPublishedPostReply.entries?.length).to.equal(2);
+        const anonymousReturnedRecordIds = anonymousPublishedPostReply.entries?.map(e => e.recordId);
+        expect(anonymousReturnedRecordIds).to.have.members([ publishedWrite.message.recordId, draftWrite.message.recordId ]);
+      });
+
+      it('should be able to query for unpublished records', async () => {
```
thehenrytsai, line 273:
```suggestion
      it('should be able to query for unpublished records only if authorized', async () => {
```
The rename implies that we test for both positive and negative cases. Alternatively you can take out the test case for Bob as its own. Bottomline is that IMO testing for unauthorized `published: false` is important, so it needs to be signaled somehow that it is being tested in the sea of tests we've got.

tests/handlers/records-query.spec.ts
```diff
@@ -988,7 +1472,7 @@ export function testRecordsQueryHandler(): void {
         expect(replyToAliceQuery.status.code).to.equal(200);
         expect(replyToAliceQuery.entries?.length).to.equal(1);
 
-        // actual test: bob should not be able to see unpublished record
+        // actual test: bob should not be able to see unpublished record without published filter
```
thehenrytsai, line 1475: Sanity: why add the "without published filter"? It seems to imply that bob should be able to see unpublished record if `published` is explicitly set to `false`?

tests/handlers/records-query.spec.ts
```diff
@@ -1083,6 +1567,44 @@ export function testRecordsQueryHandler(): void {
...
+        // anonymous query for unpublished records
+        const unpublishedAnonymous = await RecordsQuery.create({ filter: { schema: 'post', published: false } });
+        const anonymousPostReply = await dwn.processMessage(alice.did, unpublishedAnonymous.message);
+        expect(anonymousPostReply.status.code).to.equal(401);
+        expect(anonymousPostReply.status.detail).contains('Missing JWS');
+      });
+
       describe('protocol based queries', () => {
         it('does not try protocol authorization if protocolRole is not invoked', async () => {
           // scenario: Alice creates a thread and writes some chat messages writes a chat message. Alice addresses
```
thehenrytsai, line 1610: Opportunistic fix.
```suggestion
          // scenario: Alice creates a thread and writes some chat messages. Alice addresses
```

src/handlers/records-query.ts
LiranCohen: @thehenrytsai if `published` is undefined the filter for that property isn't evaluated at all.
That's today's behavior without this change. So a DWN owner would get both published/unpublished without it.

For a non-owner, we create multiple queries depending on the author/recipient/protocol auth and injecting additional filtered properties to the individual queries.

I wonder how I could make this intent clearer.

src/handlers/records-query.ts
LiranCohen: @thehenrytsai same response as above, without either filters it will return both published and unpublished values.

With one of either published explicitly true or a datePublished filter set, it will only return `published` records.

Maybe I can make that clearer in the comments?

tests/handlers/records-query.spec.ts
LiranCohen: Hrm, yeah look at it again I think that pulling Bob's case into it's own would describe the specific intent better. I'll do that.

tests/handlers/records-query.spec.ts
LiranCohen: Ah yeah I should remove that. I was going to add a negative test here but It's already covered somewhere else with the intent more clear for that case.

I'll revert to:
```suggestion
        // actual test: bob should not be able to see unpublished record
```

json-schemas/interface-methods/records-filter.json
LiranCohen: If this check wasn't in place it would actually return only `published` records, because we automatically inject `published: true` when you are filtering by `datePublished`.

So since the nonsensical intent of the query isn't clear, I think it's better to reject it rather than return published results.

tests/handlers/records-query.spec.ts
LiranCohen: Good catch! I renamed the variables too, I had this testing the 401 but then moved to a more explicit `should return 401 for anonymous queries that filter explicitly for unpublished records`, forgot to rename the comment/variables for the positive check.

json-schemas/interface-methods/records-filter.json
thehenrytsai: Got it. Good that I asked, 'coz I thought it was an perf optimization thing.

src/handlers/records-query.ts
thehenrytsai: Same as the other comment. Can we rename to `filterIncludesPublishedRecords()`.

src/handlers/records-query.ts
```diff
@@ -110,9 +110,23 @@ export class RecordsQueryHandler implements MethodHandler {
...
+   *    1. published records; and
+   *    2. unpublished records intended for the query author (where `recipient` is the query author); and
+   *    3. unpublished records authorized by a protocol rule.
+   *
+   * B) PUBLISHED:
+   *    1. only published records;
+   *
+   * C) UNPUBLISHED:
+   *    1. unpublished records intended for the query author (where `recipient` is the query author); and
+   *    3. unpublished records authorized by a protocol rule.
```
thehenrytsai, line 128: Unimportant nit:
```suggestion
   *    2. unpublished records authorized by a protocol rule.
```

src/handlers/records-query.ts
```diff
@@ -216,20 +230,26 @@ export class RecordsQueryHandler implements MethodHandler {
...
+   * Checks if the recordQuery filter supports returning unpublished records.
    */
   private static filtersForUnpublishedRecords(recordsQuery: RecordsQuery): boolean {
     const { filter } = recordsQuery.message.descriptor;
-    return filter.published !== true && filter.datePublished === undefined;
+    // When `published` and `datePublished` range are both undefined, unpublished records can be returned.
+    if (filter.datePublished === undefined && filter.published === undefined) {
+      return true;
+    }
+    return filter.published === false && filter.datePublished === undefined;
```
thehenrytsai, line 253: Sanity: would this suffice

```suggestion
    return filter.published !== true || filter.datePublished === undefined;
```

thehenrytsai, line 253: Sanity: would this be good enough?
```suggestion
    return filter.published === false;
```

src/handlers/records-query.ts
```diff
@@ -216,20 +230,26 @@ export class RecordsQueryHandler implements MethodHandler {
...
+   * Checks if the recordQuery filter supports returning published records.
    */
   private static filtersForPublishedRecords(recordsQuery: RecordsQuery): boolean {
     const { filter } = recordsQuery.message.descriptor;
-    return filter.datePublished !== undefined || filter.published !== false;
+    // When `published` and `datePublished` range are both undefined, published records can be returned.
+    if (filter.datePublished === undefined && filter.published === undefined) {
+      return true;
+    }
+    return filter.datePublished !== undefined || filter.published === true;
```
thehenrytsai, line 241: Sanity: why is the original implementation not sufficient?

```suggestion
    return filter.datePublished !== undefined || filter.published !== false;
```

src/handlers/records-query.ts
LiranCohen: Yup, since we fail for the invalid filter of published = false + datePublished range in the schema check. fixed.

src/handlers/records-query.ts
LiranCohen: It was, and since the comment is there it shows more clear intent so reverted back to original implementation.

src/handlers/records-query.ts
LiranCohen: Yup, it makes sense and reads well through the rest of the code.

thehenrytsai (approved):
🙏

Issue #324, opened 2023-04-26
Validate timestamps when parsing incoming message


thehenrytsai:
Currently there is no validation of timestamp properties such as `dateCreated`, `datePublished` etc when parsing the incoming DWN message. We should add validation logic such that the string given is not an invalid timestamp.

AdiDecodes:
@thehenrytsai picking this up

EbonyLouis:
@AdiDecodes Yay! I'll assign it to you now.

 I'd also like to note that Hacktoberfest officially kicks off on **October 1st**. Contributions made before the event may not be eligible for the challenge. **However, feel free to start working on it, but I'd recommend holding off on submitting your PR until the event begins.** 

> to be eligible for Hacktoberfest's official prizes make sure you're registered, registration [is now open.](https://hacktoberfest.com/auth/) Reach out if you need any help 🥳 

Good Luck 🍁

AdiDecodes:
> @AdiDecodes Yay! I'll assign it to you now.
> 
> I'd also like to note that Hacktoberfest officially kicks off on **October 1st**. Contributions made before the event may not be eligible for the challenge. **However, feel free to start working on it, but I'd recommend holding off on submitting your PR until the event begins.**
> 
> > to be eligible for Hacktoberfest's official prizes make sure you're registered, registration [is now open.](https://hacktoberfest.com/auth/) Reach out if you need any help 🥳
> 
> Good Luck 🍁

Thankyou, waiting till hacktoberfest starts!

EbonyLouis:
Hey @AdiDecodes how are things going with this issue? Is this something you still want to work on?

AdiDecodes:
> Hey @AdiDecodes how are things going with this issue? Is this something you still want to work on?

Sorry, But really I am trying to figure out what the issue is. I understood What I need to do. But I didn't understood the file that needs to be modified. (Just started Backend development).

flothjl:
@EbonyLouis  if this is inactive I would pick this up.

EbonyLouis:
@flothjl Hey, yes you can pick this up.

flothjl:
Hey @thehenrytsai !

It looks like #147  was closed around the time this was created which added datetime regex validation in the schema. 

I updated schemas missing the reference to the `date-time` regex, but before I dive into adding validation to ensure these datetimes are parseable I wanted to make sure this is something you still want additional validaiton on beyond the schema regex.

flothjl:
[...]

PR #580, opened 2023-10-27
adding time schema validation to other interface method schemas

flothjl:
Addresses #324 

After digging into this, it appears a lot of progress has already been made on schema-based validation in #147 but adding some more robust validation by ensuring the timestamp is parseable. 

- updated `interface-method` schemas that have timestamps without the `date-time` schema reference
- added `validateTimestamp` to ensure timestamps are valid


thehenrytsai (approved):
This is good to merge in the first pass!
🙏 🙏 🙏

PR #582, opened 2023-10-27
Delegated grant support for RecordsWrite

thehenrytsai:
This is the first iteration of many to come on delegated grants. But even when intentionally keeping scope of PR to minimum, 30 files are touched...

1. Added basic delegated grant support for RecordsWrite
1. Necessary differentiation between "author" and "signer"
1. Added tests to prove that delegated grants work with protocol roles
1. 100% code coverage on new code added

What's to come in future PRs:
1. Tests. A lot more tests. Intent is shown with tests shown in `xit()`
1. Add delegated grant support for `RecordsRead`, `RecordsQuery`, and `RecordsDelete`
1. Scoping support for delegated grants. There are nuance here there will need be unpacked, specifically around working in conjunction with protocol role.
1. Renames, a bunch of names requiring renames are left untouched intentionally to avoid noise drowning out (further) the essence of the PR.
1. Consider refactoring opportunity to not needing add delegated grant support one method at a time (may or may not be difficult/worth it).


tests/interfaces/records-write.spec.ts
```diff
@@ -287,13 +316,48 @@ describe('RecordsWrite', () => {
...
+      });
+
+      const recordsWrite = await RecordsWrite.create({
+        signer         : Jws.createSigner(alice),
+        delegatedGrant : grantToBob.asDelegatedGrant(),
+        dataFormat     : 'application/octet-stream',
+        data           : TestDataGenerator.randomBytes(10),
+      });
+
+      delete recordsWrite.message.authorization!.authorDelegatedGrant; // intentionally remove `authorDelegatedGrant`
```
diehuxx, line 353: Sanity: This looks to be the opposite of what is described in the test name `should throw if delegated grant is given but signer is not given`.

diehuxx (dismissed):
One question/comment, but otherwise, given that the TODOs will be addressed in a follow up, this LGTM

tests/handlers/records-write.spec.ts
```diff
@@ -478,6 +478,262 @@ export function testRecordsWriteHandler(): void {
...
+            protocol     : protocolDefinition.protocol,
+            protocolPath : 'thread/participant',
+            contextId    : threadRecord.message.contextId,
+            parentId     : threadRecord.message.recordId,
+            data         : new TextEncoder().encode('Alice is my friend'),
+          });
+          const participantRoleReply = await dwn.processMessage(carol.did, participantRoleRecord.message, participantRoleRecord.dataStream);
+          expect(participantRoleReply.status.code).to.equal(202);
+
+          // Alice creates a delegated grant for device X and device Y
```
LiranCohen, line 650:
```suggestion
          // Alice creates a delegated grant for Bob to act as Alice.
```

LiranCohen (approved):
Looks good, just a comment correction on one of the tests.

Also TIL about `xit()`!

tests/interfaces/records-write.spec.ts
thehenrytsai: Ooops, thanks for the keen eyes, C&P error (there is another test with the exact same name), fixed!

tests/handlers/records-write.spec.ts
thehenrytsai: Thanks for picking it up! Yet another C&P error! Fixed.

LiranCohen (approved):

Issue #570, opened 2023-10-24
Create a utility method for creating a DWN compatible timestamp


thehenrytsai:
### Background:
When creating a message, there is ability to supply your own timestamp in IOS 8601 format with microsecond precision, but we don't provide a utility method for this.

### Task Details:
Create a utility method for creating a timestamp accepted by DWN.

A potential method signature:

```ts
  /**
   * Creates a UTC ISO-8601 timestamp in microsecond precision accepted by DWN.
   */
  public static createTimestamp(year: number, month: number, day: number, hour: number:, minute: number, second: number, subsecond: number): string {
    ...
  }

```

For implementation reference you can check out [this test utility](https://github.com/TBD54566975/dwn-sdk-js/blob/e5a1a31f973060353654a9cf83d18249a092e515/tests/utils/test-data-generator.ts#L830).

## Picking Up This Issue:
- If you'd like to work on this, please comment "picking this up" below, and I'll assign the issue to you 

## Questions:
- If you need assistance or clarification, feel free to comment below.
- Feeling overwhelmed? Join our [`#hack-together` channel](https://discord.gg/tbd) and collaborate with a buddy.
- New to open-source? Check out our [Hacktober `#getting-started` channel](https://discord.gg/tbd).
- [Leaderboard](https://github.com/TBD54566975/developer.tbd.website/issues/721): Check your ranking after finishing this task.

### Resources:
- **Creating a Pull Request**: If you're new to GitHub and unsure how to create a pull request, follow this [step-by-step guide](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Remember, communication is key! If you have any questions or face any challenges, we're here to help so please don't hesitate to reach out.

Good Luck! 🍁

Dhoni77:
picking this up

flothjl:
@Dhoni77  just seeing if you are still wanting to work on this. If not, I'm happy to pick it up! 
@EbonyLouis

Dhoni77:
> @Dhoni77 just seeing if you are still wanting to work on this. If not, I'm happy to pick it up! @EbonyLouis

Hi @flothjl you can pick it👍

thehenrytsai:
This is now done.

PR #583, opened 2023-10-28
adding helper function to create a timestamp

flothjl:
Addresses #570 

- Adds and exports a helper function to create timestamps to use in dwn messages.

### DIFF
--- a/src/utils/time.ts
+++ b/src/utils/time.ts
@@ -19,6 +19,34 @@ export function getCurrentTimeInHighPrecision(): string {
 }
 
 /**
+ * Creates a UTC ISO-8601 timestamp in microsecond precision accepted by DWN.
+ * @param year
+ * @param month
+ * @param day
+ * @param hour
+ * @param minute
+ * @param second
+ * @param millisecond
+ * @param microsecond
+ * @returns string
+ */
+export function createTimestamp(
+  year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number
+): string {
+  return Temporal.ZonedDateTime.from({
+    timeZone: 'UTC',
+    year,
+    month,
+    day,
+    hour,
+    minute,
+    second,
+    millisecond,
+    microsecond
+  }).toInstant().toString({ smallestUnit: 'microseconds' });
+}
+
+/**
  * We must sleep for at least 2ms to avoid timestamp collisions during testing.
  * https://github.com/TBD54566975/dwn-sdk-js/issues/481
  */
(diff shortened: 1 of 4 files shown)


csuwildcat (commented):
Does the method allow for calling with no parameters and returning an auto-generated timestamp too?

flothjl:
@csuwildcat  currently no. There is another helper method `getCurrentTimeInHighPrecision`

LiranCohen:
@csuwildcat when you say auto-generated do you mean of the current time as @flothjl mentioned or a random timestamp for testing?

We do have a crude `randomTimestamp` in the `TestDataGenerator` class. Maybe the intent of these can be combined?

https://github.com/TBD54566975/dwn-sdk-js/blob/main/tests/utils/test-data-generator.ts#L834-L844

csuwildcat:
I meant getting the current time in this format. I know the records create/update code does this automatically if omitted, just might be worth it exposing auto-generated/- filled timestamps if you call with no params or less than full params. Could be a waste of time though, because the most important thing is just having a function to cast values into the right format.

LiranCohen:
@csuwildcat ah yeah, in that case I think keeping the current `getCurrentTimeInHighPrecision` makes more sense, that way the two different functions have clearer intents without having the optional values.

src/utils/time.ts
```diff
@@ -18,6 +18,34 @@ export function getCurrentTimeInHighPrecision(): string {
...
+ * @param day
+ * @param hour
+ * @param minute
+ * @param second
+ * @param millisecond
+ * @param microsecond
+ * @returns string
+ */
+export function createTimestamp(
+  year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number
```
thehenrytsai, line 34: All parameters should be optional for this function just like the `from` method in the `Temporal` lib used in this implementaiton.

src/utils/time.ts
```diff
@@ -31,7 +31,7 @@ export function getCurrentTimeInHighPrecision(): string {
  * @returns string
  */
 export function createTimestamp(
-  year: number, month: number, day: number, hour: number, minute: number, second: number, millisecond: number, microsecond: number
+  year?: number, month?: number, day?: number, hour?: number, minute?: number, second?: number, millisecond?: number, microsecond?: number
```
thehenrytsai, line 34: Wrap in options so that it is always clear what properties are being set, so there is less chance for misinterpretation or usage.
```suggestion
  options: { year?: number, month?: number, day?: number, hour?: number, minute?: number, second?: number, millisecond?: number, microsecond?: number }
```

thehenrytsai (approved):
🙏 🔥 🎖️

PR #594, opened 2023-11-02
Fixed bug on usage of `GeneralJwsVerifier` and improved its usability

thehenrytsai:
1. Fixed bug where a new cache is created every time the DWN performs a signature check
1. Removed the need to instantiate a new instance of `GeneralJwsVerifier` every time DWN performs a signature check
1. Improved perf by not doing DID resolution when the signature check result is already cached
1. Minor error code naming improvements

### DIFF
--- a/tests/jose/jws/general.spec.ts
+++ b/tests/jose/jws/general.spec.ts
@@ -193,7 +193,7 @@ describe('General JWS Sign/Verify', () => {
     const verifySignatureSpy = sinon.spy(Jws, 'verifySignature');
     const cacheSetSpy = sinon.spy((GeneralJwsVerifier as any).singleton.cache, 'set');
 
-    // intentionally calling verifySignature twice on the same JWS
+    // intentionally calling verifySignatures() multiple times on the same JWS
     await GeneralJwsVerifier.verifySignatures(jws, resolverStub);
     await GeneralJwsVerifier.verifySignatures(jws, resolverStub);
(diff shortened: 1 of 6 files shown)


tests/jose/jws/general.spec.ts
```diff
@@ -195,16 +190,16 @@ describe('General JWS Sign/Verify', () => {
...
     const verifySignatureSpy = sinon.spy(Jws, 'verifySignature');
-    const cacheSetSpy = sinon.spy(verifier.cache, 'set');
+    const cacheSetSpy = sinon.spy((GeneralJwsVerifier as any).singleton.cache, 'set');
 
-    await verifier.verify(resolverStub);
-    await verifier.verify(resolverStub);
+    // intentionally calling verifySignature twice on the same JWS
+    await GeneralJwsVerifier.verifySignatures(jws, resolverStub);
+    await GeneralJwsVerifier.verifySignatures(jws, resolverStub);
+    await GeneralJwsVerifier.verifySignatures(jws, resolverStub);
```
LiranCohen, line 199:
```suggestion
```

sanity: comment says twice, was there a reason to call it 3 times or just a mistake? It shouldn't affect the test outcome.

LiranCohen (dismissed):
Looks good! Just a comment in the tests, but isn't an actual problem.

LiranCohen (approved):
🚀

Issue #564, opened 2023-10-17
Add support for delegated grants for DWAs that use the grantedFor field


shamilovtim:
We need support for delegated grants in order to accommodate Web5 Connect on DWAs that use the `grantedFor` field

thehenrytsai:
This is done.

PR #597, opened 2023-11-03
#564 - Added delegated grant support for RecordsQuery

thehenrytsai:
1. Added delegated grant support for `RecordsQuery`
1. Updated `Message.createAuthorization()` method signature
1. Moved all delegated grant scenario tests into its own file, establishing a new pattern for how tests should be organized going forward, as `RecordsWrite` tests is growing out of control.
1. Updated Q&A md on why `protocolPath` is a required filter property when invoking a protocol role.

P.S. Delegated grant support for `RecordsRead` will be the next PR, at which time, there maybe refactoring opportunity to observe patterns and see if we can reduce code duplication.


LiranCohen (dismissed):
This is great! 🚀

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+
+      const grantToDeviceX = await PermissionsGrant.create({
+        delegated   : true, // this is a delegated grant
+        dateExpires : Time.createOffsetTimestamp({ seconds: 100 }),
+        description : 'Allow device X to write as me in chat protocol',
+        grantedBy   : alice.did,
+        grantedTo   : bob.did,
+        grantedFor  : alice.did,
+        scope       : scope,
+        signer      : Jws.createSigner(alice)
```
diehuxx, line 236: It seems like the intention is to create a grant allowing deviceX to do `RecordsQuery` as Alice. But this is actually creating a grant allowing Bob to `RecordsWrite` as Alice. Below, the grant is being used to authorize deviceX to query, so if this test is passing as it is currently written, that may be a bug.

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -0,0 +1,422 @@
...
+        signer         : Jws.createSigner(deviceX),
+        delegatedGrant : grantToDeviceX.asDelegatedGrant(),
+        protocolRole   : 'thread/participant',
+        filter         : {
+          protocol,
+          contextId    : threadRecord.message.contextId,
+          protocolPath : 'thread/chat'
+        }
+      });
+      const deviceXRecordsQueryReply = await dwn.processMessage(bob.did, recordsQueryByDeviceX.message);
```
diehuxx, line 273: might be worth testing that `recordsQueryByDeviceX.recordsQuery.author === alice.did`.

tests/scenarios/delegated-grant.spec.ts
diehuxx: RecordsQueryHandler doesn't seem to be checking that the granteTo of the `DelegatedPermissionsGrant` matches the author of the `RecordsQuery` nor that the `scope` is `interface: 'Records', method: 'Query'`

diehuxx: For `RecordsWrites` with delegated grants, we check that `grantedTo` and `author` match in `RecordsWrite.validateIntegrity`. Maybe we could move/add that check to `authenticate` since we already do some verification of the delegated grant in that function. For checking `scope`, we need to add a new method `RecordsGrantAuthorization.authorizeQuery` and call it from `RecordsQueryHandler`.

thehenrytsai: @diehuxx, thanks for spotting the dumb mistake in tests on line 233!  Yeah, I am aware I haven't checked author, which is why the test (line 395) is not implemented like I've done for `RecordsWrite` (line 293), but since I screwed up the positive test and ended up testing for a negative case, I'll just add the logic now, so that `RecordsQuery` is on parity with `RecordsWrite`.

Thanks for calling out on scoping, I intend to consider after all main Records methods are addressed, right now any delegated grant from Alice gives you full power to act as Alice. This is why scoping tests (line 401 ~ 411) are left as `xit()` placeholders.

On moving some checks into `authenticate`, definitely interested in refactoring to reduce code duplication, I'll take a look if it makes sense (or easy) to do this now. In general I expect a refactoring to take place when delegated grant support for `RecordsRead` and `RecordsDelete` are added.

diehuxx: > Thanks for calling out on scoping, I intend to consider after all main Records methods are addressed, right now any delegated grant from Alice gives you full power to act as Alice. 

Ahh, I didn't realize that was the intention. Is the intention to continue supporting an all-powerful delegated grant, or is this just an interim state of affairs before we fully set up scoping?

thehenrytsai: Updated the PR. I refactored the delegated grant referential validation into its own method, but not quite ready to move it into `authenticate()` yet, would love to discuss the pros and cons with you offline.

tests/scenarios/delegated-grant.spec.ts
thehenrytsai: Is the new test added what you are looking for or something else?

tests/scenarios/delegated-grant.spec.ts
thehenrytsai: > is this just an interim state of affairs before we fully set up scoping?

It can go either way, have yet to discuss if all-powerful grant is needed, but I assume we don't support it for permissions grant, then we don't support when scope is evaluated. But yeah, it speaks to your point about delegated grant being different to permissions grant.

diehuxx (approved):

tests/scenarios/delegated-grant.spec.ts
diehuxx: Yep, that's sufficient

tests/scenarios/delegated-grant.spec.ts
diehuxx: Maybe both delegated and regular permissions grants should have an "all-powerful" grant. We can revisit in the future.

Issue #599, opened 2023-11-06
Remove circular dependencies 


diehuxx:
We ran into a circular dependency issue when trying to update `dwn-sdk-js` version in `dwn-sql-store`. https://github.com/TBD54566975/dwn-sql-store/pull/9. I was able to resolve it quickly with #598.

But, we have plenty of other circular dependencies in `dwn-sdk-js`. They don't seem to be causing immediate problems, but we should resolve them soon since they are hard to debug. Consider using a tool like [madge](https://www.npmjs.com/package/madge) to detect circ dependencies.

thehenrytsai:
This should be done now.

PR #601, opened 2023-11-07
#599 - Removed remaining circular dependencies and integrated madge in build

thehenrytsai:
1. Fixed remaining more tricky circular dependencies
1. Added `madge` to detect future circular dependencies

Basically the work involves painful shuffling of code and some refactoring in some cases, but this is "good" pain as circular dependencies almost always reveal design/code issues.


package.json
```diff
@@ -153,10 +154,11 @@
     "compile-validators": "node ./build/compile-validators.js",
     "lint": "eslint . --ext .ts --max-warnings 0",
     "lint:fix": "eslint . --ext .ts --fix",
-    "test:node": "npm run compile-validators && tsc && c8 mocha \"dist/esm/tests/**/*.spec.js\"",
+    "madge": "madge --circular ./src/index.ts",
+    "test:node": "npm run madge && npm run compile-validators && tsc && c8 mocha \"dist/esm/tests/**/*.spec.js\"",
```
diehuxx, line 158: I have a slight preference for keeping `madge` out of `test:node` since I use `test:node` often locally during development. I see `madge` similar to `lint` where it is good to have in CI, and would be tedious if it blocked testing locally while I'm still working/debugging locally.

diehuxx (approved):
looks great! One comment about putting `madge` in our testing/CI.

package.json
thehenrytsai: The change is easy either way so I'll make that change. Though, I am slightly more on the fence on this one comparing to `lint`, mainly because the changes are more involved if you only found out after you submit a PR that you've introduced a circular dependency, it can be more annoying, but the fault is on the author so fair enough I guess!

diehuxx (approved):

PR #603, opened 2023-11-07
Authorize direct recipient of protocol records

diehuxx:
Allow actions rules of the format `who: 'recipient', can: ...` for `update` and `delete`.

We will not use direct recipient for `read`, `write`, or `query` because:
* Recipients are always allowed to `read`.
* `write` entails ability to create and update, whereas `update` only allows for updates. There is no 'recipient' until the record has been created, so it makes no sense to allow `recipient` to `write`.
* At this time, `query` is only authorized using `roles`, so allowing direct recipients to query is outside the scope of this PR.

### DIFF
--- a/src/interfaces/protocols-configure.ts
+++ b/src/interfaces/protocols-configure.ts
@@ -135,6 +135,11 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
       }
 
       // Validate that if `who === recipient` and `of === undefined`, then `can` is either `delete` or `update`
+      // We will not use direct recipient for `read`, `write`, or `query` because:
+      // - Recipients are always allowed to `read`.
+      // - `write` entails ability to create and update, whereas `update` only allows for updates.
+      //    There is no 'recipient' until the record has been created, so it makes no sense to allow recipient to write.
+      // - At this time, `query` is only authorized using roles, so allowing direct recipients to query is outside the scope of this PR.
       if (action.who === ProtocolActor.Recipient &&
           action.of === undefined &&
           !['update', 'delete'].includes(action.can)
--- a/src/interfaces/protocols-configure.ts
+++ b/src/interfaces/protocols-configure.ts
@@ -2,6 +2,7 @@ import type { Signer } from '../types/signer.js';
 import type { ProtocolDefinition, ProtocolRuleSet, ProtocolsConfigureDescriptor, ProtocolsConfigureMessage } from '../types/protocols-types.js';
 
 import { Message } from '../core/message.js';
+import { ProtocolActor } from '../types/protocols-types.js';
 import { Time } from '../utils/time.js';
 import { DwnError, DwnErrorCode } from '../core/dwn-error.js';
 import { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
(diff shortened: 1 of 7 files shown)


src/interfaces/protocols-configure.ts
```diff
@@ -133,8 +134,19 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
         );
       }
 
-      // Validate that if `who` is not set to `anyone` then `of` is set
-      if (action.who !== undefined && ['author', 'recipient'].includes(action.who) && !action.of) {
+      // Validate that if `who === recipient` and `of === undefined`, then `can` is either `delete` or `update`
```
thehenrytsai, line 137: Your PR comment on why only `delete` and `update` are valid is very insightful, can you add those to the comments here?

src/interfaces/protocols-configure.ts
```diff
@@ -133,8 +134,19 @@ export class ProtocolsConfigure extends Message<ProtocolsConfigureMessage> {
...
+          DwnErrorCode.ProtocolsConfigureInvalidRecipientOfAction,
+          'Rules for `recipient` without `of` property must have `can` === `delete` or `update`'
+        );
+      }
+
+      // Validate that if `who` is set to `author` then `of` is set
+      if (action.who === ProtocolActor.Author && !action.of) {
         throw new DwnError(
           DwnErrorCode.ProtocolsConfigureInvalidActionMissingOf,
           `'of' is required at protocol path (${protocolPath})`
```
thehenrytsai, line 152: Error message can be misleading consider something like:

```suggestion
          `'of' is required when 'author' is specified as 'who'`
```

tests/interfaces/protocols-configure.spec.ts
```diff
@@ -285,7 +285,35 @@ describe('ProtocolsConfigure', () => {
           .to.be.rejectedWith(DwnErrorCode.ProtocolsConfigureInvalidActionOfNotAllowed);
       });
 
-      it('rejects protocol definitions with actions that don\'t contain `of` and  `who` is `author` or `recipient`', async () => {
+      it('rejects protocol definitions with actions that have recipient-can rules with actions other than delete or update', async () => {
```
thehenrytsai, line 288: Sanity:

```suggestion
      it('rejects protocol definitions with actions that have direct-recipient-can rules with actions other than delete or update', async () => {
```

tests/handlers/records-write.spec.ts
```diff
@@ -1258,6 +1258,54 @@ export function testRecordsWriteHandler(): void {
...
+            // Carol is unable to delete the 'post'
+            const carolRecordsWrite = await TestDataGenerator.generateFromRecordsWrite({
+              author        : carol,
+              existingWrite : recordsWrite.recordsWrite
+            });
+            const carolRecordsWriteReply = await dwn.processMessage(alice.did, carolRecordsWrite.message);
+            expect(carolRecordsWriteReply.status.code).to.eq(401);
+            expect(carolRecordsWriteReply.status.detail).to.contain(DwnErrorCode.ProtocolAuthorizationActionNotAllowed);
+
+            // Bob is able to delete the post
```
LiranCohen, line 1301:
```suggestion
            // Bob is able to update the post
```

tests/handlers/records-write.spec.ts
```diff
@@ -1258,6 +1258,54 @@ export function testRecordsWriteHandler(): void {
...
+            const recordsWrite = await TestDataGenerator.generateRecordsWrite({
+              author       : alice,
+              recipient    : bob.did,
+              protocol     : protocolDefinition.protocol,
+              protocolPath : 'post',
+            });
+            const recordsWriteReply = await dwn.processMessage(alice.did, recordsWrite.message, recordsWrite.dataStream);
+            expect(recordsWriteReply.status.code).to.eq(202);
+
+            // Carol is unable to delete the 'post'
```
LiranCohen, line 1292:
```suggestion
            // Carol is unable to update the 'post'
```

LiranCohen (commented):
This is great! 🎸 Just those small comment nits.

src/interfaces/protocols-configure.ts
diehuxx: Added

LiranCohen (approved):

thehenrytsai (approved):
🎸 🏅

PR #605, opened 2023-11-10
#564 - Added delegated grant support for RecordsRead

thehenrytsai:


src/types/records-method.ts
```diff
@@ -0,0 +1,24 @@
+import type { GenericMessage, GenericSignaturePayload } from '../types/message-types.js';
+
+/**
+ * A signer that is capable of generating a digital signature over any given bytes.
```
diehuxx, line 4: copy-paste error. Also, what is the reason for `RecordsMethod` interface? It looks like any `Message` will satisfy this interface.

src/types/records-method.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export interface RecordsMethod<M extends GenericMessage> {
+  /**
+   * Valid JSON message representing this RecordsQuery.
```
diehuxx, line 8: Copy paste error

src/types/authorization-model.ts
diehuxx, line 1: empty file?

src/types/records-method.ts
diehuxx: Is part of the intention to have `RecordsQuery` and `RecordsRead` implement `RecordsMethod`? Looks like only `RecordsWrite implements RecordsMethod`.

thehenrytsai: > It looks like any Message will satisfy this interface.

It does for the most part with the exception that `RecordsWrite` can't extend `Message` easily due to its actual implementation. So the intention is to have the newly defined "real" interface act as the true type (as programming interface intends) when passing the variable instead of depending on an implementation (`Message`).

`RecordsQuery`, and `RecordsRead` still indirectly implements `RecordsMethod` though `Message`.

I originally was thinking to create the new interface just for `Records` related methods, but looks like it can (and should) be renamed to something more generic across DWN Interface, so I am now calling it `InterfaceMethod` instead... not happy with the name at all, so open to suggestions.

src/types/authorization-model.ts
thehenrytsai: Yeah... hence deleting. I think it got merged in my earlier PR attempting to address circular dependencies.

src/types/records-method.ts
thehenrytsai: Thanks! Fixed.

src/types/records-method.ts
thehenrytsai: Actually... really don't like the names, gonna change it.

src/interfaces/records-write.ts
```diff
@@ -140,7 +141,7 @@ export type CreateFromOptions = {
 /**
  * A class representing a RecordsWrite DWN message.
  */
-export class RecordsWrite {
+export class RecordsWrite implements MessageInterface<RecordsWriteMessage> {
```
LiranCohen, line 144: Sanity: Just want to make sure I understand this, we're not extending the new AbstractMessage here because this is a message that is potentially unsigned and returns the error when attempting to access?

LiranCohen (commented):
Looks good overall! Nice abstraction, should help make things clean for other interfaces also.

src/interfaces/records-write.ts
thehenrytsai: Mainly because of conflicting `_message` type is not just of `M` generic type.

LiranCohen: Thanks, makes sense! Added note helps for future reference.

LiranCohen (approved):
Looks good!

PR #609, opened 2023-11-14
#564 - Added delegated grant support for RecordsDelete

thehenrytsai:
1. Added delegated grant support for `RecordsDelete`
1. Merged mainline positive and negative cases of delegated grants into one test to reduce context switching.


LiranCohen (approved):
Very nice! Having the negative tests right there is really nice.

PR #612, opened 2023-11-16
Allowed protocol-authorized RecordsWrite to be schema-less

thehenrytsai:


LiranCohen (commented):
Curious what prompted this change vs requiring protocols to define a schema for a record path/type?

Not sure I have an opinion in either direction, but just curious what some thoughts around this are.

Also, do we have a test for the opposite scenario where the `schema` is defined in the protocol definition but one is not provided during the write or provides the wrong schema?

thehenrytsai:
> Curious what prompted this change

Mainly from @csuwildcat needing to attach a schema-less record, such as an image to some record in protocol-space, which  is fair enough IMO. Also, FYI as a side node, `schema` at this point has become a first-class tag that bears not much operational meaning, so observant devs can intentionally not require a value thus abuse it to serve their needs at this point.

> do we have a test for the opposite scenario where the schema is defined in the protocol definition but one is not provided during the write or provides the wrong schema?

Yes, we do: `should fail authorization if record schema is not an allowed type for protocol-based RecordsWrite`

LiranCohen:
> such as an image to some record in protocol-space

Yeah this makes total sense! I want to review the `Records` interface in the spec sometime next week and make sure that it's communicated properly.

LiranCohen (approved):
🚀

Issue #611, opened 2023-11-16
Allow signed ProtocolsQuery to at least return published protocol definitions


thehenrytsai:
Allow signed ProtocolsQuery to at least return published protocol definitions.

Currently signed ProtocolsQuery will error out unless it is signed by the author or a DID with a permissions grant.

thehenrytsai:
This is now done.

PR #613, opened 2023-11-16
#611 - Allowed unauthorized ProtocolsQuery to fetch published ProtocolsConfigure

thehenrytsai:


tests/handlers/protocols-query.spec.ts
```diff
@@ -117,23 +120,45 @@ export function testProtocolsQueryHandler(): void {
...
         const conditionalQueryReply = await dwn.processMessage(alice.did, conditionalQuery.message);
-
         expect(conditionalQueryReply.status.code).to.equal(200);
         expect(conditionalQueryReply.entries?.length).to.equal(1); // only 1 entry should match the query on protocol
 
         const protocolConfigured = conditionalQueryReply.entries![0] as ProtocolsConfigureMessage;
         expect(protocolConfigured).to.deep.equal(protocol2.message);
 
-        // testing fetch-all query without filter
+        // testing authenticated but unauthorized (to private ProtocolsConfigures) conditional query
```
LiranCohen, line 129: The "to private ProtocolsConfigures" part of the comment was confusing a bit for me, had to re-read it a couple of times before I understood what it meant.
Maybe something like this is clearer?

```suggestion
        // testing authenticated but unauthorized conditional query. It should return only public messages.
```

tests/handlers/protocols-query.spec.ts
```diff
@@ -117,23 +120,45 @@ export function testProtocolsQueryHandler(): void {
...
         });
 
         const fetchAllQueryReply = await dwn.processMessage(alice.did, fetchAllQuery.message);
-
         expect(fetchAllQueryReply.status.code).to.equal(200);
         expect(fetchAllQueryReply.entries?.length).to.equal(2);
         expect(fetchAllQueryReply.entries).to.deep.include(protocol2.message);
         expect(fetchAllQueryReply.entries).to.deep.include(protocol3.message);
+
+        // testing authenticated but authorized (to private ProtocolsConfigures) fetch-all query without filter
```
LiranCohen, line 152: Same comment as above, but also  thing you meant unauthorized here.
```suggestion
        // testing authenticated but unauthorized (to private ProtocolsConfigures) fetch-all query without filter
```

LiranCohen (commented):
Looks good! Just a couple of nits wrt the comments in the testing section.

Also the PR should be renamed to `ProtocolsQuery` instead of `RecordsQuery`.

tests/handlers/protocols-query.spec.ts
thehenrytsai: Fixed!

tests/handlers/protocols-query.spec.ts
thehenrytsai: Fixed!

LiranCohen (approved):

PR #617, opened 2023-11-18
#564 - Added delegated grant scoping for `RecordsWrite`

thehenrytsai:
1. Added delegated grant scoping for `RecordsWrite`.
1. Refactored code so that delegated grant evaluation can reuse the permissions grant evaluation logic.

### DIFF
--- a/tests/scenarios/delegated-grant.spec.ts
+++ b/tests/scenarios/delegated-grant.spec.ts
@@ -537,5 +537,8 @@ export function testDelegatedGrantScenarios(): void {
 
     xit('should fail if presented with a delegated grant with mismatching grant ID in the payload of the message signature', async () => {
     });
+
+    xit('should fail if presented with a revoked delegated grant', async () => {
+    });
   });
 }
--- a/tests/scenarios/delegated-grant.spec.ts
+++ b/tests/scenarios/delegated-grant.spec.ts
@@ -57,7 +57,7 @@ export function testDelegatedGrantScenarios(): void {
       await dwn.close();
     });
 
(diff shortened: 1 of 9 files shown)


tests/scenarios/delegated-grant.spec.ts
```diff
@@ -464,7 +464,60 @@ export function testDelegatedGrantScenarios(): void {
...
     });
 
-    xit('should evaluate scoping correctly when invoking a delegated grant to write', async () => {
+    it('should fail if delegated grant has a mismatching protocol scope - write', async () => {
+      // scenario:
+      // 1. Alice creates a delegated grant for device X to act as her for a protocol that is NOT email protocol
+      // 2. Bob has email protocol configured for his DWN
+      // 3. Device X attempts to use the delegated grant to write an email to Bob as Alice
+      // 4. Bob's DWN should reject Device X's message
+      const alice = await DidKeyResolver.generate();
```
LiranCohen, line 473: Clarity:

Maybe adding a comment that the protocol allows anyone can write, or even showing the control test of Alice being able to write could be helpful.

LiranCohen (approved):
Looks good, only comment I have is regarding the failed delegated test, I think maybe just a comment would be good to denote that Alice would be able to write, since I'm pretty sure some positive tests exist above it somewhere.

tests/scenarios/delegated-grant.spec.ts
thehenrytsai: Updated the comment.

LiranCohen (approved):

PR #619, opened 2023-11-20
Add `author` to `RecordsFilter`

LiranCohen:
We already index the `author` property for `RecordsWrite` but we currently don't expose it as a filter.


csuwildcat (dismissed):

frankhinek:
@LiranCohen Shouldn't we add tests that use this new filter property?

LiranCohen:
@frankhinek yup! Missed that.

thehenrytsai (approved):

PR #621, opened 2023-11-21
Remove Nested Object Indexing

LiranCohen:
This cleans up a lot of the types used during indexing/filtering. Wanted to make this its own PR so that it's not as messy with the current review of the indexing/EventsQuery PR.

For the time being we aren't doing any indexing of nested objects, we also would probably be better off flattening anything necessary while constructing the indexes/filters themselves.


src/types/query-types.ts
```diff
@@ -10,7 +10,9 @@ export enum SortDirection {
   Ascending = 1
 }
 
-export type EqualFilter = string | number | boolean;
```
thehenrytsai, line 13: I see the intent of wanting to be more specific with indexed value types. But `FilterIndex` seems to vague of an abstraction name for this, code just doesn't read well with it.

A alternative that shows the intent could be to use this type instead:

`type Indexes = { [key: string]: string | number | boolean };`

Or more generically:

`type KeyValues = { [key: string]: string | number | boolean };`

Or have both being the same type like what you've done for `EqualFilter` and `FilterIndex` and use the "right" one in their context.

LiranCohen: 💯  This is much better, fixed.

thehenrytsai (approved):

PR #625, opened 2023-11-22
EventsQuery interface and LevelIndex enhancements.

LiranCohen:
This PR contains a new interface method, `EventsQuery`, which is primarily used to query for filtered events in order to enable SelectiveSync.

### **EventLogLevel:**
EventLogLevel now takes an additional property `indexes` to it's `append()` function. These indexes are key-value indexes in the same format that we use withe the MessageStore index.

### **IndexLevel:**
To enable this functionality `IndexLevel` has been enhanced to become a more general purpose LevelDB Index that supports sorting as well as a cursor pointer property.


src/handlers/permissions-grant.ts
```diff
@@ -31,22 +34,44 @@ export class PermissionsGrantHandler implements MethodHandler {
...
+    if (scope.interface === DwnInterfaceName.Records) {
+      const { protocol, protocolPath, schema, contextId } = scope as RecordsPermissionScope;
+      indexes = {
+        ...indexes,
+        contextId,
+        protocol,
+        protocolPath,
+        schema,
+      };
+    }
```
thehenrytsai, line 72: Pending decision on method specific filters vs generic filters vs ...

src/handlers/records-delete.ts
```diff
@@ -117,21 +119,28 @@ export class RecordsDeleteHandler implements MethodHandler {
...
 
-  return indexes;
-}
+  /**
+  * Indexed properties needed for MessageStore indexing.
+  */
+  static constructIndexes(recordsDelete: RecordsDelete, recordsWrite: RecordsWriteMessage): Record<string, string> {
+    const message = recordsDelete.message;
+    const descriptor = { ...message.descriptor };
+    const { protocol, protocolPath, recipient, schema, parentId, dataFormat, dateCreated } = recordsWrite.descriptor;
```
thehenrytsai, line 129: Can we add a comment here describing why these properties? I assume it's because these are the immutable properties?

src/utils/records.ts
```diff
@@ -263,6 +265,11 @@ export class Records {
       filterCopy.datePublished = datePublishedFilter;
     }
 
+    //if sorting by published, results must filter for published
+    if (dateSort === DateSort.PublishedAscending || dateSort === DateSort.PublishedDescending) {
+      filterCopy.published = true;
+    }
```
thehenrytsai, line 271: This check feels strange to be part of "convertFilter", consider having it else where.

src/types/event-types.ts
```diff
@@ -1,12 +1,19 @@
...
 import type { GenericMessageReply } from '../core/message-reply.js';
+import type { RecordsFilter } from './records-types.js';
 import type { AuthorizationModel, GenericMessage } from './message-types.js';
 import type { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
 
+export type EventsFilter = RecordsFilter & {
+  /** optional array of methods to filter */
+  method?: string[];
+  /** optional array of interfaces to filter */
+  interface?: string[];
```
thehenrytsai, line 10: Pending design decision.

src/interfaces/events-query.ts
```diff
@@ -0,0 +1,62 @@
...
+    return new EventsQuery(message);
+  }
+
+  private static normalizeFilters(filters: EventsFilter[]): EventsFilter[] {
+    // currently all normalization filters are shared with `Records`.
+    return filters.map(filter => Records.normalizeFilter(filter));
+  }
+
+  /**
+   *  Converts an incoming EventsFilter into a Filter usable by EventLog.
```
thehenrytsai, line 53: Incorrect documentation, incoming filters is an array.

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
+
+    // create an index entry for each property index
+    // these indexes are all sortable lexicographically.
+    for (const indexName in indexes) {
+      const indexValue = indexes[indexName];
+      // the key is indexValue followed by the itemId as a tie-breaker.
+      // for example if the property is messageTimestamp the key would look like:
+      // '"2023-05-25T18:23:29.425008Z"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
+      const key = IndexLevel.keySegmentJoin(IndexLevel.encodeValue(indexValue), itemId);
+      const item: IndexedItem = { itemId, indexes };
```
thehenrytsai, line 90: Guess I will find out as I read on why indexes need to be included for every index write/push. Immediate reaction assuming I am reading it correctly is that this seems rather heavy. e.g. for every RecordsWrite, we could be looking at close to 20 key-value pairs stored repetitively ~20 times.

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
+   * If a cursor is provided it starts the iteration at the cursor point.
+   */
+  async sortedIndexQuery(tenant: string, filters: Filter[], queryOptions: QueryOptions, options?: IndexLevelOptions): Promise<string[]> {
+    const { limit, cursor , sortProperty } = queryOptions;
+    const matches: string[] = [];
+
+    // if there is a cursor we fetch the starting key given the sort property, otherwise we start from the beginning of the index.
+    const startKey = cursor ? await this.getStartingKeyForCursor(tenant, cursor, sortProperty, filters) : '';
+    if (startKey === undefined) {
+      // getStartingKeyForCursor returns undefined if an invalid cursor is provided, we return an empty result set.
```
thehenrytsai, line 220: Should we throw an error instead of resulting empty array?

src/store/index-level.ts
```diff
@@ -278,29 +528,20 @@ export class IndexLevel {
...
+   * NOTE: we currently only use this for strings, numbers and booleans.
    */
-  private static delimiter = `\x00`;
-  private join(...values: unknown[]): string {
-    return values.join(IndexLevel.delimiter);
+  static encodeValue(value: string | number | boolean): string {
+    switch (typeof value) {
+    case 'string':
+      // We can't just `JSON.stringify` as that'll affect the sort order of strings.
+      // For example, `'\x00'` becomes `'\\u0000'`.
```
thehenrytsai, line 539: Should we remove? This comment doesn't make sense to me as we don't deal with \x00 here. Maybe lost meaning due to iterations of refactoring?

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
+    if (sortIndexes === undefined) {
+      // invalid itemId
+      return;
+    }
+
+    const sortValue = sortIndexes[property];
+    if (sortValue === undefined) {
+      // invalid sort property
       return;
     }
```
thehenrytsai, line 282: Should we throw instead?

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
-    for (const propertyName in indexes) {
-      const propertyValue = indexes[propertyName];
-      const key = this.join(propertyName, this.encodeValue(propertyValue), dataId);
-      ops.push({ type: 'del', key });
+    // we find the cursor point and only return the result starting there + the limit.
+    // if there is no cursor index, we just start in the beginning.
+    const cursorIndex = cursor ? sortedValues.findIndex(match => match.itemId === cursor) : -1;
+    if (cursor !== undefined && cursorIndex === -1) {
+      // if a cursor is provided but we cannot find it, we return an empty result set
+      return [];
```
thehenrytsai, line 327: Should we return an error instead?

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
+      const propertyFilter = searchFilter[propertyName];
+      // We will find the union of these many individual queries later.
+      if (FilterUtility.isEqualFilter(propertyFilter)) {
+        // propertyFilter is an EqualFilter, meaning it is a non-object primitive type
+        const exactMatchesPromise = this.filterExactMatches(tenant, propertyName, propertyFilter, levelOptions);
+        filterPromises.push(exactMatchesPromise);
+      } else if (FilterUtility.isOneOfFilter(propertyFilter)) {
+        // `propertyFilter` is a OneOfFilter
+        // Support OR matches by querying for each values separately,
+        // then adding them to the promises associated with `propertyName`
```
thehenrytsai, line 360: Outdated documentation?

src/store/index-level.ts
LiranCohen: Yeah I think we can throw for both this and the invalid cursor.

src/store/message-store-level.ts
```diff
@@ -92,143 +91,58 @@ export class MessageStoreLevel implements MessageStore {
...
-   * If sorting is based on date published, records that are not published are filtered out.
-   * @param messages - Messages to be sorted if dateSort is present
-   * @param sort - Sorting scheme
-   * @returns Sorted Messages
-   */
-  public static async sortMessages(
-    messages: GenericMessage[],
-    messageSort: MessageSort = { }
-  ): Promise<GenericMessage[]> {
+  static getQueryOptions(messageSort: MessageSort = {}, pagination: Pagination = {}): QueryOptions {
```
thehenrytsai, line 119: Nit: consider rename to `buildQueryOptions` or `createQueryOptions`

src/store/message-store-level.ts
```diff
@@ -244,11 +158,15 @@ export class MessageStoreLevel implements MessageStore {
   async put(
     tenant: string,
     message: GenericMessage,
-    indexes: { [key: string]: string | boolean },
+    indexes: KeyValues,
     options?: MessageStoreOptions
   ): Promise<void> {
     options?.signal?.throwIfAborted();
 
+    if (indexes.messageTimestamp === undefined) {
```
thehenrytsai, line 166: For my understanding: why such a check? Doesn't every message have `messageTimestamp` already?

src/store/index-level.ts
```diff
@@ -1,213 +1,435 @@
...
 
-  async clear(): Promise<void> {
-    return this.db.clear();
+  /**
+   * Execute a filtered query against a single filter and return all results.
+   */
+  private async executeSingleFilterQuery(
+    tenant: string,
+    searchFilter: Filter,
+    matchFilters: Filter[],
```
thehenrytsai, line 342: Design: love the direction that you are going with a mix of db query and in-memory matches. Though it's odd here that the entire OR filters (`matchFilters`) "bleed" into every individual execution of each `searchFilter`.

Does it make sense to treat individual filters in the OR filters truly independently when executing them? We can still have the `searchFilter` but just singular `matchFilter` and have the filter properties being mutually exclusive for intent clarity.

This is probably related to the inner workings of `FilterSelector` and I have followed your instruction to not look at it at all, but maybe it's finally the right time to bring it into discussions?

LiranCohen: Good call out here, I was thinking there are a few different directions that we can go with optimizing this.

The one-to-one searchFilter/matchFilter is interesting, but I think it makes the `cursor` starting point a bit more complicated. The supplied `cursor` could have come exclusively from one of the `searchFilter->matchFilter` pair.

I was also thinking we could go the opposite way by attempting to bring down an array of `Filter` into a singular `Filter`. Reducing down any common fields into the existing `OneOfFilter`, the bigger change would be creating something like a `OneOfRange` to handle OR ranges (not in use practically in code yet).

So there are definitely a number of enhancements that could happen here, as well as `FilterSelector` itself.

Happy to discuss `FilterSelector` with you, basically I'd like to come up with a good starting place for some **rules** that it follows in order to take multiple filters, and reduce them into a common set of searchFilters, and in certain scenarios fall back to returning an empty filter array, which would go to db-query.

These rules can change over time as we find better heuristics... I was kind of playing around with some ideas but don't love on where I landed.

src/store/message-store-level.ts
LiranCohen: Yes, and I don't mind removing it if you think we don't need it here.

I thought it would be a good check since `MessageStore` is it's own interface, but it's currently checked in the various schema validators.

src/interfaces/events-query.ts
```diff
@@ -0,0 +1,62 @@
...
+    return new EventsQuery(message);
+  }
+
+  private static normalizeFilters(filters: EventsFilter[]): EventsFilter[] {
+    // currently all normalization filters are shared with `Records`.
+    return filters.map(filter => Records.normalizeFilter(filter));
+  }
+
+  /**
+   *  Converts an incoming array of EventsFilter into a Filter usable by EventLog.
```
thehenrytsai, line 53:
```suggestion
   *  Converts an incoming array of EventsFilter into an array of Filter usable by EventLog.
```

src/store/index-level.ts
thehenrytsai: Discussed offline, would be nice to confirm it being too costly to look up the indexes for every messageCid, if it's not "too" costly, it's probably worth just doing the lookup and avoid storing the same indexes again and again.

But we can definitely investigate in a separate PR if it's not low hanging fruit.

src/store/message-store-level.ts
thehenrytsai: I guess I am trying to understand why `messageTimestamp` is more special that we are making explicit check for it? I mean why don't we check for say `interface`, `method`, `author` etc?

If no more special, then we can just remove, else you'll need to write test for it (or you already have!)

src/store/index-level.ts
thehenrytsai: Bounced ideas offline, will review again in the next iteration.

tests/handlers/records-query.spec.ts
```diff
@@ -1468,10 +1585,31 @@ export function testRecordsQueryHandler(): void {
...
+            const compareB = messageB.descriptor[compareProperty] as string;
+
+            if (compareA > compareB) {
+              return 1;
+            } else if (compareA < compareB) {
+              return -1;
+            } else {
+              return lexicographicalCompare(messageIdA, messageIdB);
+            }
+          };
```
thehenrytsai, line 1607: Sanity: could `Message.compareMessageTimestamp()` be used here instead?

LiranCohen:
@thehenrytsai 

I've updated this PR based on our discussions and the comments left.

I've separated out the logic for choosing a paging strategy, we now have `shouldQueryWithInMemoryPaging` to handle that decision.

And for in-memory paging I've introduced the `FilterSelector.reduceFilters` method to reduce the incoming filters into a more efficient set of filters.

I'd like to have some discussion regarding typing for `EventsQueryFilter` to get some input on improving that, I have a couple of ideas there.

For throwing an error for an invalid `sortProperty` rather than returning empty results, I'd like to do that in a separate PR as it has some edge cases I'd like to properly address that will be easier to review in that context. I'll open an issue and do that one and cursor together.

src/utils/records.ts
thehenrytsai: Discussed offline, @LiranCohen tried to pull it out elsewhere but it didn't make much sense on it's own. Since it's converting the types the same thing would need to happen in several places after the conversion (or adding to the input).

src/store/index-level.ts
thehenrytsai: Separate PR it is.

thehenrytsai: Filed a bug https://github.com/TBD54566975/dwn-sdk-js/issues/631 to track this.

src/store/index-level.ts
```diff
@@ -1,213 +1,434 @@
...
+   *
+   * @param filters Array of filters that are treated as an OR query.
+   * @param queryOptions query options for sort and pagination, requires at least `sortProperty`. The default sort direction is ascending.
+   * @param options IndexLevelOptions that include an AbortSignal.
+   * @returns {string[]} an array of itemIds that match the given filters.
+   */
+  async query(tenant: string, filters: Filter[], queryOptions: QueryOptions, options?: IndexLevelOptions): Promise<string[]> {
+
+    // check if we should query using in-memory paging or iterator paging
+    if (IndexLevel.shouldQueryWithInMemoryPaging(filters, queryOptions)) {
```
thehenrytsai, line 199: Hey @LiranCohen. I sanity overrode the `shouldQueryWithInMemoryPaging()` to always return `true`, just to test that things work regardless of the query strategy used, and 32 tests blew up.

I debugged a little and found that it seems that at least some tests are not passing any filter which caused the failures.

I think should we ether fixed the tests to have at least one filter so that the tests are not dependent on query strategy to pass, or change the code to accept no filter. What do you think?

We should probably optionally add a regression test to override `shouldQueryWithInMemoryPaging()` to make sure the test works on both strategy.

thehenrytsai: Created a PR to address this: https://github.com/TBD54566975/dwn-sdk-js/pull/632, let me know what you think!

LiranCohen: 🔥

src/store/index-level.ts
thehenrytsai: Added https://github.com/TBD54566975/dwn-sdk-js/issues/633 to track.

src/store/index-level.ts
thehenrytsai: Added https://github.com/TBD54566975/dwn-sdk-js/issues/633 to track.

Issue #628, opened 2023-11-28
Allow `RecordsWrite` to be written without data not just as a privileged action


thehenrytsai:

thehenrytsai:
This is now done I believe by #629, @LiranCohen feel free to reopen if not!

PR #629, opened 2023-11-28
Allow initial RecordsWrite without data to be written to the DWN.

LiranCohen:
There are some cases where a DWN might have an initial `RecordsWrite` message used to create a record, but not have the data associated with it.

Until now this was only possible through a special method on the DWN `synchronizePrunedInitialRecordsWrite` which was only used for sync. However, there are multi-participant scenarios where this same behavior is necessary.

This PR exposes that functionality to the general processing of `RecordsWrite` messages.

If a `RecordsWrite` message is supplied without any data, it is checked that it is the initial `RecordsWrite` and indexes it as not being the latest state, this prevents it from being queryable without any data until a subsequent `RecordsWrite` with data is available.

Addresses: https://github.com/TBD54566975/dwn-sdk-js/issues/628

### DIFF
--- a/src/core/dwn-error.ts
+++ b/src/core/dwn-error.ts
@@ -117,6 +117,7 @@ export enum DwnErrorCode {
   RecordsWriteMissingAuthorizationSigner = 'RecordsWriteMissingAuthorizationSigner',
   RecordsWriteMissingSigner = 'RecordsWriteMissingSigner',
   RecordsWriteMissingDataInPrevious = 'RecordsWriteMissingDataInPrevious',
+  RecordsWriteMissingEncodedDataInPrevious = 'RecordsWriteMissingEncodedDataInPrevious',
   RecordsWriteMissingDataAssociation = 'RecordsWriteMissingDataAssociation',
   RecordsWriteMissingDataStream = 'RecordsWriteMissingDataStream',
   RecordsWriteMissingProtocol = 'RecordsWriteMissingProtocol',
--- a/src/core/dwn-error.ts
+++ b/src/core/dwn-error.ts
@@ -114,7 +114,6 @@ export enum DwnErrorCode {
   RecordsWriteGetEntryIdUndefinedAuthor = 'RecordsWriteGetEntryIdUndefinedAuthor',
   RecordsWriteGetInitialWriteNotFound = 'RecordsWriteGetInitialWriteNotFound',
   RecordsWriteImmutablePropertyChanged = 'RecordsWriteImmutablePropertyChanged',
-  RecordsWriteMissingAuthorizationSigner = 'RecordsWriteMissingAuthorizationSigner',
   RecordsWriteMissingSigner = 'RecordsWriteMissingSigner',
   RecordsWriteMissingDataInPrevious = 'RecordsWriteMissingDataInPrevious',
   RecordsWriteMissingEncodedDataInPrevious = 'RecordsWriteMissingEncodedDataInPrevious',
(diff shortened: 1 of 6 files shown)


src/core/dwn-error.ts
```diff
@@ -110,9 +110,9 @@ export enum DwnErrorCode {
   RecordsWriteGetEntryIdUndefinedAuthor = 'RecordsWriteGetEntryIdUndefinedAuthor',
   RecordsWriteGetInitialWriteNotFound = 'RecordsWriteGetInitialWriteNotFound',
   RecordsWriteImmutablePropertyChanged = 'RecordsWriteImmutablePropertyChanged',
-  RecordsWriteMissingAuthorizationSigner = 'RecordsWriteMissingAuthorizationSigner',
```
LiranCohen, line 113: Removed this as i noticed it's not being used anywhere.

src/handlers/records-write.ts
```diff
@@ -146,77 +137,109 @@ export class RecordsWriteHandler implements MethodHandler {
...
+      // we preform this check in case a user attempts to gain access to data by knowing the dataCid,
+      // https://github.com/TBD54566975/dwn-sdk-js/issues/359
+      // so we insure that the data is already associated with the existing newest message
+      const dataResults = await this.dataStore.get(tenant, previousWriteMessageCid, message.descriptor.dataCid);
+      if (dataResults === undefined) {
+        throw new DwnError(
+          DwnErrorCode.RecordsWriteMissingDataInPrevious,
+          `No dataStream was provided and unable to get data from previous message`
+        );
+      }
```
thehenrytsai, line 214: Sanity: I am wondering why would this condition be true in real life? The newest existing RecordsWrite should have the data right?

thehenrytsai (commented):
Added some comments, and sent a PR to address some of the comments:
https://github.com/TBD54566975/dwn-sdk-js/pull/635

src/handlers/records-write.ts
```diff
@@ -92,46 +87,42 @@ export class RecordsWriteHandler implements MethodHandler {
...
-      if (dataStream === undefined && newestExistingMessage?.descriptor.method === DwnMethodName.Delete) {
-        return messageReplyFromError(new DwnError(DwnErrorCode.RecordsWriteMissingDataStream, 'No data stream was provided with the previous message being a delete'), 400);
+    try {
+      // we set latestBaseState initially to false, due to the fact that we allow `RecordsWrite` without a data stream in some cases.
+      // depending on rules within `processMessageWithDataStream` or `processMessageWithoutDataStream` it will be set to true to allow for query/reads
+      let isLatestBaseState = false;
+      let messageWithOptionalEncodedData = message as RecordsWriteMessageWithOptionalEncodedData;
+
+      if (dataStream !== undefined) {
+        ({ isLatestBaseState, messageWithOptionalEncodedData } = await this.processMessageWithDataStream(tenant, message, dataStream));
```
thehenrytsai, line 97: Optional: I think the `processMessageWithDataStream()` and `processMessageWithoutDataStream()` are vaguely named with prefix "process" is because it attempt to handle both the size greater 30K and under cases. I couldn't come up with better names myself! But if we were to split the logic into separate function based on that condition, the function name could be way clear but I can see why one might think it's refactoring not worth taking. So this is optional.

LiranCohen: Yeah I was considering going the other direction wrt encoded data vs store data paths... but ultimately chose this path.

Potentially refactoring this down the road could be more clear depending on what additional rules/features touch these paths.

thehenrytsai (approved):
👍 💯

PR #632, opened 2023-11-29
Modified code such that in-memory-paging supports empty filter and array

thehenrytsai:
1. Modified code such that in-memory-paging supports empty filter and array, so that queries (and tests) can work regardless of the query strategy (true black-box).
2. Removed the need for FilterSelector. 

Disclaimer: I didn't read `FilterSelector` deeply but I think I got the intent that you wanted to minimize the total number DB queries (let me know if I understood wrong)! But I am not convinced that it's worth it for the amount of code introduced, so I took your general intent and made some tweaks, and essentially got rid most functions in FilterSelector.

Took me a long time to debug and make all queries work for in-memory paging strategy, but I took a stab at showing what I had in mind (which is pretty dumb), let me know what you think. 

I have not bothered with writing code coverage and making this super tidy anticipating that you might not like it or there's problem with it, it does demonstrate my intent and should be mergeable to be worked on further otherwise!


LiranCohen (approved):
>Disclaimer: I didn't read FilterSelector deeply but I think I got the intent that you wanted to minimize the total number DB queries (let me know if I understood wrong)! But I am not convinced that it's worth it for the amount of code introduced, so I took your general intent and made some tweaks, and essentially got rid most functions in FilterSelector.

Spot on... the reduction of code and legibility here is a definite huge win so I'm all for it. Especially considering the in-practice gains would really be made for non-tenant queries which would happen on a remote (hopefully running SQL or something more efficient than level).

I'll merge and get coverage up to par on the other branch.

PR #634, opened 2023-11-30
`messageCid` cursor & reverse index kept around during delete

LiranCohen:
There can be a scenario where a user attempts to use a `messageCid` as a cursor for a message after it's been deleted.

In order to remedy this, we keep the reverse index lookup for the items when issuing a `delete`. Then when we have a query with a cursor, we can look up the indexes, and use the item as a cursor point for the in-memory pagination.

I added a `purge` method that truly purges all indexes + reverse, although we never use it in practice (yet).  I think I should probably remove it and we can always add it back later as it's simple.

### DIFF
--- a/src/event-log/event-log-level.ts
+++ b/src/event-log/event-log-level.ts
@@ -64,13 +64,4 @@ export class EventLogLevel implements EventLog {
 
     await Promise.all(indexDeletePromises);
   }
-
-  async purgeEventsByCid(tenant: string, messageCids: Array<string>): Promise<void> {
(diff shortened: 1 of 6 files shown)


LiranCohen:
Discussed with @thehenrytsai

Will open issue to address this immediately after the other large merge instead of this band-aid.

The proposed solution would be to use the `sortProperty` + `sortPropertyValue` + `dataCid` as a cursor that could be cross implementation and not require storing additional data that's not necessary.

This will also move us towards removing the need for `EventLog` and use `MessageStore` as the necessary source of truth, reducing indexing needs.

PR #638, opened 2023-12-03
EventLog / MessageStore cursor should not be a `messageCID`

LiranCohen:
Prior to this PR we used `messageCid` as a cursor for paginating the `MessageStore`, as well as `EventLog`.

This generally works well since all messages throughout the system can be identified by a `messageCid`, however it becomes tricky if a `messageCid` which was returned as a pagination cursor, was no longer in the `MessageStore` by the time the subsequent call came in using that `messageCid` as a cursor.

To remedy this the `cursor` is now a string which represents the `value` and `itemId` of the last item returned from a query. The `value` is the value of the `sortProperty` key in the item's indexes.

This allows us to construct a sorted point in which to either start iterating through (gt iterator in level), or an index to slice the in-memory paging results.

Additionally the `IndexLevel` query nos returns the entire `IndexedItem` so that the caller can build a cursor from the item returned. In `MessageStore` we query 1 additional record in order to check if there are additional results, so the cursor is sometimes built from a different item than the last.


src/types/query-types.ts
```diff
@@ -49,4 +49,9 @@ export type RangeCriterion = {
    * Inclusive end date-time.
    */
   to?: string;
+};
+
+export type PaginationCursor = {
+  itemId: string;
```
thehenrytsai, line 55: Any reason why this cannot be named `messageCid`?

LiranCohen: Yup! And I went ahead and just removed references to `itemId` as they don't really make sense anymore.

thehenrytsai (dismissed):
🙏

csuwildcat (dismissed):

thehenrytsai (approved):
Blind approve.

PR #396, opened 2023-06-14
#315 - updated all dependencies to latest version and fixed breaking changes

thehenrytsai:

PR #643, opened 2023-12-05
Remove unnecessary conversions to hex format

frankhinek:
## Summary

This PR primarily focuses on performance improvement for the two digital signature algorithms supported by `dwn-sdk-js.

## Context

In [this Discord thread](https://discord.com/channels/937858703112155166/1181470744023146538) @shamilovtim noted significant delays while profiling [web5-wallet](https://github.com/TBD54566975/web5-wallet) performance.

During the troubleshooting process, a tangentially related issue was discovered that unnecessarily reduces the performance of `Ed25519` and `secp256k1` sign operations.  While updating `dwn-sdk-js` to the v2.0 versions of `@noble/ed5519` and `@noble/secp256k1` in PR #396, unnecessary conversions to hex format were added in the `sign()` method (but not `verify()`).

The result of this change is that `sign()` performance degrades rapidly as the data size of the content increases:

Screenshot 2023-12-05 at 12 29 40 PM

With a 100MB data payload, the sign operation in `dwn-sdk-js` is ⚠️ **40 times slower** ⚠️  than it should be.

If we switch the time axis to log scale, you can see that the verify operations both in `dwn-sdk-js` and `@noble/ed25519` track almost identically, since verify does not include the conversion to hex before verification.

Screenshot 2023-12-05 at 12 30 35 PM

## Changes

- Remove all occurrences of hex format conversions from both:
    - `src/jose/algorithms/signing/ed25519.ts`
    - `src/utils/secp256k1.ts`

After implementing these changes, the performance of the sign operation is nearly identical whether called from the `dwn-sdk-js` wrapper or directly to the `@noble` lib and scales linearly as the payload size increases:

Screenshot 2023-12-05 at 12 49 27 PM

### DIFF
--- a/src/jose/algorithms/signing/ed25519.ts
+++ b/src/jose/algorithms/signing/ed25519.ts
@@ -27,11 +27,9 @@ export const ed25519: SignatureAlgorithm = {
   sign: async (content: Uint8Array, privateJwk: PrivateJwk): Promise<Uint8Array> => {
     validateKey(privateJwk);
 
-    const contentHex = Ed25519.etc.bytesToHex(content);
     const privateKeyBytes = Encoder.base64UrlToBytes(privateJwk.d);
-    const privateKeyHex = Ed25519.etc.bytesToHex(privateKeyBytes);
 
-    return Ed25519.signAsync(contentHex, privateKeyHex);
+    return Ed25519.signAsync(content, privateKeyBytes);
   },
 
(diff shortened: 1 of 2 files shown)


csuwildcat (approved):

frankhinek:
Another +1 for migrating `dwn-sdk-js` to use the `@web5/crypto` libs 😃

thehenrytsai (approved):
This is great! Thanks!

PR #646, opened 2023-12-07
fix uint8arrays bug 

shamilovtim:
pin uint8arrays to 4.0.8 to resolve https://github.com/achingbrain/uint8arrays/issues/74

### DIFF
--- a/.github/workflows/integrity-check.yml
+++ b/.github/workflows/integrity-check.yml
@@ -59,10 +59,10 @@ jobs:
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/integrity-check.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v3
         with:
-          fail_ci_if_error: true
+          continue-on-error: true
```
shamilovtim, line 62: this was not a reasonable setting for codecov IMO

shamilovtim:
fixed upstream

PR #648, opened 2023-12-20
Align `UnionMessageReply` with the various replies that `processMessage` 

LiranCohen:
`UnionMessageReply` was not fully in sync with the actual replies.

- `data` is actually within `record` for a `RecordsReadReply`.
- modified `events` to `entries` for `EventsGetReply` and `EventsQueryReply`
- modified `messages` to `entries` for `MessagesGetReply`

This way all of the individual replies can be represented by:

```typescript
/**
 * Catch-all message reply type. It is recommended to use GenericMessageReply or a message-specific reply type wherever possible.
 */
export type UnionMessageReply = GenericMessageReply & {
  /**
   * Resulting message entries or events returned from the invocation of the corresponding message.
   * e.g. the resulting messages from a RecordsQuery, or array of messageCid strings for EventsGet or EventsQuery
   * Mutually exclusive with `record`.
   */
  entries?: QueryResultEntry[] | ProtocolsConfigureMessage[] | MessagesGetReplyEntry[] | string[];

  /**
   * Record corresponding to the message received if applicable (e.g. RecordsRead).
   * Mutually exclusive with `entries` and `cursor`.
   */
  record?: RecordsWriteReply;

  /**
   * A cursor for pagination if applicable (e.g. RecordsQuery).
   * Mutually exclusive with `record`.
   */
  cursor?: string;

  /**
   * A subscription object if a subscription was requested.
   */
  subscription?: GenericMessageSubscription;
};
```

### DIFF
--- a/src/core/message-reply.ts
+++ b/src/core/message-reply.ts
@@ -24,8 +24,8 @@ export function messageReplyFromError(e: unknown, code: number): GenericMessageR
 export type UnionMessageReply = GenericMessageReply & {
   /**
    * Resulting message entries or events returned from the invocation of the corresponding message.
-   * e.g. the resulting messages from a RecordsQuery
-   * Mutually exclusive with `data`.
+   * e.g. the resulting messages from a RecordsQuery, or array of messageCid strings for EventsGet or EventsQuery
+   * Mutually exclusive with `record`.
    */
   entries?: QueryResultEntry[] | MessagesGetReplyEntry[] | string[];
 
@@ -37,7 +37,7 @@ export type UnionMessageReply = GenericMessageReply & {
 
   /**
    * A cursor for pagination if applicable (e.g. RecordsQuery).
-   * Mutually exclusive with `data`.
+   * Mutually exclusive with `record`.
    */
   cursor?: string;
 };
--- a/src/core/message-reply.ts
+++ b/src/core/message-reply.ts
@@ -1,15 +1,6 @@
 import type { MessagesGetReplyEntry } from '../types/messages-types.js';
-import type { QueryResultEntry } from '../types/message-types.js';
 import type { RecordsWriteReply } from '../types/records-types.js';
-
-type Status = {
-  code: number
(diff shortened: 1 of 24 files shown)


src/core/message-reply.ts
```diff
@@ -23,20 +16,20 @@ export function messageReplyFromError(e: unknown, code: number): GenericMessageR
...
-  entries?: QueryResultEntry[];
+  entries?: QueryResultEntry[] | ProtocolsConfigureMessage[] | MessagesGetReplyEntry[] | string[];
 
   /**
-   * Data corresponding to the message received if applicable (e.g. RecordsRead).
+   * Record corresponding to the message received if applicable (e.g. RecordsRead).
    * Mutually exclusive with `entries` and `cursor`.
    */
-  data?: Readable;
+  record?: RecordsWriteReply;
```
thehenrytsai, line 28: Design: shouldn't this be called `RecordsReadReply`?

Even so it would still be odd because `Reply` types should ideally be of the same level as all other `Reply` types, not be a type for a nested property.

Can we do something like directly:

```suggestion
  record?: RecordsWriteMessage & {
    /**
     * The initial write of the record if the returned RecordsWrite message itself is not the initial write.
     */
    initialWrite?: RecordsWriteMessage;
    data: Readable;
  }
```

LiranCohen: Yeah I agree it's weird... It's because `RecordsReadReply` has a nested `record: RecordsWriteReply` as a property 😅 

Maybe this is an opportunity to rename `RecordsWriteReply`? It almost feels like it's just a `Record`, but I'm not sure if that's too general.

I'm ok with just making it explicit within the `UnionMessageReply` for now though.

Thoughts?

thehenrytsai: I am also okay with just declaring the structure (like shown in the above) if that's what you mean?

thehenrytsai (approved):

PR #649, opened 2023-12-22
Export `TestDataGenerator` class

LiranCohen:
I've found myself wanting to reach for this when trying to create benchmarking in an external repo, and potentially useful within web5-js (I think they duplicate much of this class though).

A community member is also looking to create some benchmarking tools and was asking for this to be exported.


thehenrytsai (approved):
May be a foot gun but so be it...!

LiranCohen:
@thehenrytsai hopefully not! I just think we need to be careful if/when we decide to use this in `web5-js`. Maybe when that time comes some refactoring would be done to split some duplicated stuff into their own classes. 🤷

PR #655, opened 2024-01-05
Renamed isTenant() to isActiveTenant() to better support intent

thehenrytsai:
Upon adding support tenant registration and tenant gate in `dwn-server`, it occurred to me that `isTenant()` is not a perfect fit, because they can be a tenant, but have agreed to an old/expired terms-of-service, in which case they still should be denied access.

`isActiveTenant()` address the intent better.

All ears if you have an even better word than "active".


LiranCohen (approved):
Yeah that makes sense. ✅

PR #658, opened 2024-01-10
EventsSubscribe

LiranCohen:


json-schemas/interface-methods/events-subscribe.json
```diff
@@ -0,0 +1,47 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://identity.foundation/dwn/json-schemas/events-subscribe.json",
+    "type": "object",
+    "additionalProperties": false,
+    "required": [
+      "descriptor"
```
thehenrytsai, line 7: Sanity: `authorization` is required also no?

LiranCohen: Yeah it should be.

I think I initially looked at `records-query.json` as a scaffold. It also doesn't exist there, is that on purpose?
https://github.com/TBD54566975/dwn-sdk-js/blob/b3a75778bcae8bbe0acc4168e8655d09a5415477/json-schemas/interface-methods/records-query.json#L6

src/types/subscriptions.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { GenericMessage, GenericMessageHandler, GenericMessageSubscription } from './message-types.js';
+
+export type EmitFunction = (tenant: string, message: GenericMessage, indexes: KeyValues) => void;
+
+/**
+ * The EventStream interface implements a pub/sub system based on Message filters.
+ */
+export interface EventStream {
+  subscribe(tenant: string, message: EventsSubscribeMessage, filters: Filter[], messageStore: MessageStore): Promise<EventsSubscription>;
+  subscribe(tenant: string, message: GenericMessage, filters: Filter[], messageStore: MessageStore): Promise<GenericMessageSubscription>;
```
thehenrytsai, line 15: For my understanding: why would we need the `GenericMessage` version of the `subscribe()` method?

LiranCohen: I can remove it from this PR, but this is because I was overloading the method with a subscription signature for `RecordsSubscribeMessage`.

In the original PR I made a comment that I am 50/50 on the abstraction and overloading vs just having `GenericMessage` and a generic handler. I see pluses for either argument. Would be happy to chat about that if you'd like.

The only current difference between the two returned objects is that the handler for `RecordsSubscription` will emit `RecordsWriteMessage` and `RecordsDeleteMessage` only vs `GenericMessage`.

json-schemas/interface-methods/events-subscribe.json
thehenrytsai: Yes, that's on purpose for `RecordsQuery`, since it can be an anonymous query.

src/event-log/event-stream.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  async close(): Promise<void> {
+    this.isOpen = false;
+    this.eventEmitter.removeAllListeners();
+  }
+
+  emit(tenant: string, message: GenericMessage, indexes: KeyValues): void {
+    if (!this.isOpen) {
+      // silently ignore.
```
thehenrytsai, line 91: Throwing error seems to be better, else the caller would just assume the event is emitted?

src/event-log/subscription.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export class SubscriptionHandlerBase implements SubscriptionHandler {
+  protected eventEmitter: EventEmitter;
+  protected messageStore: MessageStore;
+  protected filters: Filter[];
+  protected tenant: string;
+  protected message: GenericMessage;
+
+  #unsubscribe: () => Promise<void>;
+  #id: string;
```
thehenrytsai, line 21: The use of # syntax was once undone when I started using it (https://github.com/TBD54566975/dwn-sdk-js/pull/466), while requirements may have changed, I'd use the _ prefix instead even if just for the sake of consistency. Also, support for it in VSC debugging isn't quite seamless yet.

src/event-log/subscription.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.eventEmitter.off(this.eventChannel, handler);
+      }
+    };
+  }
+
+  onError(handler: (error: DwnError) => void): void {
+    this.eventEmitter.on(this.errorEventChannel, handler);
+  }
+
+  async close(): Promise<void> {
```
thehenrytsai, line 79: For my understanding: Where is being invoked? If not invoked, this method can be removed, along with `#unsubscribe`.

src/event-log/subscription.ts
thehenrytsai, line 1: Rename to match with the exported class: `subscription-handler-base.ts`. Honestly, as the files grew, I'd prefer just naming the file exactly the same as the class being exported.... ie. PascalCasing. the cabab-casing was convention that was handed to me.

src/event-log/event-stream.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  constructor(config?: EventStreamConfig) {
+    this.reauthorizationTTL = config?.reauthorizationTTL || 0; // if set to zero it does not reauthorize
+
+    // we capture the rejections and currently just log the errors that are produced
+    this.eventEmitter = config?.emitter || new EventEmitter({ captureRejections: true });
+  }
+
+  private get eventChannel(): string {
+    return `${eventChannel}_bus`;
```
thehenrytsai, line 35: This seems rather unnecessary, why add `_bus` and why not just a `const` or `private static readonly` and be done with it?

src/event-log/event-stream.ts
```diff
@@ -0,0 +1,100 @@
...
+import { DwnError, DwnErrorCode } from '../core/dwn-error.js';
+
+const eventChannel = 'events';
+
+type EventStreamConfig = {
+  emitter?: EventEmitter;
+  reauthorizationTTL?: number;
+};
+
+export class EventStreamEmitter implements EventStream {
```
thehenrytsai, line 20: EventStreamEmitter sounds like an Emitter of EventStreams (e.g. EventEmitter is and Emitter of Events).

Consider renaming to class to` EventStream` and rename the interface to `EventStreamInterface` or I am even onboard with `IEventStream` if you fine "Interface" too silly/verbose as a convention.

src/types/events-types.ts
```diff
@@ -41,11 +42,36 @@ export type EventsGetReply = GenericMessageReply & {
...
+  descriptor: EventsSubscribeDescriptor;
+};
+
+export type EventsHandler = (message: GenericMessage) => void;
+
+export type EventsSubscription = {
+  id: string;
+  on: (handler: EventsHandler) => { off: () => void };
+  onError: (handler: (error: DwnError) => void) => void;
+  close: () => Promise<void>;
```
thehenrytsai, line 56: Design question: What's the difference between `close()` and `off: () => void`? Is there a possibility to condense down to one?

src/event-log/subscription.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.eventEmitter.on(this.eventChannel, handler);
+    return {
+      off: (): void => {
+        this.eventEmitter.off(this.eventChannel, handler);
+      }
+    };
+  }
+
+  onError(handler: (error: DwnError) => void): void {
+    this.eventEmitter.on(this.errorEventChannel, handler);
```
thehenrytsai, line 76: Sanity: I am not seeing error ever being emitted? If that's true, consider removing code related to this until a later PR =P.

src/types/subscriptions.ts
```diff
@@ -0,0 +1,34 @@
+import type { DwnError } from '../core/dwn-error.js';
+import type { GenericMessageReply } from '../types/message-types.js';
+import type { MessageStore } from './message-store.js';
+import type { EventsSubscribeMessage, EventsSubscription } from './events-types.js';
+import type { Filter, KeyValues } from './query-types.js';
+import type { GenericMessage, GenericMessageHandler, GenericMessageSubscription } from './message-types.js';
+
+export type EmitFunction = (tenant: string, message: GenericMessage, indexes: KeyValues) => void;
```
thehenrytsai, line 8: Rename. Perhaps:

```suggestion
export type EventHandler/Listener = (tenant: string, message: GenericMessage, indexes: KeyValues) => void;
```

thehenrytsai (commented):
After reading the code a bit more, maybe I am overlooking something, but it seems that the design of `EventStream` can be simplified, specifically around the `subscribe()` method:

`subscribe(id: string, listener: EventListener)` or if absolutely necessary:
`subscribe(id: string, subscriptionHandler: SubscriptionHandler)`

might get the job done.

I think ideally `EventStream` should not know (or care) about who is interested in its events and what they do with them. This way the `EventsSubscribeHandler` and `RecordsSubscribeHandler` can construct the event handler however they see fit, and just tell `EventStream`: "Hey, I am interested in your events, send them to me".

Let me know what you think, happy to jump in a call tomorrow!

src/types/events-types.ts
LiranCohen: You can call the `.on()` method multiple times  with different handler functions on each, so the off that is returned will stop emitting to that specific handler. I figured it would be a better pattern to do that than add an `off()` method which must take the same exact handler.

`.close()` just closes the entire subscription no longer emitting to any handlers.

src/event-log/subscription.ts
LiranCohen: Yup! We can def remove this until the other PR which makes use of it.

LiranCohen:
> I think ideally `EventStream` should not know (or care) about who is interested in its events and what they do with them. This way the `EventsSubscribeHandler` and `RecordsSubscribeHandler` can construct the event handler however they see fit, and just tell `EventStream`: "Hey, I am interested in your events, send them to me".

I think I understand what you're getting at and it kind of makes sense. Admittedly I designed `EventStream` first and the handler portions as an after-thought when thinking about more specific cases.

Would definitely like to jump on a call and chat just to give some additional context and make sure we are on the same page.

LiranCohen:
@thehenrytsai I've got much of the refactor in place and have been thinking about what we spoke about wrt the handler function.

Would like your input on some things before I go with them as they're touching many tests.

In order to support the change to treat `handler: GenericMessageHandler` similarly to how `dataStream: Readable` is treated I propose we change the processMessage signature as follows:

Currently:
```typescript
  public async processMessage(tenant: string, rawMessage: GenericMessage, dataStream?: Readable): Promise<UnionMessageReply>;
```

Proposed:
```typescript

export type MessageOptions = {
  dataStream?: Readable;
  handler?: GenericMessageHandler
};

public async processMessage(tenant: string, rawMessage: GenericMessage, options?: MessageOptions): Promise<UnionMessageReply>;

```

If you like this approach, I think it would be worthwhile to update that signature in it's own PR first, with only `dataStream` being a part of the `MessageOptions`, as this method is touching nearly every single test file we have.

LiranCohen:
@thehenrytsai This is ready for review again. I still need to write the scenario tests for all the specific filters, but I wanted to get the review going for the general refactor.

I do think that in a later PR we will introduce the unsubscribe functionality we spoke about vs the `.close()` method, but I left it like that for now to make the reviews easier.

The logic is simplified a lot further than it was in many places. When I add `RecordsSubscribe` some things will change and bring in more complexity but we can review those more independently.

tests/event-log/event-emitter-stream.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Clean up after each test by closing and clearing the event stream
+    await messageStore.close();
+    await eventStream.close();
+  });
+
+  it('should remove listeners when unsubscribe method is used', async () => {
+    const emitter = new EventEmitter();
+    eventStream = new EventEmitterStream({ emitter });
+
+    // count the `events_bus` listeners, which represents all listeners
```
thehenrytsai, line 43: Sanity
```suggestion
    // count the `events` listeners, which represents all listeners
```

tests/event-log/event-emitter-stream.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  after(async () => {
+    console.error = originalConsoleErrorFunction;
+    // Clean up after each test by closing and clearing the event stream
+    await messageStore.close();
+    await eventStream.close();
+  });
+
+  it('should remove listeners when unsubscribe method is used', async () => {
```
thehenrytsai, line 39: Nit:
```suggestion
  it('should remove listeners when `close` method is used', async () => {
```

src/types/events-types.ts
thehenrytsai: Seems no longer applicable.

src/core/message-reply.ts
```diff
@@ -40,4 +40,9 @@ export type UnionMessageReply = GenericMessageReply & {
    * Mutually exclusive with `record`.
    */
   cursor?: PaginationCursor;
+
+  /**
+   * A subscription object if a subscription was requested.
+   */
+  subscription?: GenericMessageSubscription;
```
thehenrytsai, line 47: For my understanding: why can't there be just one interface as opposed to having both `EventSubscription` and `GenericMessageSubscription`?

src/types/events-types.ts
```diff
@@ -42,11 +42,39 @@ export type EventsGetReply = GenericMessageReply & {
...
+};
+
+export type EventsSubscribeMessage = {
+  authorization: AuthorizationModel;
+  descriptor: EventsSubscribeDescriptor;
+};
+
+export type EventsHandler = (message: GenericMessage) => void;
+
+export type EventsSubscription = {
```
thehenrytsai, line 57: Style nit: I think as a convention, when there is method declaration in the typing, we should use `interface` instead of `type, it indicates that this is more than "just" a data holder.

src/types/message-types.ts
```diff
@@ -75,6 +76,13 @@ export type QueryResultEntry = GenericMessage & {
   encodedData?: string;
 };
 
+export type GenericMessageSubscriptionHandler = (message: GenericMessage) => void;
```
thehenrytsai, line 79: The prefix seems to not add much value, consider:
```suggestion
export type SubscriptionHandler = (message: GenericMessage) => void;
```
or
```suggestion
export type MessageSubscriptionHandler = (message: GenericMessage) => void;
```

src/types/message-types.ts
```diff
@@ -16,6 +16,7 @@ export type GenericMessage = {
  */
 export type MessageOptions = {
   dataStream?: Readable;
+  handler?: GenericMessageSubscriptionHandler;
```
thehenrytsai: Consider renaming to `subscriptionHandler` to disambiguate from "message handlers" we already have.

src/types/message-types.ts
```diff
@@ -75,6 +76,13 @@ export type QueryResultEntry = GenericMessage & {
   encodedData?: string;
 };
 
+export type MessageSubscriptionHandler = (message: GenericMessage) => void;
+
+export interface GenericMessageSubscription {
```
thehenrytsai, line 81: Paired with the earlier comment, consider:
```suggestion
export interface MessageSubscription {
```

src/utils/events.ts
```diff
@@ -0,0 +1,87 @@
...
+    const messageTimestampFilter = dateUpdated ? FilterUtility.convertRangeCriterion(dateUpdated) : undefined;
+    if (messageTimestampFilter) {
+      filterCopy.messageTimestamp = messageTimestampFilter;
+      delete filterCopy.dateUpdated;
+    }
+    return filterCopy as Filter;
+  }
+
+  private static isMessagesFilter(filter: EventsFilter): filter is EventsMessageFilter {
+    return 'method' in filter || 'interface' in filter || 'dateUpdated' in filter || 'author' in filter;
```
thehenrytsai, line 70: Sanity: sorry for missing this in earlier PR, doesn't a record filter qualifies for this check too?

src/handlers/records-delete.ts
```diff
@@ -86,12 +90,15 @@ export class RecordsDeleteHandler implements MethodHandler {
     }
 
     const recordsWrite = await RecordsWrite.getInitialWrite(existingMessages);
```
thehenrytsai, line 92: Nit: let's rename this to `initialWrite` or `initialRecordsWrite`.

src/handlers/records-delete.ts
```diff
@@ -86,12 +90,15 @@ export class RecordsDeleteHandler implements MethodHandler {
...
-    const indexes = RecordsDeleteHandler.constructIndexes(recordsDelete, recordsWrite);
-    await this.messageStore.put(tenant, message, indexes);
-
+    const published = (newestExistingMessage as RecordsWriteMessage).descriptor.published;
+    const indexes = recordsDelete.constructIndexes(recordsWrite, published);
```
thehenrytsai, line 94: Design: consider the following which I think shows the intent better.

```suggestion
    const indexes = recordsDelete.constructIndexes(recordsWrite, newestExistingMessage);
```

src/handlers/records-delete.ts
```diff
@@ -86,12 +90,15 @@ export class RecordsDeleteHandler implements MethodHandler {
     }
 
     const recordsWrite = await RecordsWrite.getInitialWrite(existingMessages);
-    const indexes = RecordsDeleteHandler.constructIndexes(recordsDelete, recordsWrite);
-    await this.messageStore.put(tenant, message, indexes);
-
+    const published = (newestExistingMessage as RecordsWriteMessage).descriptor.published;
```
thehenrytsai, line 93: For my (and future reader's) understanding: can you add a comment note as to why it is critical to have the latest `published` in the index? I vague remember you mention it has something to do with selective sync needing to detect the delete or something but not 100% sure.

Though, on the related note, should we use the latest value of other mutable indexes? Such as `dataSize` and `dataCid`?

src/interfaces/records-delete.ts
```diff
@@ -82,4 +116,8 @@ export class RecordsDelete extends AbstractMessage<RecordsDeleteMessage> {
       messageStore
     });
   }
+
+  public static isRecordsDeleteMessage(message: GenericMessage): message is RecordsDeleteMessage {
```
thehenrytsai, line 120: Sanity: not used? Consider removing?

src/interfaces/records-write.ts
```diff
@@ -932,4 +932,8 @@ export class RecordsWrite implements MessageInterface<RecordsWriteMessage> {
     const attesters = attestationSignatures.map((signature) => Jws.getSignerDid(signature));
     return attesters;
   }
+
+  public static isRecordsWriteMessage(message: GenericMessage): message is RecordsWriteMessage {
```
thehenrytsai, line 936: Sanity: same question as `isRecordsDeleteMessage`. Needed?

src/index.ts
```diff
@@ -51,10 +52,14 @@ export { Signer } from './types/signer.js';
...
 
-// store interfaces
+// store implementations
 export { DataStoreLevel } from './store/data-store-level.js';
 export { EventLogLevel } from './event-log/event-log-level.js';
 export { MessageStoreLevel } from './store/message-store-level.js';
 
+// eventing implementations
+export { EventEmitterStream } from './event-log/event-emitter-stream.js';
+export { EventsSubscribe , EventsSubscribeOptions } from './interfaces/events-subscribe.js';
```
thehenrytsai, line 62: Super nit: group this export to above with other DWN interfaces?

src/interfaces/events-subscribe.ts
```diff
@@ -0,0 +1,58 @@
...
+  messageTimestamp?: string;
+  filters?: EventsFilter[]
+  permissionsGrantId?: string;
+};
+
+export class EventsSubscribe extends AbstractMessage<EventsSubscribeMessage> {
+  public static async parse(message: EventsSubscribeMessage): Promise<EventsSubscribe> {
+    Message.validateJsonSchema(message);
+    await Message.validateSignatureStructure(message.authorization.signature, message.descriptor);
+
```
thehenrytsai, line 22: Sanity: should we do more here like other similar interface methods, such as validate structure of the filter?

src/interfaces/events-subscribe.ts
```diff
@@ -0,0 +1,58 @@
...
+    Time.validateTimestamp(message.descriptor.messageTimestamp);
+    return new EventsSubscribe(message);
+  }
+
+  /**
+   * Creates a SubscriptionRequest message.
```
thehenrytsai, line 28:
```suggestion
   * Creates a EventsSubscribe message.
```

src/interfaces/events-subscribe.ts
```diff
@@ -0,0 +1,58 @@
...
+   *
+   * @throws {DwnError} when a combination of required SubscriptionRequestOptions are missing
```
thehenrytsai, line 30: Consider remove, the comment is vague, and should be `EventsSubscribeOptions`

src/interfaces/events-subscribe.ts
```diff
@@ -0,0 +1,58 @@
...
+import { removeUndefinedProperties } from '../utils/object.js';
+import { Time } from '../utils/time.js';
+import { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
+
+
+export type EventsSubscribeOptions = {
+  signer: Signer;
+  messageTimestamp?: string;
+  filters?: EventsFilter[]
+  permissionsGrantId?: string;
```
thehenrytsai, line 15: Sanity: not needed?

src/interfaces/events-subscribe.ts
```diff
@@ -0,0 +1,58 @@
...
+    const descriptor: EventsSubscribeDescriptor = {
+      interface        : DwnInterfaceName.Events,
+      method           : DwnMethodName.Subscribe,
+      filters          : options.filters ?? [],
+      messageTimestamp : options.messageTimestamp ?? currentTime,
+    };
+
+    removeUndefinedProperties(descriptor);
+
+    // only generate the `authorization` property if signature input is given
```
thehenrytsai, line 47: Remove.

src/interfaces/records-delete.ts
```diff
@@ -68,7 +70,35 @@ export class RecordsDelete extends AbstractMessage<RecordsDeleteMessage> {
   }
 
   /**
-   * Authorizes the delegate who signed this message.
+  * Indexed properties needed for MessageStore indexing.
```
thehenrytsai, line 73: Nit: indent still off by 1 char, annoying linter doesn't just fix this...

src/interfaces/records-delete.ts
```diff
@@ -68,7 +70,35 @@ export class RecordsDelete extends AbstractMessage<RecordsDeleteMessage> {
   }
 
   /**
-   * Authorizes the delegate who signed this message.
+    * Indexed properties needed for MessageStore indexing.
```
thehenrytsai, line 73: Nit: still off by 1 char, just saying... 🤣

src/utils/events.ts
```diff
@@ -0,0 +1,104 @@
...
+  // if it were the only property that matched, it could be handled by `EventMessageFilter`
+  private static isRecordsFilter(filter: EventsFilter): filter is EventsRecordsFilter {
+    return 'author' in filter ||
+      'dateCreated' in filter ||
+      'dataFormat' in filter ||
+      'dataSize' in filter ||
+      'parentId' in filter ||
+      'recordId' in filter ||
+      'schema' in filter ||
+      ('protocolPath' in filter && 'protocol' in filter) ||
```
thehenrytsai, line 97: Sanity: I probably forgot, do you recall what if only `protocolPath` is given?

tests/test-event-stream.ts
```diff
@@ -0,0 +1,29 @@
+import type { EventStream } from '../src/index.js';
+
+import { EventEmitterStream } from '../src/index.js';
+
+/**
+ * Class that manages store implementations for testing.
```
thehenrytsai, line 6: Forgot to update the C&P comments, throughout this file.

tests/utils/test-data-generator.ts
```diff
@@ -766,6 +780,30 @@ export class TestDataGenerator {
...
+    removeUndefinedProperties(options);
+
+    const eventsSubscribe = await EventsSubscribe.create(options);
+    const message = eventsSubscribe.message;
+
+    return {
+      author,
+      eventsSubscribe,
+      message
+    };
```
thehenrytsai, line 804: For consideration: there was a time that neither `message` and `author` was in the `Message` instance generated. They are now both there so we might want to just return the `EventsSubscribe` instance for simplicity. Other interface methods in this util still on old pattern mainly because we didn't get around to the low-pri refactoring.

src/types/events-types.ts
```diff
@@ -23,10 +23,10 @@ export type EventsRecordsFilter = {
   dateCreated?: RangeCriterion;
 };
 
-export type EventsQueryFilter = EventsMessageFilter | EventsRecordsFilter | ProtocolsQueryFilter;
+export type EventsFilter = EventsMessageFilter | EventsRecordsFilter | ProtocolsQueryFilter;
```
thehenrytsai, line 26: Maybe its a naming thing, but I got confused... like:
- What's the point of select few property for `EventsMessageFilter` if `EventFilter` accept `EventsRecordsFilter` and `ProtocolsQueryFilter`?

Can you add comments to both `EventsFilter` and `EventsMessageFilter` to show the differences?

tests/scenarios/subscriptions.spec.ts
```diff
@@ -0,0 +1,408 @@
...
+        const write2Response = await dwn.processMessage(alice.did, write2proto1.message, { dataStream: write2proto1.dataStream });
+        expect(write2Response.status.code).equals(202);
+
+        // create another record for proto2
+        const write2proto2 = await TestDataGenerator.generateRecordsWrite({ author: alice, protocol: proto2, ...postProperties });
+        const write2Proto2Response = await dwn.processMessage(alice.did, write2proto2.message, { dataStream: write2proto2.dataStream });
+        expect(write2Proto2Response.status.code).equals(202);
+
+        // proto1 messages from handler do not change.
+        expect(proto1Messages.length).to.equal(1, 'proto1 after subscription.off()');
```
thehenrytsai, line 269: Nit: no more `.off()` method, 2 occurences.

tests/interfaces/events-subscribe.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { EventsSubscribe } from '../../src/interfaces/events-subscribe.js';
+import { DidKeyResolver, DwnInterfaceName, DwnMethodName, Jws, Time } from '../../src/index.js';
+
+import { expect } from 'chai';
+
+describe('EventsSubscribe', () => {
+  describe('create()', () => {
+    it('should be able to create and authorize EventsSubscribe', async () => {
```
thehenrytsai, line 8: Missing "and authorize EventsSubscribe" part.

src/event-log/event-emitter-stream.ts
```diff
@@ -0,0 +1,53 @@
...
+
+type EventStreamEmitterConfig = {
+  emitter?: EventEmitter;
+};
+
+export class EventEmitterStream implements EventStream {
+  private eventEmitter: EventEmitter;
+  private isOpen: boolean = false;
+
+  constructor(config?: EventStreamEmitterConfig) {
```
thehenrytsai, line 17: For my understanding: what's the reason for needing to pass and `EventEmitter` as opposed to just create it for the caller?

tests/event-log/event-stream.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    eventStream = TestEventStream.get();
+    await eventStream.open();
+  });
+
+  after(async () => {
+    // Clean up after each test by closing and clearing the event stream
+    await eventStream.close();
+  });
+
+  it('emits all messages to each subscriptions', async () => {
```
thehenrytsai, line 25: Sanity: there is only one subscription though? Nit on typo "subcription"

tests/event-log/event-stream.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const message2 = await TestDataGenerator.generateRecordsWrite({});
+    eventStream.emit('did:alice', message2.message, {});
+
+    await Time.minimalSleep();
+
+    expect(messageCids).to.have.members([ message1Cid ]);
+  });
+
+  it('does not emit messages if emitter is closed', async () => {
```
thehenrytsai, line 81: Nit: rename, emitter is an implementation detail.

```suggestion
  it('does not emit messages if event stream is closed', async () => {
```

src/event-log/event-emitter-stream.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async close(): Promise<void> {
+    this.isOpen = false;
+    this.eventEmitter.removeAllListeners();
+  }
+
+  emit(tenant: string, message: GenericMessage, indexes: KeyValues): void {
+    if (!this.isOpen) {
+      // silently ignore
```
thehenrytsai, line 44: Discussion: we should at least emit an error once?

src/types/events-types.ts
```diff
@@ -23,10 +24,12 @@ export type EventsRecordsFilter = {
   dateCreated?: RangeCriterion;
 };
 
-export type EventsQueryFilter = EventsMessageFilter | EventsRecordsFilter | ProtocolsQueryFilter;
+// a union type of the different types of filters a user can use when issuing an EventsQuery or EventsSubscribe
```
thehenrytsai, line 27: Use `/**` syntax so SDK doc-gen tool and IDE intelli-sense can pick it up. Same goes to two type comments above.

src/handlers/records-delete.ts
thehenrytsai: Sanity follow up: the `published` index is no longer need for scenario you were after?

tests/handlers/permissions-request.spec.ts
```diff
@@ -57,6 +57,25 @@ export function testPermissionsRequestHandler(): void {
         await dwn.close();
       });
 
+      it('should process a PermissionsRequest with EventStream not set', async () => {
```
thehenrytsai, line 60: For my understanding: what is this test attempting to cover? I didn't think `EventStream` non-existence could inadvertently cause issue to handling of `PermissionsRequest`. What am I missing?

tests/event-log/event-emitter-stream.spec.ts
```diff
@@ -37,8 +36,8 @@ describe('EventEmitterStream', () => {
   });
 
   it('should remove listeners when `close` method is used', async () => {
-    const emitter = new EventEmitter();
-    eventStream = new EventEmitterStream({ emitter });
+    eventStream = new EventEmitterStream();
+    const emitter = (eventStream as any).eventEmitter as EventEmitter;
```
thehenrytsai, line 40: Arguably nicer syntax, 2 occurrences:

```
    const emitter = eventStream['eventEmitter'];
```

tests/handlers/permissions-request.spec.ts
thehenrytsai: Ah, looks like a code coverage thing, just noticed you added it for all Interface Methods. What line is CC complaining about...?!

thehenrytsai:
🐐 🔥 🥇 🚀

PR #660, opened 2024-01-12
create a message options for data stream and future data structures

LiranCohen:


src/dwn.ts
```diff
@@ -106,14 +105,16 @@ export class Dwn {
   public async processMessage(tenant: string, rawMessage: RecordsDeleteMessage): Promise<GenericMessageReply>;
   public async processMessage(tenant: string, rawMessage: RecordsQueryMessage): Promise<RecordsQueryReply>;
   public async processMessage(tenant: string, rawMessage: RecordsReadMessage): Promise<RecordsReadReply>;
-  public async processMessage(tenant: string, rawMessage: RecordsWriteMessage, dataStream?: Readable): Promise<GenericMessageReply>;
-  public async processMessage(tenant: string, rawMessage: unknown, dataStream?: Readable): Promise<UnionMessageReply>;
-  public async processMessage(tenant: string, rawMessage: GenericMessage, dataStream?: Readable): Promise<UnionMessageReply> {
+  public async processMessage(tenant: string, rawMessage: RecordsWriteMessage, options?: MessageOptions): Promise<GenericMessageReply>;
+  public async processMessage(tenant: string, rawMessage: unknown, options?: MessageOptions): Promise<UnionMessageReply>;
+  public async processMessage(tenant: string, rawMessage: GenericMessage, options: MessageOptions = {}): Promise<UnionMessageReply> {
```
thehenrytsai, line 110: For my understanding:

Too annoying to do: `options?: MessageOptions`?

src/dwn.ts
```diff
@@ -106,14 +105,16 @@ export class Dwn {
   public async processMessage(tenant: string, rawMessage: RecordsDeleteMessage): Promise<GenericMessageReply>;
   public async processMessage(tenant: string, rawMessage: RecordsQueryMessage): Promise<RecordsQueryReply>;
   public async processMessage(tenant: string, rawMessage: RecordsReadMessage): Promise<RecordsReadReply>;
-  public async processMessage(tenant: string, rawMessage: RecordsWriteMessage, dataStream?: Readable): Promise<GenericMessageReply>;
-  public async processMessage(tenant: string, rawMessage: unknown, dataStream?: Readable): Promise<UnionMessageReply>;
-  public async processMessage(tenant: string, rawMessage: GenericMessage, dataStream?: Readable): Promise<UnionMessageReply> {
+  public async processMessage(tenant: string, rawMessage: RecordsWriteMessage, options?: MessageOptions): Promise<GenericMessageReply>;
```
thehenrytsai, line 108: Can we introduce `RecordsWriteMessageOptions` now with the assumption that there will be another param(s) added to the `MessageOption` in the next PR?

src/dwn.ts
LiranCohen: Yeah was easier to be able to just pull out the optional fields within it:
```
const { dataStream } = options;
```
And since all of the fields are optional when taking them out of options and using them in the generic handle function, it felt cleaner.

src/dwn.ts
LiranCohen: Yup will do this now! I was going to do it but i figured usually we wait until there's the need for it.

thehenrytsai (approved):

PR #665, opened 2024-01-19
Enhance EventEmitterStream with options

LiranCohen:
Set `maxListeners` option on `EventEmitter` to 0, which represents infinity.
This is a soft limit of when to issue warnings into the stderr of the system but does not have an effect on actual event listeners.
https://nodejs.org/api/events.html#eventsdefaultmaxlisteners

Exposed a way to capture errors, with the default being sent to `console.error` as before.

### DIFF
--- a/src/event-log/event-emitter-stream.ts
+++ b/src/event-log/event-emitter-stream.ts
@@ -9,16 +9,6 @@ const EVENTS_LISTENER_CHANNEL = 'events';
 
 export interface EventEmitterStreamConfig {
   /**
-   * The maximum number of listeners per eventName before a warning is emitted.
-   * 0 would represent infinity, which is the default if no value is set.
-   *
-   * this is not a hard limit, only a limit for warnings to be emitted so that a memory leak could be found.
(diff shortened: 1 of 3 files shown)


src/event-log/event-emitter-stream.ts
```diff
@@ -3,24 +3,42 @@ import type { KeyValues } from '../types/query-types.js';
...
+   */
+  errorHandler?: (error: any) => void;
+};
+
 export class EventEmitterStream implements EventStream {
   private eventEmitter: EventEmitter;
   private isOpen: boolean = false;
 
-  constructor() {
+  constructor(config?: EventEmitterStreamConfig) {
```
thehenrytsai, line 22: Optional: I kinda what you did as a pattern that I saw somewhere.

```suggestion
  constructor(config: EventEmitterStreamConfig = {}) {
```

thehenrytsai (dismissed):

thehenrytsai (approved):
🐐

PR #667, opened 2024-01-19
RecordsSubscribe

LiranCohen:
This PR builds on top of the existing `EventStream` and adds a subscription specifically for the `Records` interface which follows similar rules as those for `RecordsQuery`.

Currently the subscription does not evict or re-authorize subscriptions after the initial processing of the message. So if a revocation of permissions happens after the subscription is live, the user will continue getting notified until they close the connection.

This will be addressed in a separate PR to keep things more concise and easier to follow. https://github.com/TBD54566975/dwn-sdk-js/issues/668


Q_AND_A.md
```diff
@@ -79,3 +79,6 @@
   - `write` - allows a DID to create and update the record they have created
   - `update` - allows a DID to update a record, regardless of the initial author
 
+  ## Subscriptions
+ - Same considerations as for `RecordsQuery` wrt invoking a protocol role for a specific `protocolPath`.
+ - Currently if a subscription is no longer authorized but it is still active, the subscriber will still receive updates until they close the subscription. This will be addressed ina separate PR https://github.com/TBD54566975/dwn-sdk-js/issues/668 - last updated (2024/01/22)
```
thehenrytsai, line 84: A few thoughts:
1. Entries ought to follow the same "question followed by an answer" here, the intent of this document was mainly to capture "why questions" that people/we ask.
1. If this is more about documenting design, [technical-proposals](https://github.com/TBD54566975/technical-proposals) seems to be a much better place. I also understand if a total separate technical proposal seems like an overkill at times, in which case, consider a separate light weight `design.md` that captures all feature designs that fall in this category. I'd be down if you'd like to combine both Q&A and design into one document, in which case, let's rename to `design.md` and have two sections in the document: "Design" and "Q&A" sections.

json-schemas/interface-methods/protocol-rule-set.json
```diff
@@ -66,6 +66,7 @@
                 "enum": [
                   "delete",
                   "query",
+                  "subscribe",
```
thehenrytsai, line 69: Needing clarification on the requirements of this feature base on the statement "follows similar rules as those for `RecordsQuery`":

Without a protocol definition:
1. Can anonymous (unauthenticated) people subscribe to record streams?
1. Can authenticated recipient or author subscribe to record streams?

src/handlers/records-subscribe.ts
```diff
@@ -0,0 +1,225 @@
...
+    tenant: string,
+    recordsSubscribe: RecordsSubscribe,
+    messageStore: MessageStore
+  ): Promise<void> {
+
+    if (Message.isSignedByDelegate(recordsSubscribe.message)) {
+      await recordsSubscribe.authorizeDelegate(messageStore);
+    }
+
+    // Only run protocol authz if message deliberately invokes it
```
thehenrytsai, line 220: I am getting quite confused by this comment, going down the rabbit hole now.

Q_AND_A.md
LiranCohen: Makes sense! Maybe I will re-word it to question and answer for this case?

The intent was to document this behavior somewhere until we implement it as I think it's something that people would wonder about when using this. I'm pretty sure it will not be long before it's implemented, or it be troublesome, but wanted to make sure we get certain functionality in place before going down this route.

Maybe some wording like:
```suggestion
  ## Subscriptions
 - What happens to a subscription which is listening to events, but is no longer authorized due to revocation of a grant or role? 
 
Currently if a subscription is no longer authorized but it is still active, the subscriber will still receive updates until they close the subscription.
This will be addressed in a future upgrade and we've created an issue to track it. https://github.com/TBD54566975/dwn-sdk-js/issues/668 - last updated (2024/01/22)
```

json-schemas/interface-methods/protocol-rule-set.json
LiranCohen: Yeah I think I need to describe this a bit better somewhere. I will add a section to the readme that describes subscriptions in general and these rules.

To answer your questions:
1. Yes, they will receive messages that are marked `published:true`.

Not sure if I fully understand the question in 2. If you'd like to chat about that we can.

src/handlers/records-subscribe.ts
LiranCohen: This is copy-pasta from `handlers/records-query.ts`.

There is definitely some room for refactoring some duplicate code between read/query/subscribe in this regard. Let me know if you'd like to chat about it.

thehenrytsai: Yeah, I knew it came from query, but still was confused why the comment stated what it stated.

File this bug after digging, something that subscribe should probably take care of if not already:

https://github.com/TBD54566975/dwn-sdk-js/issues/669

Q_AND_A.md
thehenrytsai: Got it, given the intent and the way you have reworded it, the Q&A is a great place to include it.

json-schemas/interface-methods/protocol-rule-set.json
thehenrytsai: For #2, I was wondering about if the subscription is signed, but does not reference a permission or role, or protocol. Basically was imagining in my head what the perf would be like. I guess I will learn more as I read on.

src/handlers/records-subscribe.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Filters for only published records.
+   */
+  private static async filterPublishedRecords(
+    recordsSubscribe: RecordsSubscribe
+  ): Promise<Filter[]> {
+    const filter = RecordsSubscribeHandler.buildPublishedRecordsFilter(recordsSubscribe);
+    return [filter];
+  }
```
thehenrytsai, line 152: Seems redundant.

src/core/protocol-authorization.ts
```diff
@@ -152,6 +153,47 @@ export class ProtocolAuthorization {
     );
   }
 
+  public static async authorizeSubscription(
```
thehenrytsai, line 156: `authorizeSubscribe`

thehenrytsai, line 156: Design discussion (I am sure you anticipated some form of discussion): this method (as well as the `authorizeSubscribe()` in `RecordsGrantAuthorization`) appear to be entirely the same, minus the type info. Given this is not a coincidence, ie. you intended for the auth to be identical, it seems like it's worth it to not duplicate code to realize that intent. This probably involves generalizing the type differences, with the hope that this doesn't involve crazy gymnastics...

src/types/records-types.ts
```diff
@@ -146,6 +153,20 @@ export type RecordsQueryReply = GenericMessageReply & {
   cursor?: PaginationCursor;
 };
 
+export type RecordsHandler = (message: RecordsWriteMessage | RecordsDeleteMessage) => void;
```
thehenrytsai, line 156:
```suggestion
export type RecordSubscriptionHandler = (message: RecordsWriteMessage | RecordsDeleteMessage) => void;
```

Q_AND_A.md
LiranCohen: Fixed! Added `RecordsSubscribe` to the protocol answer that relates to the `protocolPath`. Let me know if that reads better and makes sense.

src/handlers/records-subscribe.ts
LiranCohen: Good catch! Will discuss the implications within that specific issue.

src/handlers/records-subscribe.ts
LiranCohen: Is redundant! removed!

src/core/protocol-authorization.ts
LiranCohen: Updated name

LiranCohen: Let's chat about this so that I make sure I'm following what you mean. I think I do.

src/types/records-types.ts
LiranCohen: Done

json-schemas/interface-methods/protocol-rule-set.json
LiranCohen: If I understand correctly this is the flow you mean:
https://github.com/TBD54566975/dwn-sdk-js/blob/dff724aacd457f695de6f70655b787f7ab6fc073/tests/scenarios/subscriptions.spec.ts#L972

Which would filter based on the author of the subscription being the author or recipient of a record. However this runs into the edge case issue which you found in https://github.com/TBD54566975/dwn-sdk-js/issues/669

src/handlers/records-query.ts
```diff
@@ -144,21 +144,21 @@ export class RecordsQueryHandler implements MethodHandler {
...
+    if (Records.filterIncludesUnpublishedRecords(filter)) {
       filters.push(RecordsQueryHandler.buildUnpublishedRecordsByQueryAuthorFilter(recordsQuery));
 
       const recipientFilter = recordsQuery.message.descriptor.filter.recipient;
       if (recipientFilter === undefined || recipientFilter === recordsQuery.author) {
         filters.push(RecordsQueryHandler.buildUnpublishedRecordsForQueryAuthorFilter(recordsQuery));
       }
 
-      if (RecordsQueryHandler.shouldProtocolAuthorizeQuery(recordsQuery)) {
+      if (Records.hasProtocolRole(recordsQuery.signaturePayload!)) {
```
thehenrytsai, line 161: Note to self: let's discuss this, it seems the original name describes the intent better, but want to hear what you are thinking.

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+      });
+
+      after(async () => {
+        await dwn.close();
+      });
+
+      it('should return a subscription object', async () => {
+        const alice = await DidKeyResolver.generate();
+
+        // subscribe for non-normalized protocol
```
thehenrytsai, line 123: What does "non-normalized protocol" mean?

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+            protocol     : protocolDefinition.protocol,
+            protocolPath : 'friend',
+            data         : new TextEncoder().encode('Bob is my friend'),
+          });
+          const friendRoleReply = await dwn.processMessage(alice.did, friendRoleRecord.message, { dataStream: friendRoleRecord.dataStream });
+          expect(friendRoleReply.status.code).to.equal(202);
+
+          const recordIds: string[] = [];
+          const addRecord = async (message: GenericMessage): Promise<void> => {
+            if (message.descriptor.interface === DwnInterfaceName.Records && message.descriptor.method === DwnMethodName.Write) {
```
thehenrytsai, line 387: Consider:

```suggestion
          const addRecord: RecordSubscriptionHandler = async (message): Promise<void> => {
            if (message.descriptor.method === DwnMethodName.Write) {
```

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+
+          // there should not be any messages in the subscription without a friend role.
+          expect(noRoleRecords.length).to.equal(0);
+
+          // should have all chat messages
+          expect(recordIds).to.have.members(chatRecordIds);
+        });
+
+        it('allows protocol authorized subscriptions', async () => {
+          // scenario: Alice writes some chat messages writes a chat message.
```
thehenrytsai, line 429: Typo in the original `RecordsQuery` test.

```suggestion
          // scenario: Alice writes some chat messages.
```

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+          expect(threadRoleReply.status.code).to.equal(202);
+
+          const filter: RecordsFilter = {
+            protocol     : protocolDefinition.protocol,
+            protocolPath : 'thread/chat',
+            contextId    : threadRecord.message.contextId,
+          };
+
+          const noRoleRecords: string[] = [];
+          const addNoRole = async (message: GenericMessage): Promise<void> => {
```
thehenrytsai, line 461: Good call on having this check. But this does bring up a point: doesn't this mean we allow anyone to subscribe to "nothing" which essentially is a very simple mechanism for DDOS? I do wonder if it is worth while thinking about prevention/mitigation further, even if outside the scope of this PR.

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+              protocolPath : 'thread/chat',
+              published    : false,
+              contextId    : threadRecord.message.contextId,
+              parentId     : threadRecord.message.recordId,
+              data         : new TextEncoder().encode('Bob can read this cuz he is my friend'),
+            });
+            const chatReply = await dwn.processMessage(alice.did, chatRecord.message, { dataStream: chatRecord.dataStream });
+            expect(chatReply.status.code).to.equal(202);
+            chatRecordIds.push(chatRecord.message.recordId);
+          }
```
thehenrytsai, line 649: Consider remove.

tests/handlers/records-subscribe.spec.ts
```diff
@@ -0,0 +1,773 @@
...
+              recipient    : alice.did,
+              protocol     : protocolDefinition.protocol,
+              protocolPath : 'chat',
+              published    : false,
+              data         : new TextEncoder().encode('Bob can read this cuz he is my friend'),
+            });
+            const chatReply = await dwn.processMessage(alice.did, chatRecord.message, { dataStream: chatRecord.dataStream });
+            expect(chatReply.status.code).to.equal(202);
+            chatRecordIds.push(chatRecord.message.recordId);
+          }
```
thehenrytsai, line 578: Consider remove.

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -227,6 +226,63 @@ export function testDelegatedGrantScenarios(): void {
       const participantRoleReply = await dwn.processMessage(bob.did, participantRoleRecord.message, { dataStream: participantRoleRecord.dataStream });
       expect(participantRoleReply.status.code).to.equal(202);
 
+      // Alice creates a delegated subscribe grant for device X to act as Alice.
```
thehenrytsai, line 229: Need to minimally update the title and scenario summary, but due to the nature of difference in sequencing, I feel like this is probably better off being it's own scenario.

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -641,9 +731,9 @@ export function testDelegatedGrantScenarios(): void {
         }
       });
 
-      const deviceXWriteReply = await dwn.processMessage(bob.did, recordsReadByDeviceX.message);
-      expect(deviceXWriteReply.status.code).to.equal(401);
-      expect(deviceXWriteReply.status.detail).to.contain(DwnErrorCode.RecordsGrantAuthorizationScopeProtocolMismatch);
+      const deviceXReadReply = await dwn.processMessage(bob.did, recordsReadByDeviceX.message);
+      expect(deviceXReadReply.status.code).to.equal(401);
+      expect(deviceXReadReply.status.detail).to.contain(DwnErrorCode.RecordsGrantAuthorizationScopeProtocolMismatch);
```
thehenrytsai, line 736: 🙏

tests/scenarios/subscriptions.spec.ts
```diff
@@ -924,5 +924,289 @@ export function testSubscriptionScenarios(): void {
         expect(messageCids).to.eql([ record1MessageCid ]);
       });
     });
+
+    describe('records subscribe', () => {
+      it('allows for anonymous subscriptions to published records', async () => {
+        const alice = await DidKeyResolver.generate();
+
+        // subscribe to this thread's events
```
thehenrytsai, line 932: C&P error? If so remove.

tests/scenarios/subscriptions.spec.ts
```diff
@@ -924,5 +924,289 @@ export function testSubscriptionScenarios(): void {
...
+
+        const chatSubscription = await TestDataGenerator.generateRecordsSubscribe({
+          author : alice,
+          filter : { protocol: protocol, protocolPath: 'thread/chat', parentId: thread.message.recordId } // chat updates
+        });
+        const chatSubscriptionReply = await dwn.processMessage(alice.did, chatSubscription.message, {
+          subscriptionHandler
+        });
+        expect(chatSubscriptionReply.status.code).to.equal(200);
+        expect(chatSubscriptionReply.subscription).to.exist;
```
thehenrytsai, line 1105: For future, but this definitely feels super forced, this probably can go away if we support one-of filter for protocol path (or for all strings while at it), we also would probably one to support the same for permission grants.

LiranCohen: Yup was thinking the same, best for a separate effort but it shouldn't be anything crazy.

tests/scenarios/delegated-grant.spec.ts
LiranCohen: Yeah I was thinking about that as well but didn't want to have too much duplication.

But if you also feel that a separate scenario wouldn't be bad I'll break it out into it's own.

tests/handlers/records-subscribe.spec.ts
LiranCohen: Fixed!

tests/handlers/records-subscribe.spec.ts
LiranCohen: Fixed here and another identical place in the tests.

tests/handlers/records-subscribe.spec.ts
LiranCohen: Fixed in both!

tests/handlers/records-subscribe.spec.ts
LiranCohen: Makes sense, done!

tests/handlers/records-subscribe.spec.ts
LiranCohen: Done!

tests/scenarios/delegated-grant.spec.ts
LiranCohen: Done, as well as in the `should fail if delegated grant has a mismatching protocol scope - subscribe`

tests/scenarios/subscriptions.spec.ts
```diff
@@ -924,5 +924,289 @@ export function testSubscriptionScenarios(): void {
...
+
+        const write1 = await TestDataGenerator.generateRecordsWrite({ author: alice, schema: 'http://schema1', recipient: bob.did });
+        const write1Reply = await dwn.processMessage(alice.did, write1.message, { dataStream: write1.dataStream });
+        expect(write1Reply.status.code).to.equal(202);
+
+        const write2 = await TestDataGenerator.generateRecordsWrite({ author: alice, schema: 'http://schema1', recipient: bob.did });
+        const write2Reply = await dwn.processMessage(alice.did, write2.message, { dataStream: write2.dataStream });
+        expect(write2Reply.status.code).to.equal(202);
+
+        // will not be emitted as it is not intended for bob
```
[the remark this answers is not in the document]
LiranCohen: Done

src/core/protocol-authorization.ts
LiranCohen: Done!

There's room for further removal of duplication as we talked about, but likely would be better in a separate effort.

src/handlers/records-query.ts
LiranCohen: renamed to `shouldProtocolAuthorize`. 

My original thinking was that the method doesn't do much, so maybe it should describe what it is actually doing, but after further consideration I renamed it and the comment really helps the user see what's going on without clicking into the function.

LiranCohen:
@thehenrytsai I've updated the PR based on comments and discussions.

src/core/protocol-authorization.ts
```diff
@@ -152,22 +153,18 @@ export class ProtocolAuthorization {
...
-   */
-  public static async authorizeQuery(
+  public static async authorizeQueryOrSubscribe(
     tenant: string,
-    incomingMessage: RecordsQuery,
+    incomingMessage: RecordsQuery | RecordsSubscribe,
     messageStore: MessageStore,
   ): Promise<void> {
-    // validate that required properties exist in query filter
+    // validate that required properties exist in the record filter
```
thehenrytsai, line 161: Nit: remove, probably a left over from multiple refactoring.

src/handlers/records-subscribe.ts
thehenrytsai: I have updated #669 to include subscribe when fixing.

tests/handlers/records-subscribe.spec.ts
thehenrytsai: There is one more block of this (starting line 563).

tests/utils/test-data-generator.ts
```diff
@@ -166,6 +168,22 @@ export type GenerateRecordsQueryOutput = {
   message: RecordsQueryMessage;
 };
 
+export type GenerateRecordsSubscribeInput = {
+    /**
+   * Treated as `false` if not given.
+   */
```
thehenrytsai, line 174: Nit: indentation, dumb that linter doesn't pick this up.

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -641,9 +759,83 @@ export function testDelegatedGrantScenarios(): void {
...
+        author       : bob,
+        recipient    : alice.did,
+        protocol     : protocolDefinition.protocol,
+        protocolPath : 'thread/participant',
+        contextId    : threadRecord.message.contextId,
+        parentId     : threadRecord.message.recordId,
+        data         : new TextEncoder().encode('Alice is my friend'),
+      });
+      const participantRoleReply = await dwn.processMessage(bob.did, participantRoleRecord.message, { dataStream: participantRoleRecord.dataStream });
+      expect(participantRoleReply.status.code).to.equal(202);
```
thehenrytsai, line 808: Consider:

1. Just include the subscribe test in the read & query like what you did. or
1. Removing this block and move the test into functional test file.

This block is basically irrelevant (thus can be removed by just setting the `contextId` to `irrelevant`), because authorization takes place first.

Giving more context: Read and delete needed the initial setup because the code path is different: they fetch the record of that real `recordId` first before authorization.

Query on the other hand require less setup, so I figured that I'd roll it into the read.

tests/handlers/records-subscribe.spec.ts
LiranCohen: Woops! Done!

src/core/protocol-authorization.ts
LiranCohen: Done

tests/utils/test-data-generator.ts
LiranCohen: Good catch with the comment indentations! Done

tests/scenarios/delegated-grant.spec.ts
LiranCohen: Yup, makes sense.

I went with #1. See if you're ok with the order of events/commenting I put in place.

tests/scenarios/delegated-grant.spec.ts
```diff
@@ -693,6 +693,36 @@ export function testDelegatedGrantScenarios(): void {
...
+        signer         : Jws.createSigner(deviceX),
+        delegatedGrant : subscribeGrantForDeviceX.asDelegatedGrant(),
+        protocolRole   : 'thread/participant',
+        filter         : {
+          protocol,
+          contextId    : threadRecord.message.contextId,
+          protocolPath : 'thread/chat'
+        }
+      });
+      const deviceXSubscribeQueryReply = await dwn.processMessage(bob.did, recordsSubscribeByDeviceX.message);
```
thehenrytsai, line 722: Nit: I think you meant to rename to `deviceXRecordsSubscribeReply` LOL.

LiranCohen: 😆  that's a good one.

LiranCohen: Done

thehenrytsai (approved):
💯

PR #677, opened 2024-02-01
Added a mock blockstore to compute CID of large data streams with no/tiny memory usage

thehenrytsai:
Another production-readiness fix. This is particularly useful when dealing with large files and a necessity in a large-scale production service environment.

Prior to this PR, we used the In-Memory Blockstore to compute the CID, which would temporarily store the entire data in memory until the the CID is computed. (e.g. 10 people uploading 1GB at the same time would require 10GB of memory)

Did a sanity test with a 5 GB data stream: original implementation would consume gigabytes of memory during the computation of the CID; the new implementation consumes zero/little additional memory (minus the process itself).


LiranCohen (approved):
🥇

PR #679, opened 2024-02-02
Bump `@web5/dids` to 0.3.0

frankhinek:
This PR will:
- Bump the `@web5/dids` dependency from 0.2.4 to 0.3.0.
- Removed the `@types/dns-packet` dev dependency since it is no longer needed.
- Removes `src/did/did.ts` and the associated tests since there are no longer any references to these methods in DWN SDK.  Equivalent functionality is available in [`DidUri`](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-uri.ts) should it ever be needed.
- Removes `TestDataGenerator.getKeyId()` since a DID method agnostic method `getSigningMethod()` is now available across all DID method implementations in `@web5/dids`.


LiranCohen (approved):
🥇  Slowly standardizing around the common libs. Love it!

PR #682, opened 2024-02-07
Optionally limit protocol message data size

ianpatton:
Adds a new protocol rule `$size` that if provided, may contain a `min` or `max` size in bytes.

If max < min, throws error on protocol configure


tests/interfaces/protocols-configure.spec.ts
```diff
@@ -370,6 +370,30 @@ describe('ProtocolsConfigure', () => {
         await expect(createProtocolsConfigurePromise)
           .to.be.rejected;
       });
+
+      it('allows $sizeLimit to be set on records', async () => {
```
LiranCohen, line 374: Suggestion:
```suggestion
      it('allows $sizeLimit to be set on a protocol path', async () => {
```

json-schemas/interface-methods/protocol-rule-set.json
```diff
@@ -84,6 +84,10 @@
     "$contextRole": {
       "$comment": "When `true`, this turns a record into `role` that may be used within a context",
       "type": "boolean"
+    },
+    "$sizeLimit": {
+      "$comment": "When set, this limits the size of the record in bytes",
+      "type": "number"
```
LiranCohen, line 90: Suggestion: implement a minimum of 0 or 1 so that schema validation blows up if they put a negative number.

@thehenrytsai @csuwildcat what are your thoughts of allowing a zero to represent no data any usefulness to that? Otherwise we can do a minimum of 1 byte?

```suggestion
      "type": "number",
      "minimum": 0
```

tests/scenarios/protocol-message-size-limits.spec.ts
```diff
@@ -0,0 +1,109 @@
+import type { EventStream } from '../../src/types/subscriptions.js';
```
LiranCohen, line 1: Great tests! 🥇 

We probably don't need a scenario section fort this test as it isn't huge.
Might fit well around here since it relates to protocol write rules specifically:

https://github.com/TBD54566975/dwn-sdk-js/blob/b4e09343946d231df19a86d9aed79619fec5992b/tests/handlers/records-write.spec.ts#L1166

tests/scenarios/protocol-message-size-limits.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+        plaintextBytes                                   : blobBytes2,
+        author                                           : alice,
+        protocolDefinition                               : protocolDefinition,
+        protocolPath                                     : 'blob',
+        encryptSymmetricKeyWithProtocolPathDerivedKey    : false,
+        encryptSymmetricKeyWithProtocolContextDerivedKey : true
+      });
+      const blobRecordReply2 = await dwn.processMessage(alice.did, blobRecord2.message, { dataStream: blobRecord2.dataStream });
+      expect(blobRecordReply2.status.code).to.equal(400);
+
```
LiranCohen, line 106: Sanity: check for the specific error code.
```suggestion
      expect(blobRecordReply2.status.code).to.equal(400);
      expect(blobRecordReply2.status.detail).to.contain(DwnErrorCode.ProtocolAuthorizationSizeLimitExceeded);
```

LiranCohen (changes requested):
👑 🐐  This is great!! Just a few small nits.

src/core/protocol-authorization.ts
```diff
@@ -579,6 +582,31 @@ export class ProtocolAuthorization {
...
+      throw new DwnError(DwnErrorCode.ProtocolAuthorizationSizeInvalid, `message size ${messageSize} is less than allowed ${min}`);
+    }
+
+    if (max === undefined) {
+      return;
+    }
+
+    if (messageSize > max) {
+      throw new DwnError(DwnErrorCode.ProtocolAuthorizationSizeInvalid, `message size ${messageSize} is more than allowed ${max}`);
+    }
```
LiranCohen, line 607: This is great! One convention we try to keep in this repo is to only use an error code once, so just splitting these two into separate codes would be great.

Something like this:
```suggestion
    if (messageSize < min) {
      throw new DwnError(DwnErrorCode.ProtocolAuthorizationMinSizeInvalid, `message size ${messageSize} is less than allowed ${min}`);
    }

    if (max === undefined) {
      return;
    }

    if (messageSize > max) {
      throw new DwnError(DwnErrorCode.ProtocolAuthorizationMaxSizeInvalid, `message size ${messageSize} is more than allowed ${max}`);
    }
```

src/types/protocols-types.ts
```diff
@@ -125,6 +125,11 @@ export type ProtocolRuleSet = {
    * by setting `protocolRole` property to the protocol path of the $contextRole record.
    */
   $contextRole?: boolean;
+  // If $size is set, the record size in bytes must be within the limits.
```
LiranCohen, line 128: Suggestion: modify comment style in this case so that `jsdoc` will pick up the comment in the IDE.
```suggestion
  /**
   * If $size is set, the record size in bytes must be within the limits.
   */
```

tests/handlers/records-write.spec.ts
```diff
@@ -3342,6 +3342,90 @@ export function testRecordsWriteHandler(): void {
...
+          const protocol = protocolDefinition.protocol;
+          const protocolConfig = await TestDataGenerator.generateProtocolsConfigure({
+            author: alice,
+            protocolDefinition,
+          });
+
+          const protocolConfigureReply = await dwn.processMessage(alice.did, protocolConfig.message);
+          expect(protocolConfigureReply.status.code).to.equal(202);
+
+          const data = TestDataGenerator.randomBytes(999);
```
LiranCohen, line 3374: Suggestion: Add a positive test somewhere here as well to show that anything over 1kb will make it in.

tests/handlers/records-write.spec.ts
```diff
@@ -3342,6 +3342,90 @@ export function testRecordsWriteHandler(): void {
...
+          const protocol = protocolDefinition.protocol;
+          const protocolConfig = await TestDataGenerator.generateProtocolsConfigure({
+            author: alice,
+            protocolDefinition,
+          });
+
+          const protocolConfigureReply = await dwn.processMessage(alice.did, protocolConfig.message);
+          expect(protocolConfigureReply.status.code).to.equal(202);
+
+          const data = TestDataGenerator.randomBytes(1001);
```
LiranCohen, line 3416: Suggestion: Same as other test, add a positive test somewhere here as well to show that anything under 1kb will make it in.

tests/interfaces/protocols-configure.spec.ts
```diff
@@ -370,6 +370,87 @@ describe('ProtocolsConfigure', () => {
...
+
+        const alice = await TestDataGenerator.generatePersona();
+
+        const createProtocolsConfigurePromise = ProtocolsConfigure.create({
+          signer: Jws.createSigner(alice),
+          definition
+        });
+
+        await expect(createProtocolsConfigurePromise)
+          .to.be.rejected;
```
LiranCohen, line 452: Nit: test for the specific error code.

```suggestion
        await expect(createProtocolsConfigurePromise).to.eventually.be.rejectedWith(DwnErrorCode.ProtocolsConfigureInvalidSize);
```

LiranCohen (changes requested):
Just a few small nits, but looks good!! 🎸 🚀

src/interfaces/protocols-configure.ts
```diff
@@ -113,6 +113,17 @@ export class ProtocolsConfigure extends AbstractMessage<ProtocolsConfigureMessag
       );
     }
 
+    if (ruleSet.$size !== undefined) {
+      const { min = 0, max } = ruleSet.$size;
+
+      if (max !== undefined && max < min) {
+        throw new DwnError(
+          DwnErrorCode.ProtocolsConfigureInvalidSize,
+          `Invalid size range found at protocol path '${protocolPath}'`
```
thehenrytsai, line 122: Nit:
```suggestion
          `Invalid size range found: max limit ${max} less than min limit ${min} at protocol path '${protocolPath}'`
```

src/core/protocol-authorization.ts
```diff
@@ -579,6 +582,31 @@ export class ProtocolAuthorization {
...
+   * Verifies that writes adhere to the $size constraints if provided
+   * @throws {Error} if size is exceeded.
+   */
+  private static verifySizeLimit(
+    incomingMessage: RecordsWrite,
+    inboundMessageRuleSet: ProtocolRuleSet
+  ): void {
+    const { min = 0, max } = inboundMessageRuleSet.$size || {};
+
+    const messageSize = incomingMessage.message.descriptor.dataSize;
```
thehenrytsai, line 595: Rename to `dataSize` to be precise.

src/core/protocol-authorization.ts
```diff
@@ -579,6 +582,31 @@ export class ProtocolAuthorization {
...
+
+    if (messageSize < min) {
+      throw new DwnError(DwnErrorCode.ProtocolAuthorizationSizeInvalid, `message size ${messageSize} is less than allowed ${min}`);
```
thehenrytsai, line 598: Change to "Data size" instead of message size. We conventionally refer "Message" as the JSON object.

tests/interfaces/protocols-configure.spec.ts
ianpatton: 👍

src/interfaces/protocols-configure.ts
ianpatton: 👍

src/core/protocol-authorization.ts
ianpatton: 👍

src/core/protocol-authorization.ts
ianpatton: 👍

tests/handlers/records-write.spec.ts
ianpatton: 👍

src/core/protocol-authorization.ts
ianpatton: 👍

tests/handlers/records-write.spec.ts
```diff
@@ -3308,6 +3308,146 @@ export function testRecordsWriteHandler(): void {
...
+          const testRecord2 = await TestDataGenerator.generateRecordsWrite({
+            author       : alice,
+            recipient    : alice.did,
+            protocol,
+            protocolPath : 'blob',
+            data         : data2
+          });
+
+          const reply2 = await dwn.processMessage(alice.did, testRecord2.message, { dataStream: testRecord2.dataStream });
+          expect(reply2.status.code).to.equal(202);
```
LiranCohen, line 3364: Super small nit if you don't mind adding the negative test here, so a failure.
I know the failure is tested elsewhere but we like to include negative tests in the same place where positives are tested in most places.

Same comments for the other 2 below.

tests/handlers/records-write.spec.ts
LiranCohen: Super small nit, although we have a positive test in a separate test with both ranges set, would be nice to get a positive test here also.

tests/handlers/records-write.spec.ts
LiranCohen: Super small nit, although we have a positive test in a separate test with both ranges set, would be nice to get a positive test here also.

LiranCohen (changes requested):
Almost there! Just some really tiny nits within the tests.

LiranCohen (approved):
🚀🚀🚀

PR #685, opened 2024-02-09
Bump `@web5/dids` and add DID Resolver cache

frankhinek:
This PR will:
- Bump the `@web5/dids` package from `0.3.0` to `0.4.0`.
- Adds a DID resolver cache to the default instantiate of a `Dwn`.  The cache is a TTL cache with a 15 minute default timeout.  It uses a persistent LevelDB cache which will keep approximately 8,000 DID documents in an in-memory cache before flushing to disk.  Since the cache entries are persisted to disk, they will survive process restarts.
- Bump `@tbd54566975/dwn-sdk-js` to `0.2.17` so it can be published to NPM Registry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ try.js
 MESSAGESTORE
 DATASTORE
 EVENTLOG
+RESOLVERCACHE
 # location for levelDB data storage for non-browser tests
 TEST-DATASTORE
 TEST-MESSAGESTORE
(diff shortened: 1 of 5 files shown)


thehenrytsai (approved):
🐐🐐🐐

LiranCohen (dismissed):
🚀

csuwildcat (approved):

PR #687, opened 2024-02-13
Generalized roles so they can be specified under any context/sub-context

thehenrytsai:
The next PR will be replacing/removing the idea of "global role" and "context role" with just "role"


tests/scenarios/nested-roles.spec.ts
```diff
@@ -0,0 +1,281 @@
...
+        recipient       : daniel.did,
+        protocol        : protocolDefinition.protocol,
+        protocolPath    : 'community/gatedChannel/participant',
+        parentContextId : channel2Record.message.contextId,
+      });
+      const channel2ParticipantDanielRecordReply
+        = await dwn.processMessage(alice.did, channel2ParticipantDanielRecord.message, { dataStream: channel2ParticipantDanielRecord.dataStream });
+      expect(channel2ParticipantDanielRecordReply.status.code).to.equal(202);
+
+      // 6. Carol CANNOT add anyone as a participant in the gated-channel 2 since she is not a participant in the channel
```
LiranCohen, line 180: Not sure if it's relevant to add the participant test case as well?

```suggestion
      // 6. Carol send a message to the gated-channel 2 since she is not a participant in the channel
```

LiranCohen (commented):
🐐  Nice test scenario cases! Just a small comment mistake in the test case.

And I'm assuming the removal of `$globalRole` is coming soon after.

tests/scenarios/nested-roles.spec.ts
thehenrytsai: Hey @LiranCohen, the comment should be correct, it is testing to make sure Carol can't add any participant (note the expected 401 code).

Let me know if I am not making sense!

LiranCohen: Am I misreading it?  Below I see `carolChatMessage` which has `protocolPath` of `'community/gatedChannel/message'`.

Seems like she is trying to add a message to the chat using a participant role she doesn't have.

thehenrytsai: No no, you are right, the intend was correct, but my code was wrong! Fixing now!

LiranCohen: Ok SUPER small nit lol, `carolChatMessage` should probably be renamed.

LiranCohen (approved):
🎸

PR #696, opened 2024-02-26
Added tests to demonstrate direct child-role authorization in slack-like protocol

thehenrytsai:


tests/scenarios/nested-roles.spec.ts
```diff
@@ -54,23 +54,26 @@ export function testNestedRoleScenarios(): void {
       await dwn.close();
     });
 
-    it('should support Slack-like protocol with community and gated channels', async () => {
+    it.only('should support Slack-like protocol with community and gated channels', async () => {
```
LiranCohen, line 57:
```suggestion
    it('should support Slack-like protocol with community and gated channels', async () => {
```

LiranCohen (changes requested):
Looks good! Just forgot the `.only` in the test.

tests/scenarios/nested-roles.spec.ts
thehenrytsai: Thanks for the lightning fast review!

LiranCohen (approved):

Issue #690, opened 2024-02-16
Allow dataFormat to be modified


csuwildcat:
At least in Protocols, when multiple formats are specified, allow modification of the format declared for the file.

thehenrytsai:
This is now done.

PR #697, opened 2024-03-01
#690 - Allowed `dataFormat` to be modified

thehenrytsai:


tests/scenarios/subscriptions.spec.ts
```diff
@@ -824,11 +826,10 @@ export function testSubscriptionScenarios(): void {
 
         // wait for messages to emit and handler to process
         await Time.minimalSleep();
-        expect(imageMessages.length).to.equal(3);
+        expect(imageMessages.length).to.equal(2);
```
LiranCohen, line 829: Nit: I think it would be useful to either remove the delete of the image completely from this test, or add a comment somewhere here on why it isn't showing up in the subscription.

I know we have it in the Q&A but this might be confusing for anyone that's reading this test specifically.

LiranCohen (dismissed):
🚀

tests/scenarios/subscriptions.spec.ts
thehenrytsai: Good call! Removed.

LiranCohen (approved):
🐐

PR #702, opened 2024-03-07
Changed protocol 'can' property to an array

thehenrytsai:
1. Changed protocol 'can' property to an array
2. Disallowed protocol update action without create
3. Disallowed duplicated actors or roles within a protocol rule set
4. Various renames to improve code readability

### DIFF
--- a/src/interfaces/protocols-configure.ts
+++ b/src/interfaces/protocols-configure.ts
@@ -144,9 +144,11 @@ export class ProtocolsConfigure extends AbstractMessage<ProtocolsConfigureMessag
       }
     }
 
-    // Validate $actions in the rule set
-    const actions = ruleSet.$actions ?? [];
-    for (const actionRule of actions) {
+    // validate each action rule
+    const actionRules = ruleSet.$actions ?? [];
+    for (let i = 0; i < actionRules.length; i++) {
+      const actionRule = actionRules[i];
+
       // Validate the `role` property of an `action` if exists.
       if (actionRule.role !== undefined) {
         // make sure the role contains a valid protocol paths to a role record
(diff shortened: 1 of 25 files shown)


src/interfaces/protocols-configure.ts
```diff
@@ -144,52 +144,90 @@ export class ProtocolsConfigure extends AbstractMessage<ProtocolsConfigureMessag
...
-          action.of === undefined &&
-          ![ProtocolAction.CoUpdate, ProtocolAction.CoDelete].includes(action.can as ProtocolAction)
-      ) {
-        throw new DwnError(
-          DwnErrorCode.ProtocolsConfigureInvalidRecipientOfAction,
-          'Rules for `recipient` without `of` property must have `can` === `co-delete` or `co-update`'
-        );
+      if (actionRule.who === ProtocolActor.Recipient && actionRule.of === undefined) {
+
+        // throw if `can` contains a value that is not co-update` or `co-delete`
```
LiranCohen, line 180:
```suggestion
        // throw if `can` contains a value that is not `co-update` or `co-delete`
```

tests/handlers/protocols-configure.spec.ts
```diff
@@ -289,6 +292,147 @@ export function testProtocolsConfigureHandler(): void {
         expect(protocolsConfigureReply.status.detail).to.contain(DwnErrorCode.AuthorizationAuthorNotOwner);
       });
 
+      it('should reject ProtocolsConfigure with action rule containing duplicated actor (`who`) within a rule set', async () => {
```
LiranCohen, line 295: Suggestion:
```suggestion
      it('should reject ProtocolsConfigure with action rule containing duplicated actor (`who`) or (`of`) within a rule set', async () => {
```

Or potentially split it up into two tests.

LiranCohen (dismissed):
🥇  This is excellent! Really makes the protocol definitions easier to digest. 🐐

tests/handlers/protocols-configure.spec.ts
thehenrytsai: Done. Reworded for better clarity.

src/interfaces/protocols-configure.ts
thehenrytsai: Thanks! Done.

LiranCohen (approved):

PR #704, opened 2024-03-11
Removed `write` protocol action

thehenrytsai:
This is considered an unnecessary and potentially confusing "sugar" now that we have explicit `create` and `update` actions.


LiranCohen (approved):
🚀 🚀 🚀

PR #705, opened 2024-03-13
add tenant to EventsStream interface subscribe method

LiranCohen:
Adding `tenant` parameter to the subscribe method in order to better organize the pub/sub.


thehenrytsai (approved):
🚀

Issue #642, opened 2023-12-05
Add Tags to `RecordsWrite`


LiranCohen:
Allowing a user to `tag` a `record`, and then query for records that match the tags.

```typescript
export type Tags = {
  [tag: string]: Array<string | number | boolean>
};


export type RecordsWriteDescriptor = {
  interface: DwnInterfaceName.Records;
  method: DwnMethodName.Write;
  // ...
  tags: Tags;
};
```

Write:
```json
{
  "schema": "https://schema",
  "protocol": "https://protocol",
  "protocolPath": "some/path",
  "tags": {
    "tag1": ["val1", "val2", "val3"],
    "tag2": [123, 456],
    "tag3": [true]
  }
}
```

Query:
```json
{
  "filter": {
    "schema": "https://schema",
    "tags": {
       "tag1": "val2",
    }
  }
}
```

csuwildcat:
What about starting out with the Phase 1 structure/behavior below, then at some later date we can add top-level array passage of multiple objects to enable ORs later:

```javascript

// Phase 1:

{
  tags: { // if tags is an object, its full set of keys and values are evaluated as exact match
      tag1 ["val1", "val2"] } // Must have 'tag1' with both 'val1' and 'val2'
      tag2: [123] // Must also have 'tag2' and a value '123'
    }
}

// Phase 2:

{
  tags: [ // allow tags to be an array, and if it's an array, each member is an OR
    { // every object's full set of keys and values is evaluated as exact match
      tag1 ["val1", "val2"] } // Must have 'tag1' with both 'val1' and 'val2'
      tag2: [123] // Must also have 'tag2' value '123'
    },
    { // if the first matching object misses, this second object would be tried
      tag2: [123] // Must have both 'val1' and 'val2' 
    }
  ]
}
```

csuwildcat:
What I liked about the proposal above is that 1) the matcher objects retain the same behavior, 2) the only future change is to make `tags` polymorphic but bound to OR, and 3) at no point do we introduce any DSL strings, the whole thing remains an object literal syntax that seems rather intuitive if you know that a top-level array = OR and value arrays = AND.

LiranCohen:
@csuwildcat interesting with the array of tag objects like that for the queries to represent both AND and OR.

Our current typing/representation of filters (although not being used) would read the array as an OR, have to think about what that looks like in the implementation.

[...]

PR #706, opened 2024-03-14
Add `tags` functionality to `Records`.

LiranCohen:
This is a first pass at adding the functionality for `tags` within records as mentioned here https://github.com/TBD54566975/dwn-sdk-js/issues/642.

There will be subsequent PRs for any protocol specific enforcement.

```typescript
{
    published : true,
    schema    : 'post',
    tags      : {
        someString : 'some-value',
        someStringArray : ['some-value1', 'some-value2'],
        someNumber : 54566975,
        someNumberArray : [ 1, 2, 3, 4, 5 ],
        someBoolean: true,
        someOtherBoolean: false
    }
}
```

### Filtering
A filter for a tag is treated as an Equal filter. If any of the values within the tag array match the filter, the record matches.
#### Exact Match

```typescript
{
  filter : {
    schema : 'post',
    tags   : {
      someString: 'some-value'
    }
  }
}
```

#### Range Filters
```typescript
{
  filter : {
    schema : 'post',
    tags   : {
      someString: { gte: 'b', lt: 'c' }, // returns anything that starts with `b`
      someNumber: { gt: 0, lt: 10 }, // (exclusive) returns anything greater than 0 but less than 10 
      someStringArray: { gte: 'b', lt: 'c' }, // returns anything that starts with `b` within the array
      someNumberArray: { gt: 0, lt: 10 } // (exclusive) returns anything greater than 0 but less than 10 within the array 
    }
  }
}
```
#### Prefix Filters
```typescript
{
  filter : {
    schema : 'post',
    tags   : {
      someString: { prefix: 'some' }, // returns anything that starts with the string prefix `some`
      someStringArray: { prefix: 'some' }, // returns anything that starts with the string prefix `some` within the array
    }
  }
}
```


json-schemas/interface-methods/records-filter.json
```diff
@@ -26,6 +26,38 @@
...
+            "properties": {
+              "prefix": {
+                "type": "string"
+              }
+            }
+          },
+          {
+            "$ref": "https://identity.foundation/dwn/json-schemas/string-range-filter.json"
+          }, {
+            "$ref": "https://identity.foundation/dwn/json-schemas/number-range-filter.json"
```
thehenrytsai, line 56: Sanity: where is the `number-range-filter` defined?

src/interfaces/records-write.ts
```diff
@@ -708,6 +714,13 @@ export class RecordsWrite implements MessageInterface<RecordsWriteMessage> {
       entryId   : await RecordsWrite.getEntryId(this.author, this.message.descriptor)
     };
 
+    // we don't want the tags properties to occupy the descriptor index namespace
```
thehenrytsai, line 717: Had to read multiple times to understand the meaning of "occupy the descriptor index namespace", I think.

Consider:
```suggestion
    // in order to avoid name clashes with system-level index keys
```

src/utils/filter.ts
```diff
@@ -43,34 +43,65 @@ export class FilterUtility {
...
+        this.matchFilterIndex(filterValue, indexValue);
+
+      if (!matched) {
+        return false;
+      }
+
+      missingPropertyMatches.delete(filterProperty);
+    }
+
+    return missingPropertyMatches.size === 0;
```
thehenrytsai, line 57: Doesn't look there is a point to keeping `missingPropertyMatches` state, looks like can just return `true`?

src/utils/filter.ts
```diff
@@ -43,34 +43,65 @@ export class FilterUtility {
...
+  }
+
+  /**
+   * Returns true if any of the index values match the filter.
+   *
+   * @param filterValue the filter for a particular property.
+   * @param indexValues an array of values to match the filter against.
+   */
+  private static matchArrayFilterIndex(filterValue: FilterValue, indexValues: string[] | number[] | boolean[]): boolean {
```
thehenrytsai, line 66: Rename:

```suggestion
  private static matchAnyIndexValue(filterValue: FilterValue, indexValues: string[] | number[] | boolean[]): boolean {
```

src/utils/filter.ts
```diff
@@ -43,34 +43,65 @@ export class FilterUtility {
...
+    return false;
+  }
+
+  /**
+   * Returns true if the filter matches the given index value.
+   *
+   * @param filterValue the filter for a particular property.
+   * @param indexValue a single value to match the filter against.
+   */
+  private static matchFilterIndex(filterValue: FilterValue, indexValue: string | number | boolean) : boolean {
```
thehenrytsai, line 82: Rename:

```suggestion
  private static matchIndexValue(filterValue: FilterValue, indexValue: string | number | boolean) : boolean {
```

src/utils/records.ts
```diff
@@ -272,14 +272,54 @@ export class Records {
     return filterCopy;
   }
 
+
+  public static tagsFilterIsPrefix(filter: RecordsWriteTagsFilter): filter is PrefixFilter {
```
thehenrytsai, line 276: Rename: `isPrefixFilter`

src/utils/records.ts
```diff
@@ -272,14 +272,54 @@ export class Records {
...
+    return typeof filter === 'object' && ('prefix' in filter && typeof filter.prefix === 'string');
+  }
+
+  /**
+   * This will create individual keys for each of the tags that look like `tag.tag_property`
+   */
+  public static flattenTags(tags: RecordsWriteTags): KeyValues {
```
thehenrytsai, line 283: Consider rename `buildTagIndexes()`

src/utils/records.ts
```diff
@@ -272,14 +272,54 @@ export class Records {
...
+      const value = tags[property];
+      tagValues[`tag.${property}`] = value;
+    }
+    return tagValues;
+  }
+
+  /**
+   * This will create individual keys for each of the tag filters that look like `tag.tag_filter_property`
+   */
+  public static flattenTagFilters( tags: { [property: string]: RecordsWriteTagsFilter}): Filter {
```
thehenrytsai, line 295: Change this to `private` and optionally rename to something like `convertTagsFilter()` to align with the naming of `convertFilter` and `RecordsWriteTagsFilter`.

src/utils/records.ts
```diff
@@ -272,14 +272,54 @@ export class Records {
...
   /**
    *  Converts an incoming RecordsFilter into a Filter usable by MessageStore.
    *
    * @param filter A RecordsFilter
    * @returns {Filter} a generic Filter able to be used with MessageStore.
    */
   public static convertFilter(filter: RecordsFilter, dateSort?: DateSort): Filter {
-    const filterCopy = { ...filter } as Filter;
+    // we process tags separately from the remaining filters.
+    // this is because we prepend each field within the `tags` object with a `tag.` to bring it into the flat-space filter.
```
thehenrytsai, line 313: "flat-space" is used a term vs "protocol-space", so let's avoid reusing it.

```suggestion
    // this is because we prepend each field within the `tags` object with a `tag.` to avoid name clashing with first-class index keys.
```

tests/features/records-tags.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+        });
+
+        const tagsRecord1ReadReply = await dwn.processMessage(alice.did, tagsRecord1Read.message);
+        expect(tagsRecord1ReadReply.status.code).to.equal(200);
+        expect(tagsRecord1ReadReply.record).to.not.be.undefined;
+        expect(tagsRecord1ReadReply.record!.descriptor.tags).to.deep.equal({ stringTag, numberTag, booleanTag, stringArrayTag, numberArrayTag });
+      });
+    });
+
+    describe('RecordsQuery filter for tags', async () => {
```
thehenrytsai, line 97: Can we add a test or two on scenario of when tags are updated/deleted, the query would behavior correctly? (ie. query matching old/deleted values won't return anything and query matching new values will)

tests/dwn.spec.ts
```diff
@@ -26,14 +28,16 @@ export function testDwnClass(): void {
...
+
       const stores = TestStores.get();
       messageStore = stores.messageStore;
       dataStore = stores.dataStore;
       eventLog = stores.eventLog;
 
       eventStream = TestEventStream.get();
 
-      dwn = await Dwn.create({ messageStore, dataStore, eventLog, eventStream });
+      dwn = await Dwn.create({ didResolver, messageStore, dataStore, eventLog, eventStream });
```
thehenrytsai, line 40: What's the need for doing this?

tests/store/index-level.spec.ts
```diff
@@ -1179,15 +1313,18 @@ describe('IndexLevel', () => {
...
 
   describe('createCursorFromItem', () => {
-    it('throws if cursor value is not a number or boolean', async () => {
+    it('throws if cursor value is a boolean or an array', async () => {
+      // we can only sort by strings or numbers, so arrays or booleans should throw.
       const item: IndexedItem = {
         messageCid : 'message-cid',
         indexes    : {
-          sortProperty: true,
+          sortPropertyBool  : true,
```
thehenrytsai, line 1321: Not saying we should, just for my learning: why can't we sort by boolean?

src/store/index-level.ts
```diff
@@ -76,33 +76,36 @@ export class IndexLevel {
       throw new DwnError(DwnErrorCode.IndexMissingIndexableProperty, 'Index must include at least one valid indexable property');
     }
 
-    const indexOps: LevelWrapperBatchOperation<string>[] = [];
+    const item: IndexedItem = { messageCid, indexes };
+    const indexOpPromises: Promise<LevelWrapperBatchOperation<string>>[] = [];
```
thehenrytsai, line 80: Can we rename: `opCreationPromises`

src/store/index-level.ts
```diff
@@ -76,33 +76,36 @@ export class IndexLevel {
...
-      indexOps.push(partitionOperation);
+      if (Array.isArray(indexValue)) {
+        for (const indexValueItem of indexValue) {
+          const partitionOperationPromise = this.createPutIndexedItemOperation(tenant, item, indexName, indexValueItem);
+          indexOpPromises.push(partitionOperationPromise);
+        }
+        continue;
+      }
+
+      const partitionOperationPromise = this.createPutIndexedItemOperation(tenant, item, indexName, indexValue);
```
thehenrytsai, line 97: Nit: This is a case where I think `else` conveys the intent (thus more readable) better than `continue`.

src/store/index-level.ts
```diff
@@ -111,7 +114,7 @@ export class IndexLevel {
    *  Deletes all of the index data associated with the item.
    */
   async delete(tenant: string, messageCid: string, options?: IndexLevelOptions): Promise<void> {
```
thehenrytsai, line 116: Same comments as put.

src/store/index-level.ts
```diff
@@ -76,33 +76,36 @@ export class IndexLevel {
...
 
-    const indexOps: LevelWrapperBatchOperation<string>[] = [];
+    const item: IndexedItem = { messageCid, indexes };
+    const indexOpPromises: Promise<LevelWrapperBatchOperation<string>>[] = [];
 
     // create an index entry for each property index
     // these indexes are all sortable lexicographically.
+    // the key is the index property's value followed by the messageCid as a tie-breaker.
+    // for example if the property is messageTimestamp the key would look like:
+    // '"2023-05-25T18:23:29.425008Z"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
```
thehenrytsai, line 86: Consider remove, `key` is refactored away, reader probably will be left scratching their head when reading this.

src/store/index-level.ts
```diff
@@ -120,27 +123,77 @@ export class IndexLevel {
...
   }
 
+  /**
+   * Creates an IndexLevel `put` operation for indexing an item, creating a partition by `tenant` and by `indexName`
+   *
+   * The key is indexValue followed by the messageCid as a tie-breaker.
+   * for example if the property is messageTimestamp the key would look like:
+   * '"2023-05-25T18:23:29.425008Z"\u0000bafyreigs3em7lrclhntzhgvkrf75j2muk6e7ypq3lrw3ffgcpyazyw6pry'
+   *
+   * The value is the JSON string of the `item`
```
thehenrytsai, line 155: Consider moving this context info within the implementation (line 165), same for `generateDeleteIndexedItemOperation()`.

tests/features/records-tags.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+            }
+          }
+        });
+
+        const tagsQueryMatchTrueReply = await dwn.processMessage(alice.did, tagsQueryMatchTrue.message);
+        expect(tagsQueryMatchTrueReply.status.code).to.equal(200);
+        expect(tagsQueryMatchTrueReply.entries?.length).to.equal(1);
+        expect(tagsQueryMatchTrueReply.entries![0].recordId).to.equal(tagsRecordTrue.message.recordId);
+
+        // do an exact match for the true tag value
```
thehenrytsai, line 258: Sanity:

```suggestion
        // negative result same tag different value
```

tests/features/records-tags.spec.ts
```diff
@@ -0,0 +1,599 @@
...
+          }
+        });
+
+        const danielRecord = await TestDataGenerator.generateRecordsWrite({
+          author    : alice,
+          published : true,
+          schema    : 'test',
+          tags      : {
+            firstName : 'daniel',
+            score     : 100,
```
thehenrytsai, line 508: Why does Daniel get to score 100? 🤣

thehenrytsai (commented):
🐐 🐐 🐐 ! Thanks for adding this powerful feature. This alone will take dev-ex to the next level!!

json-schemas/interface-methods/records-filter.json
```diff
@@ -26,6 +26,38 @@
...
+          },
+          {
+            "type": "boolean"
+          },
+          {
+            "type": "object",
+            "minProperties": 1,
+            "additionalProperties": false,
+            "properties": {
+              "prefix": {
```
thehenrytsai, line 48: Consider renaming to `startsWith` which is probably more conventional.

tests/features/records-tags.spec.ts
LiranCohen: Nah that's a good call out for a test scenario. I don't think it matters technically since the messageCid would be different, but who knows if some other indexing in the future would do something weird. Not bad to have those logical cases.

tests/dwn.spec.ts
LiranCohen: When working on this feature I linked this repo to `dwn-sql-store` to run the test suite and make sure I didn't run into any gotchas w/ the tag indexing.

I started getting failures that didn't seem related to what I was adding, and when debugging I was getting a DB locked error from the resolver (default `Dwn.create()` uses the LevelDB cache for the resolver).

I did want to spend a little time digging into what changed between the previous release and now, but haven't gotten a chance to. However, this behavior is consistant when I link latest `main` in both repos.

tests/store/index-level.spec.ts
LiranCohen: Sorting by boolean would effectively be the same as splitting the group into 2(true and false) and tie-breaking the rest (sorting by messageCid). It doesn't seem like a sortable property that has any meaningfulness behind it.

tests/features/records-tags.spec.ts
LiranCohen: Because he's an over-achiever 😆

tests/features/records-tags.spec.ts
LiranCohen: Good call out! Although I tested updating and such on the index specific tests, I didn't have 'tags' in the mutable properties array, so the RecordsWrite as an update threw. Added it and some different tests on both the write, and the query side of things.

src/store/index-level.ts
LiranCohen: Went with `putOperationPromises`

json-schemas/interface-methods/records-filter.json
thehenrytsai: Would be good to rename other related type, methods etc. (e.g. `PrefixFilter` -> `StartsWithFilter`, `isPrefixFilter` -> `isStartsWithFilter` etc)

src/store/index-level.ts
thehenrytsai: Definitely not a battle I wanna engaged in, as it is arguable a nit, but just adding context to my thought: I thought the word "creation" is important for describing what the promises are because without it it would appear that the array is holding the result of "execution" of put operations themselves.

tests/features/records-tags.spec.ts
thehenrytsai: Sanity: but it is the same tag (`booleanTag`) though?

tests/features/records-tags.spec.ts
thehenrytsai: Sanity: I see tests with queries after updates now, but don't see one for delete. I'll trust your judgement if you think delete is not necessary here, just sanity to make sure this is indeed intentional. Feel free to resolve this one as is.

tests/dwn.spec.ts
thehenrytsai: Got it, so if I understand correctly, this little change worked around the DB lock issue when test suite is running in `dwn-sql-store` repo.

Feel free to close if I understood correctly.

tests/store/index-level.spec.ts
thehenrytsai: I am not 100% convinced but also am not losing sleep over it, resolving.

src/store/index-level.ts
LiranCohen: Ah yeah I understand what you were trying to convey there. I renamed it and also changed the name of `generateDeleteIndexedItemOperation` to `createDeleteIndexedItemOperation` so that it fits better in the delete method.

tests/features/records-tags.spec.ts
LiranCohen: I ended up adding it now, it's not really necessary for query i don't think, but maybe good to have in case anything blows up with deleting a record w/ tags in general.

tests/features/records-tags.spec.ts
LiranCohen: Whoops, i reworded those comments a bit and added a better scenario description.

json-schemas/interface-methods/records-filter.json
LiranCohen: Done

thehenrytsai (dismissed):
🐐 🐐 🐐 🚀 🚀 🚀

thehenrytsai (approved):
👍 once more!

PR #711, opened 2024-03-21
Enforce protocol defined RecordsWrite `tags`

LiranCohen:
This will allow protocol defined tags to adhere to schema rules by adding a $tags property under the protocolPath definition..
```typescript
const protocolDefinition: ProtocolDefinition = {
  protocol  : 'http://example.com/protocol/withTags',
  published : true,
  types     : {
    foo: {}
  },
  structure: {
    foo: {
      $tags: {
      // tag definitions go here
      }
    }
  },
};
```

Tags are defined as objects within the $tag object. Each property is the tag name, and the object is a JSON Schema definition.

The following definition requires a `draft` boolean tag to be present as well as a `status` tag which can only have the value of `success` or `failure`.
Additional undefined tags are allowed without any validation due to `allowUndefinedTags` set to `true`, this is set to `false` by default.

```typescript
$tags: {
  allowUndefinedTags: true,
  requiredTags: ['status', 'draft'],
  status: {
    type: 'string',
    enum: ['success', 'failure']
  },
  draft: {
    type: 'boolean'
  }
}
```

Types:
✅  string
✅  number
✅  array
✅  integer

Value-testing props:
✅ minLength
✅ maxLength
✅ minimum
✅ exclusiveMinimum
✅ maximum
✅ exclusiveMaximum
✅ minItems
✅ maxItems
✅ minContains
✅ maxContains
✅ uniqueItems

Array evaluation props:
✅  contains (bound to same restricted list value-testing props above)
✅  items (bound to same restricted list value-testing props above)


json-schemas/interface-methods/protocol-rule-set.json
```diff
@@ -106,10 +106,54 @@
...
+      "patternProperties": {
+        "requiredTags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "allowUndefinedTags": {
+          "type": "boolean"
+        },
```
thehenrytsai, line 122: Sanity: it's odd that these two properties are placed within `patternProperties` given that they are predefined. Is it possible to move them out:

```json
      "properties": {
        "requiredTags": {
            "type": "array",
            "items": {
                "type": "string"
            }
        },
        "allowUndefinedTags": {
            "type": "boolean"
        }
      },
```

json-schemas/interface-methods/protocol-rule-set.json
```diff
@@ -106,10 +106,54 @@
...
+        "^(?!requiredTags$|allowUndefinedTags$).*$": {
```
thehenrytsai, line 123: Should we prefix `requiredTags` and `allowUndefinedTags` with `$` for consistency with patterns we've established?

Assuming we do, this regular expression can be simplified.

json-schemas/interface-methods/protocol-rule-set.json
```diff
@@ -106,10 +106,54 @@
...
+      "type": "object",
+      "minProperties": 1,
+      "patternProperties": {
+        "requiredTags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "allowUndefinedTags": {
```
thehenrytsai, line 120: For my understanding: what is a use case for this property?

json-schemas/interface-methods/protocol-rule-set.json
thehenrytsai: Seems to be okay: https://github.com/TBD54566975/dwn-sdk-js/pull/722

tests/features/records-tags.spec.ts
```diff
@@ -779,6 +2679,54 @@ export function testRecordsTags(): void {
...
+            }
+          }
+        });
+
+        const tagsQueryMatchReply = await dwn.processMessage(alice.did, tagsQueryMatch.message);
+        expect(tagsQueryMatchReply.status.code).to.equal(200);
+        expect(tagsQueryMatchReply.entries?.length).to.equal(1);
+        expect(tagsQueryMatchReply.entries![0].recordId).to.equal(tagsRecord1.message.recordId);
+
+
```
thehenrytsai, line 2729: Sanity: is this a dupe of the above test? (Search for `should delete record with tags`)

tests/features/records-tags.spec.ts
```diff
@@ -54,6 +58,1902 @@ export function testRecordsTags(): void {
     });
 
     describe('RecordsWrite with tags', () => {
+      describe('protocol rules', () => {
```
thehenrytsai, line 61: Can we add a sanity test (I did not see it) for making sure the definition is rejected when an invalid JSON schema was given for the `$tags`.

e.g.  a string is assigned to `maximum` instead of a number.

thehenrytsai (commented):
🐐 🐐 🐐 !!!

You've outdone yourself yet again! Pretty smart to reuse JSON schema validation and man are the tests thorough!

json-schemas/interface-methods/protocol-rule-set.json
LiranCohen: 🙏  Thanks!! Yeah that is much better (and probably more performant) great catch!

json-schemas/interface-methods/protocol-rule-set.json
LiranCohen: Yeah I'll do that, that was actually in my initial design, don't really remember why i removed it after reviewing a proposal with Daniel, but I prefer it as well.

json-schemas/interface-methods/protocol-rule-set.json
LiranCohen: To give the option for giving rules around specific necessary tags, as well as allowing open-ended tagging for some use-cases, figured it would give some flexibility there.

tests/features/records-tags.spec.ts
LiranCohen: Yup! Must have been copy-pasta, will remove.

tests/features/records-tags.spec.ts
LiranCohen: GREAT catch! It actually isn't validating that.

Let me see if there is a better way to validate that instead of explicitly writing those rules in the already large JSON schema definition for tags.

thehenrytsai (approved):
🐐 🐐 🐐 🎸 🎸 🎸

Issue #716, opened 2024-03-29
Revisit logic for deleting grant-authorized records when handling a permission revocation `RecordsWrite`


thehenrytsai:
Revisit logic in `postProcessingForCoreRecordsWrite()` for deleting grant-authorized records when handling a permission revocation `RecordsWrite`.

It appears that there was no test for it and it does not look like the code worked:
- not seeing `permissionsGrantId` being an index
- not seeing `this.dataStore` being called to delete actual data
- test coverage is missing for the main delete logic

PR #717, opened 2024-03-29
Swapped to Permissions protocol records for permission grant authorization

thehenrytsai:
There are a LOT of remaining naming inconsistencies due to this change which I will address in a subsequent PR as a final clean up. I can roll all of them into this PR if preferred, but it will be at least 100s more lines of noise scattered throughout the change.


src/handlers/records-write.ts
```diff
@@ -156,14 +160,51 @@ export class RecordsWriteHandler implements MethodHandler {
...
+      const grantAuthorizedMessagesQuery = {
+        permissionsGrantId,
+        dateCreated: { gte: recordsWrite.message.descriptor.messageTimestamp },
+      };
+      const { messages: grantAuthorizedMessagesAfterRevoke } = await this.messageStore.query(tenant, [ grantAuthorizedMessagesQuery ]);
+      const grantAuthorizedMessageCidsAfterRevoke: string[] = [];
+      for (const grantAuthorizedMessage of grantAuthorizedMessagesAfterRevoke) {
+        const messageCid = await Message.getCid(grantAuthorizedMessage);
+        await this.messageStore.delete(tenant, messageCid);
+      }
```
thehenrytsai, line 203: NOTE: this code was never covered in the original code before moving it here. Filed an issue #716 to track this.

src/handlers/records-write.ts
```diff
@@ -156,14 +160,51 @@ export class RecordsWriteHandler implements MethodHandler {
...
+    // If this message is a Permission revocation, we need to delete all grant-authorized messages with timestamp after revocation
+    // TODO: https://github.com/TBD54566975/dwn-sdk-js/issues/716
+    // This code is a direct copy and paste from the original PermissionsRevokeHandler (no longer exists),
+    // but it appears that there was no test for it and it does not look like the code worked:
+    // - not seeing `permissionsGrantId` being an index
+    // - not seeing `this.dataStore` being called to delete actual data
+    // - test coverage is missing for the main delete logic
+    if (recordsWrite.message.descriptor.protocol === PermissionsProtocol.uri &&
+      recordsWrite.message.descriptor.protocolPath === PermissionsProtocol.revocationPath) {
+      const permissionsGrantId = recordsWrite.message.recordId;
```
LiranCohen, line 193: Sanity: I know this part isn't tested, but should this be the parentId of a revocation message?

src/protocols/permissions.ts
```diff
@@ -248,6 +264,70 @@ export class PermissionsProtocol {
...
+    }
+  }
+
+
+
+  /**
+   * Fetches PermissionsGrantMessage with CID `permissionsGrantId`.
+   * @returns the PermissionsGrantMessage with CID `permissionsGrantId` if message exists
+   * @throws {Error} if PermissionsGrantMessage with CID `permissionsGrantId` does not exist
+   */
```
LiranCohen, line 299: Replace `PermissionsGrantMessage` with `PermissionsGrant` and optionally change the descriptions a bit.

```suggestion
  /**
   * Fetches PermissionsGrant with a `recordId` matching `permissionsGrantId`.
   * @returns the PermissionsGrant with a `recordId` matching `permissionsGrantId` if message exists
   * @throws {Error} if PermissionsGrant with a `recordId` matching `permissionsGrantId` does not exist
   */
```

LiranCohen (commented):
🐐 🐐 🐐 A++ as usual.


Just a couple of small questions/nits.

src/protocols/permissions.ts
```diff
@@ -274,4 +354,35 @@ export class PermissionsProtocol {
...
+
+    if (scope.protocol !== undefined) {
+      // `contextId` and `protocolPath` are mutually exclusive
+      if (scope.contextId !== undefined && scope.protocolPath !== undefined) {
+        throw new DwnError(
+          DwnErrorCode.PermissionsProtocolValidateScopeContextIdProhibitedProperties,
+          'Permission grants cannot have both `contextId` and `protocolPath` present'
+        );
+      }
+    }
```
LiranCohen, line 386: For my own understanding, do we disallow these types of things to avoid inconsistent values?

Ie a `contextId` that does not match the `protocol` provided, or `protocol` being redundant with `protocolPath` provided?

json-schemas/interface-methods/records-write-unidentified.json
```diff
@@ -1,8 +1,7 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
   "$id": "https://identity.foundation/dwn/json-schemas/records-write-unidentified.json",
   "type": "object",
-  "additionalProperties": false,
```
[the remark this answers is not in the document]
thehenrytsai: Correct, all child schema extending this parent schema will need to have `unevaluatedProperties: false`  if they disallow additional properties. The moment you have `"additionalProperties": false`, ajv just disallow ignores properties even in extended shema.

src/core/grant-authorization.ts
```diff
@@ -138,17 +107,16 @@ export class GrantAuthorization {
 
     // Check if grant has been revoked
     const query = {
-      interface : DwnInterfaceName.Permissions,
-      method    : DwnMethodName.Revoke,
-      permissionsGrantId,
```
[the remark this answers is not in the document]
thehenrytsai: Sounds good. Done.

src/handlers/records-write.ts
thehenrytsai: You are right. I'll change this now, but there are multiple other issues with this block that requires revisit by #716, e.g.
1. No deletion of actual data.
2. Atomic operation/failure resumption.

src/protocols/permissions.ts
thehenrytsai: Thanks for picking this up!!

src/protocols/permissions.ts
thehenrytsai: Good question man, shows you are paying amazing attention. It is more to do with aligning encryption key scope. Currently encryption key support are derived based on either schema, protocol paths, or context. There is no point in giving out a permission with greater scope if the private key you have access to can only have a less scope.

LiranCohen: Ah ok that makes sense! Didn't think of that for encryption, thanks!

LiranCohen (dismissed):
🏅

LiranCohen (approved):

PR #725, opened 2024-04-23
Only index tags + values if the record is `latestBaseState`

LiranCohen:


src/interfaces/records-write.ts
```diff
@@ -781,7 +781,8 @@ export class RecordsWrite implements MessageInterface<RecordsWriteMessage> {
 
     // in order to avoid name clashes with first-class index keys
     // we build the indexes with `tag.property_name` for each tag property.
-    if (tags !== undefined) {
+    // we only index tags if the message is the latest base state, as that's the only time filtering for tags is relevant.
+    if (tags !== undefined && isLatestBaseState === true) {
```
thehenrytsai, line 785: For my understanding: is this more of an optimization/safe-guard or there is a functional issue?

I'd imagine the `isLatestBaseState` in filter would avoid bad behavior (granted one provides it). I ask because why wouldn't other properties like `published` or `author` fall under the same pattern...?

thehenrytsai (approved):
🚀

PR #734, opened 2024-05-08
getAuthor moved to Message util

LiranCohen:
Move `getAuthor` method to the `records` util so that it can be used on both `RecordsWriteMessage` and `RecordsDeleteMessage`


thehenrytsai (approved):
🚀

PR #735, opened 2024-05-09
Added `prune` protocol action + `ProtocolAuthorization` refactoring

thehenrytsai:
1. Added support for `prune` protocol action.
1. Renamed "ancestor chain" to "record chain" because the chain often contains not just the ancestor records, but also record in concern which leads to confusion.
1. Simplified context and implementation of `constructRecordChain()` 
1. Refactored `authorizeAgainstAllowedActions()` to make it more easily understood.

### DIFF
--- a/src/core/protocol-authorization.ts
+++ b/src/core/protocol-authorization.ts
@@ -77,9 +77,18 @@ export class ProtocolAuthorization {
     incomingMessage: RecordsWrite,
     messageStore: MessageStore,
   ): Promise<void> {
-    // fetch ancestor message chain
-    const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, incomingMessage, messageStore);
+    const existingInitialWrite = await ProtocolAuthorization.fetchInitialWrite(tenant, incomingMessage.message.recordId, messageStore);
+
+    let descendantRecordId;
+    if (existingInitialWrite === undefined) {
+      // NOTE: the incoming message "should" be an initial write at this point,
+      // but we don't check if it is here because it is checked in the RecordsWriteHandler already (even though at a later stage).
+      descendantRecordId = incomingMessage.message.descriptor.parentId;
+    } else {
+      descendantRecordId = incomingMessage.message.recordId;
+    }
+
+    const ancestorMessageChain = await ProtocolAuthorization.constructAncestorMessageChain(tenant, descendantRecordId, messageStore);
 
     // fetch the protocol definition
     const protocolDefinition = await ProtocolAuthorization.fetchProtocolDefinition(
@@ -127,7 +136,7 @@ export class ProtocolAuthorization {
   ): Promise<void> {
     // fetch ancestor message chain
     const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, newestRecordsWrite, messageStore);
+      await ProtocolAuthorization.constructAncestorMessageChain(tenant, newestRecordsWrite.message.recordId, messageStore);
 
     // fetch the protocol definition
     const protocolDefinition = await ProtocolAuthorization.fetchProtocolDefinition(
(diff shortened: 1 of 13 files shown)


csuwildcat (dismissed):

src/core/protocol-authorization.ts
```diff
@@ -198,21 +206,25 @@ export class ProtocolAuthorization {
       tenant,
       incomingMessage,
       ruleSet,
-      [], // ancestor chain is not relevant to subscriptions
+      [], // record chain is not relevant to subscriptions
```
LiranCohen, line 209: NIT while this comment is being updated
```suggestion
      [], // record chain is not relevant to queries or subscriptions
```

src/core/protocol-authorization.ts
```diff
@@ -77,9 +77,18 @@ export class ProtocolAuthorization {
     incomingMessage: RecordsWrite,
     messageStore: MessageStore,
   ): Promise<void> {
-    // fetch ancestor message chain
-    const ancestorMessageChain: RecordsWriteMessage[] =
-      await ProtocolAuthorization.constructAncestorMessageChain(tenant, incomingMessage, incomingMessage, messageStore);
+    const existingInitialWrite = await ProtocolAuthorization.fetchInitialWrite(tenant, incomingMessage.message.recordId, messageStore);
+
+    let descendantRecordId;
```
LiranCohen, line 82: NIT:  I understand why `descendantRecordId ` is named such within the context of `constructRecordChain`, is there a better name for it within this context?

Feels strange calling it a `descendentRecordId` while setting it to the `parentId` of the incoming message.

LiranCohen (commented):
Got through most of it and it looks good!

Just a couple of NITS wrt naming/comments, but code logic looks solid.

Want to have another look at `authorizeAgainstAllowedActions` to make sure I understood it correctly and also review the tests before I stamp, but looks good so far!

src/core/protocol-authorization.ts
```diff
@@ -77,9 +77,18 @@ export class ProtocolAuthorization {
...
+    if (existingInitialWrite === undefined) {
+      // NOTE: the incoming message "should" be an initial write at this point,
+      // but we don't check if it is here because it is checked in the RecordsWriteHandler already (even though at a later stage).
```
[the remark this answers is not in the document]
thehenrytsai: @LiranCohen , you are spot on. This one is tough, if I don't say anything here, then people reading the condition `if (existingInitialWrite === undefined) {` could be wondering:

Hm... why do we assume that the write is an initial write just because an initial write doesn't already exist? What if this write is NOT an initial write, would looking at the `parentId` introduce a bug?

For this reason I actually attempted to reshuffle the logic in `RecordsWriteHandler` but backed out because it comes even more involved with a multiple tests failing. May not be too hard to fix, but decided to back out to get mini-wins iteratively.

I won't have the battle if you think we should remove entirely, even better if you can suggest better wording! Perhaps:

```suggestion
      // NOTE: we can assume this message is an initial write because 
      // we check in the `RecordsWriteHandler` and fail if the incoming message is not an initial write and no initial write exists (even though at a later stage).
```

src/core/protocol-authorization.ts
thehenrytsai: Done.

src/core/protocol-authorization.ts
thehenrytsai: I thought hard and couldn't come up with something better... we could do this also if you prefer:
```suggestion
    let recordChain;
    if (existingInitialWrite === undefined) {
      recordChain = await ProtocolAuthorization.constructRecordChain(tenant, incomingMessage.message.descriptor.parentId, messageStore);
    } else {
      recordChain = await ProtocolAuthorization.constructRecordChain(tenant, incomingMessage.message.recordId, messageStore);
    }
```

src/core/protocol-authorization.ts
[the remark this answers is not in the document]
LiranCohen: Not a hill I'll die on, but maybe adding more verbosity would help clear it up vs removing it completely?
Something along the lines of:

```suggestion
// NOTE: we can assume this message is an initial write because an existing initial write does not exist.
// Additionally, we check further down in the `RecordsWriteHandler` if the incoming message is an initialWrite, so we don't check explicitly here to avoid an unnecessary duplicate check.
```

Feel free to change the wording a bit, but I think the intent is a little more clear? It's not too important of a hill for me so I'll leave it to your discretion.

src/core/protocol-authorization.ts
LiranCohen: Yeh I'm fine with the change you proposed, or calling it something like `rootRecordId` and a comment is also fine.

LiranCohen (dismissed):
🚀 🚀

src/core/protocol-authorization.ts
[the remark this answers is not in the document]
thehenrytsai: Took you suggestion.

LiranCohen (approved):
🚀

PR #737, opened 2024-05-09
Upgraded to ESLint v9

thehenrytsai:
1. `.eslintrc.cjs` is no longer supported, moving to `eslint.confg.cjs` with the new "flat config system".
1. `-ext` is no longer supported in CLI
1. `.eslintignore` is no longer supported

### DIFF
--- a/eslint.config.cjs
+++ b/eslint.config.cjs
@@ -1,17 +1,29 @@
-module.exports = {
-  parser        : '@typescript-eslint/parser',
-  parserOptions : {
-    ecmaVersion : 'latest', // Allows the use of modern ECMAScript features
-    sourceType  : 'module', // Allows for the use of imports
+const tsParser = require("@typescript-eslint/parser");
+
+const tsPlugin = require("@typescript-eslint/eslint-plugin");
+const todoPlzPlugin = require('eslint-plugin-todo-plz');
+
+module.exports = [{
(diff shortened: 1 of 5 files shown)


csuwildcat (approved):

eslint.config.cjs
```diff
@@ -1,17 +1,29 @@
...
+    "@typescript-eslint": tsPlugin,
+    'todo-plz': todoPlzPlugin // for enforcing TODO formatting to require "github.com/TBD54566975/dwn-sdk-js/issues/"
   },
+  files: [
+    '**/*.ts'
+  ],
+  // IMPORTANT and confusing: `ignores` only exclude files from the `files` setting.
+  // To exclude *.js files entirely, you need to have a separate config object altogether. (See another `ignores` below.)
+  ignores: [
+    '**/*.d.ts',
```
LiranCohen, line 25: For my own knowledge, why are type definition files ignored?

LiranCohen (approved):
🚀

eslint.config.cjs
thehenrytsai: Hey @LiranCohen , those files are generated so we don't want to run the linter over them right?

LiranCohen: Ah ok yeah that makes sense!

PR #738, opened 2024-05-13
upgrade `@web5/dids` to `v1.0.2` which includes `did:ion` updates

LiranCohen:
As per @shamilovtim there was an update to the ION pacakge within `@web5/dids`, this will allow bubbling it up to `@web5/agent`


thehenrytsai (approved):
🚀

PR #743, opened 2024-05-15
Replaced bespoke key derivation algorithm to use RFC 5869 HKDF

thehenrytsai:
1. Replaced bespoke key derivation algorithm to use RFC 5869 HKDF to align with the encryption text in DWN spec (that I wrote yesterday and merged this morning 🤣)
2. Other minor updates to align with the DWN spec.
3. Refactoring.

### DIFF
--- a/src/utils/hd-key.ts
+++ b/src/utils/hd-key.ts
@@ -115,8 +115,7 @@ export class HdKey {
   }
 
   /**
-   * Parses the given key derivation path.
-   * @returns Path segments if successfully validate the derivation path.
+   * Validates that no empty strings exist within the derivation path segments array.
    * @throws {DwnError} with `DwnErrorCode.HdKeyDerivationPathInvalid` if derivation path fails validation.
    */
   private static validateKeyDerivationPath(pathSegments: string[]): void {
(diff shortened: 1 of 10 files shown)


src/utils/hd-key.ts
```diff
@@ -56,4 +59,69 @@ export class HdKey {
...
+
+    // Convert from ArrayBuffer to Uint8Array.
+    const derivedKeyBytes = new Uint8Array(derivedKeyBuffer);
+    return derivedKeyBytes;
+  }
+
+  /**
+   * Parses the given key derivation path.
+   * @returns Path segments if successfully validate the derivation path.
+   * @throws {DwnError} with `DwnErrorCode.HdKeyDerivationPathInvalid` if derivation path fails validation.
```
LiranCohen, line 120: We aren't returning any path segments here. Since it's only checking if there are empty strings, figured it would be good to put in the comments.

```suggestion
   * Validates that no empty strings exist within the derivation path segments array.
   * @throws {DwnError} with `DwnErrorCode.HdKeyDerivationPathInvalid` if derivation path fails validation.
```

LiranCohen (changes requested):
Good stuff! Just caught a small inconsistency in the JSDoc comment

src/utils/hd-key.ts
thehenrytsai: Thanks! Keen eyes as always!

LiranCohen (approved):
🚀 🚀 🚀

PR #748, opened 2024-05-30
Permissions Grants for `MessagesGet`

LiranCohen:
This PR modifies `MessagesGet` to better suit it's use and support permissions.
- MessagesGet now only supports getting a single message
  - If a message has data, it will be provided via a `data` method similarly to `ReocrdsRead`
- Can Grant Permissions for MessagesGet scoped
  - Allows unrestricted scope (no protocol scope spcified)
  - If scoped to a protocol and the message to get is a Permissions-Protocol record, it will match against the scope.
  - Delete messages are matched against the protocol of the `RecordsWrite` they reference


src/core/dwn-error.ts
```diff
@@ -20,10 +20,6 @@ export enum DwnErrorCode {
   AuthorizationNotGrantedToAuthor = 'AuthorizationNotGrantedToAuthor',
   ComputeCidCodecNotSupported = 'ComputeCidCodecNotSupported',
   ComputeCidMultihashNotSupported = 'ComputeCidMultihashNotSupported',
-  DidMethodNotSupported = 'DidMethodNotSupported',
```
LiranCohen, line 23: Found a few unused error codes, so I went through and removed others.

thehenrytsai: 😮 🙏
Did you write a script for this?

src/core/messages-grant-authorization.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    throw new DwnError(DwnErrorCode.MessagesGetVerifyScopeFailed, 'record message failed scope authorization');
+  }
+
+  /**
+   * Get the RecordsWriteMessage associated with the given RecordsDeleteMessage in order to authorize access.
+   */
+  private static async getRecordsWriteMessageToAuthorize(
+    tenant: string,
+    message: RecordsDeleteMessage,
```
thehenrytsai, line 112: Consider a slight change in method signature and name, which would be more generic, readable, and potentially reusable.

```suggestion
  private static async getNewestRecordsWrite(
    tenant: string,
    recordId: string,
```

Optional extra mile, with the above change the method become generic thus can be moved else for reuse, though looks like we have a few variants of this but not quite the same.

src/core/messages-grant-authorization.ts
```diff
@@ -0,0 +1,132 @@
...
+          const grant = await PermissionsProtocol.fetchGrant(tenant, messageStore, recordsWriteMessage.descriptor.parentId!);
+          permissionScope = grant.scope;
+          return;
+        } else if (recordsWriteMessage.descriptor.protocolPath === PermissionsProtocol.grantPath) {
+          const grant = await PermissionGrant.parse(recordsWriteMessage as DataEncodedRecordsWriteMessage);
+          permissionScope = grant.scope;
+        } else if (recordsWriteMessage.descriptor.protocolPath === PermissionsProtocol.requestPath) {
+          const request = await PermissionRequest.parse(recordsWriteMessage as DataEncodedRecordsWriteMessage);
+          permissionScope = request.scope;
+        }
```
thehenrytsai, line 87: Nit: Consider encapsulate this in a method for readability, perhaps `getScopeFromPermissionRecord()`

src/core/messages-grant-authorization.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      // we check if the protocol is the internal PermissionsProtocol for further validation
+      if (recordsWriteMessage.descriptor.protocol === PermissionsProtocol.uri) {
+        // get the permission scope from the permission message
+        // if the permission message is a revocation, the scope is fetched from the grant that is being revoked
+        let permissionScope!: PermissionScope;
+        if (recordsWriteMessage.descriptor.protocolPath === PermissionsProtocol.revocationPath) {
+          const grant = await PermissionsProtocol.fetchGrant(tenant, messageStore, recordsWriteMessage.descriptor.parentId!);
+          permissionScope = grant.scope;
+          return;
```
thehenrytsai, line 80: Sanity: unintended `return;`? This would authorize any grant to be fetched?

src/core/messages-grant-authorization.ts
```diff
@@ -0,0 +1,132 @@
...
+import { PermissionsProtocol } from '../protocols/permissions.js';
+import { Records } from '../utils/records.js';
+import { DwnError, DwnErrorCode } from './dwn-error.js';
+import { DwnInterfaceName, DwnMethodName } from '../enums/dwn-interface-method.js';
+
+export class MessagesGrantAuthorization {
+
+  /**
+   * Authorizes a MessagesGetMessage using the given permission grant.
+   * @param messageStore Used to check if the given grant has been revoked.
```
thehenrytsai, line 21: Nit:
```suggestion
   * @param messageStore Used to check if the given grant has been revoked; and to fetch related RecordsWrites if needed.
```

src/handlers/messages-get.ts
```diff
@@ -26,55 +32,76 @@ export class MessagesGetHandler implements MethodHandler {
...
+      } else {
+        // otherwise check the data store for the associated data
+        const result = await this.dataStore.get(tenant, recordsWrite.recordId, recordsWrite.descriptor.dataCid);
+        if (result?.dataStream !== undefined) {
+          entry.message.data = result.dataStream;
+        } else {
+          // if there is no data, return with the data property undefined
+          // when records are deleted, their data is removed from the data store but the message remains in the message store
+          delete entry.message.data;
+        }
```
thehenrytsai, line 69: Sanity: the `else block seems unecessary?

src/handlers/messages-get.ts
```diff
@@ -26,55 +32,76 @@ export class MessagesGetHandler implements MethodHandler {
...
     return {
-      status  : { code: 200, detail: 'OK' },
-      entries : messages
+      status: { code: 200, detail: 'OK' },
+      entry
     };
   }
+
+  /**
+   * @param messageStore Used to check if the grant has been revoked.
```
thehenrytsai, line 80: Nit:
```suggestion
   * @param messageStore Used to fetch related permission grant, permission revocation, and/or RecordsWrites for permission scope validation.
```

thehenrytsai (commented):
Love how it is sorta in parity with `RecordsRead`! Except for the name 😉.

A few small comments.

src/core/messages-grant-authorization.ts
LiranCohen: 🔥 I moved it over to the `RecordsWrite` interface class, reads better and may be more useful elsewhere.

src/core/messages-grant-authorization.ts
LiranCohen: 😆 I actually had it that way but wasn't sure if it was useful as it's own method (i even forgot to remove the error code!!), adding it to the `PermissionsProtocol` class + tests.

src/core/messages-grant-authorization.ts
LiranCohen: Good catch! This was a mess up, fixed in the other refactored method.

src/core/dwn-error.ts
LiranCohen: No script, lol Although it would be nice to validate this during CI/lint.

But I spotted a few i knew weren't there and decided to go through the list haha.

thehenrytsai (approved):
🐐 🚀

PR #750, opened 2024-06-06
RecordsPermissionScope Grants must be scoped to a protocol

LiranCohen:
Modifies `RecordsPermissionScope` to require scoping to a specific protocol. This also removes the ability to scope to a schema as you would just scope to a protocolPath instead.

The motivation behind this was to make querying for all Messages tied to a specific protocol for subset sync. Without this change getting grants, and more importantly the grant revocations which have unrestricted scopes would be challenging/not performant.

Since we are planning on getting rid of flat-spaced records anyway, this is a step in the right direction.

- Remove `schema` from `RecordsPermissionScope`
- `RecordsPermissionScope` requires `protocol` to be set
- The `protocol` provided within the `RecordsPermissionScope` is also added as a `protocol` tag to the `RecordsWriteMessage` descriptor.
- When validating the Scope of a Grant `RecordsWriteMessage`, ensure the `protocol` tag on the record is the same as the scope `protocol` defined in the `RecordsPermissionScope`.

`grant` is covered by this PR.
`request` and `revoke` will come in subsequent PRs


src/types/permission-types.ts
```diff
@@ -68,25 +68,34 @@ export type PermissionRevocationData = {
...
+export type PermissionScope = ProtocolPermissionScope | MessagesPermissionScope | EventsPermissionScope | RecordsPermissionScope;
+
+export type ProtocolPermissionScope = {
+  interface: DwnInterfaceName.Protocols;
+  method: DwnMethodName.Configure | DwnMethodName.Query;
+};
+
+export type MessagesPermissionScope = {
+  interface: DwnInterfaceName.Messages;
+  method: DwnMethodName.Get;
```
LiranCohen, line 80: NOTE: Will add `protocol: string;` to this scope when it is implemented in a subsequent PR.

csuwildcat (approved):

src/core/records-grant-authorization.ts
```diff
@@ -138,37 +138,12 @@ export class RecordsGrantAuthorization {
...
-   * Verifies a protocol record against the scope of the given grant.
-   */
-  private static verifyProtocolRecordScope(
     recordsWriteMessage: RecordsWriteMessage,
     grantScope: RecordsPermissionScope
   ): void {
-    // Protocol records must have grants specifying the protocol
-    if (grantScope.protocol === undefined) {
-      throw new DwnError(
-        DwnErrorCode.RecordsGrantAuthorizationScopeMissingProtocol,
```
thehenrytsai, line 168: Remove this value from `DwnErrorCode`.

src/protocols/permissions.ts
```diff
@@ -355,34 +365,38 @@ export class PermissionsProtocol {
...
       return;
     }
+
     // else we are dealing with a RecordsPermissionScope
 
-    // `schema` scopes may not have protocol-related fields
-    if (scope.schema !== undefined) {
-      if (scope.protocol !== undefined || scope.contextId !== undefined || scope.protocolPath) {
-        throw new DwnError(
-          DwnErrorCode.PermissionsProtocolValidateScopeSchemaProhibitedProperties,
```
thehenrytsai, line 372: Remove this error code from `DwnErrorCode`

tests/features/permissions.spec.ts
```diff
@@ -207,216 +207,148 @@ export function testPermissions(): void {
...
-          { dataStream: DataStream.fromBytes(grant.permissionGrantBytes) }
-        );
-        expect(schemaAndProtocolGrantReply.status.code).to.eq(400);
-        expect(schemaAndProtocolGrantReply.status.detail).to.contain(DwnErrorCode.PermissionsProtocolValidateScopeContextIdProhibitedProperties);
+        try {
+          PermissionsProtocol['validateScope'](permissionScope, grantRecordsWrite.message);
+          expect.fail('Expected to throw');
+        } catch (error:any) {
+          expect(error.message).to.include(DwnErrorCode.PermissionsProtocolValidateScopeContextIdProhibitedProperties);
+        }
```
thehenrytsai, line 351: Consider:

```suggestion
        expect(
          () => PermissionsProtocol['validateScope'](permissionScope, grantRecordsWrite.message)
        ).to.throw(DwnErrorCode.PermissionsProtocolValidateScopeContextIdProhibitedProperties);
```

tests/features/permissions.spec.ts
```diff
@@ -207,216 +207,148 @@ export function testPermissions(): void {
...
+      try {
+        PermissionsProtocol.validateSchema(message, dataBytes!);
+        expect.fail('Expected to throw');
+      } catch (error:any) {
+        expect(error.message).to.include(DwnErrorCode.PermissionsProtocolValidateSchemaUnexpectedRecord);
+        expect(error.message).to.include('invalid/path');
+      }
+    });
+
+    describe('validateScope', async () => {
```
thehenrytsai, line 248: My main worry here is that the original feature tests are replaced with just unit tests. We don't actually know if the feature really works or not because there is no proof that `validateScope` is called.

At minimum as a sanity:
- Add a spy on `validateScope` in one of the permission tests above, probably the happy path, and make sure it is called.

tests/features/permissions.spec.ts
LiranCohen: Much cleaner

tests/features/permissions.spec.ts
LiranCohen: Doh! I actually noticed this while working one of the subsequent PRs and was sure i back ported it to this one. Will fix.

src/core/records-grant-authorization.ts
LiranCohen: Good catch, done.

src/protocols/permissions.ts
LiranCohen: Good catch, done.

thehenrytsai (approved):
🚀

PR #754, opened 2024-06-12
RecordsPermissionScope Revocation must be scoped to a protocol

LiranCohen:
In order to get Permission Revocation messages associated with a specific protocol from the `MessageStore` it is necessary for the revocation messages to be tagged with the `protocol` from the grant scope.

In this PR:
- added an optional `protocol` property to the `PermissionRevocationCreateOptions` which normalizes and adds the provided protocol as a tag.
- added a `preProcessingForCoreRecordsWrite` method to the `RecordsWriteHandler` where if the incoming message is a revocation message, the original grant will be fetched and the protocol tag from the revocation will be compared to the grant scope protocol.

### DIFF
--- a/src/protocols/permissions.ts
+++ b/src/protocols/permissions.ts
@@ -390,7 +390,7 @@ export class PermissionsProtocol {
   /**
    * Type guard to determine if the scope is a record permission scope.
    */
-  public static isRecordPermissionScope(scope: PermissionScope): scope is RecordsPermissionScope {
+  private static isRecordPermissionScope(scope: PermissionScope): scope is RecordsPermissionScope {
     return scope.interface === 'Records';
   }
 
--- a/src/protocols/permissions.ts
+++ b/src/protocols/permissions.ts
@@ -66,6 +66,10 @@ export type PermissionRevocationCreateOptions = {
    */
   signer?: Signer;
   grantId: string;
+  /**
+   * If the grant was scoped toa protocol, the protocol must be included in the revocation.
+   */
+  protocol?: string;
   dateRevoked?: string;
 
   // remaining properties are contained within the data payload of the record
@@ -279,6 +283,12 @@ export class PermissionsProtocol {
       description: options.description,
     };
 
+    let permissionTags = undefined;
+    if (options.protocol !== undefined) {
+      const protocol = normalizeProtocolUrl(options.protocol);
+      permissionTags = { protocol };
+    }
+
     const permissionRevocationBytes = Encoder.objectToBytes(permissionRevocationData);
     const recordsWrite = await RecordsWrite.create({
       signer          : options.signer,
@@ -287,6 +297,7 @@ export class PermissionsProtocol {
       protocolPath    : PermissionsProtocol.revocationPath,
       dataFormat      : 'application/json',
       data            : permissionRevocationBytes,
+      tags            : permissionTags,
     });
 
     return {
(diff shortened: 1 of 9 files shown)


csuwildcat (dismissed):

src/protocols/permissions.ts
```diff
@@ -66,6 +66,10 @@ export type PermissionRevocationCreateOptions = {
    */
   signer?: Signer;
   grantId: string;
+  /**
+   * If the grant being revoked was scoped to a protocol, the protocol must be included in the revocation.
+   */
+  protocol?: string;
```
thehenrytsai, line 72: Appreciate the additional documentation for the non-trivial requirement here! I wonder if it would be an improvement if we pass in the grant itself instead of the `grantId` so that this info can be inferred?

It is almost guaranteed that a dev is not gonna know they need to pass in protocol for protocol scoped grants.

And I think it is a valid assumption that surely whomever is revoking the grant has the grant they want to revoke, because how else would they know what the `protocol` value is to pass in LOL!

Totally fine if you want to address this in a separate PR even if you agree, I try to avoid bloating existing PRs myself!

tests/features/permissions.spec.ts
```diff
@@ -60,6 +60,123 @@ export function testPermissions(): void {
...
+
+      // createRequest with a protocol
+      const requestWrite = await PermissionsProtocol.createRequest({
+        signer      : Jws.createSigner(alice),
+        description : 'Requesting to write',
+        delegated   : false,
+        scope       : {
+          interface : DwnInterfaceName.Records,
+          method    : DwnMethodName.Write,
+          protocol  : 'https://example.com/protocol/test'
```
thehenrytsai, line 74: Nit: considering assigning to a variable, too many occurrences... 🤣

tests/handlers/records-write.spec.ts
```diff
@@ -91,6 +91,28 @@ export function testRecordsWriteHandler(): void {
         await dwn.close();
       });
 
+      it('should call preProcessingForCoreRecordsWrite after authorization and before storage', async () => {
+        // We create spy and stub the authorization, preProcessingForCoreRecordsWrite and processMessageWithDataStream methods
```
thehenrytsai, line 95: NIt:
```suggestion
        // We create spy or stub for authorization, preProcessingForCoreRecordsWrite and processMessageWithDataStream methods
```

thehenrytsai (dismissed):
Legit! And thanks for breaking up the PRs! Really makes reviewing a much pleasant and efficient experience.

Happy to chat if you'd like!

src/protocols/permissions.ts
LiranCohen: That's a good call, I like it! I'll do it in this PR just so that we actually get it done, i think it's better devex.

tests/handlers/records-write.spec.ts
LiranCohen: Done

tests/features/permissions.spec.ts
LiranCohen: I modified it in some of them where it made sense, in the others i just kept it.

src/protocols/permissions.ts
LiranCohen: Done

tests/features/permissions.spec.ts
thehenrytsai: Our wavelength are identical sometimes I take it for granted!

thehenrytsai (approved):
🐐 🚀 

Optional nit:

Do you think passing the message JSON directly for revocation creation is easier as dev-ex?

Non-blocking, code can be merged as is, and can be done later even if you agree.

Upon reviewing your code, I realized some of the typing isn't ideal (my code) and I probably meant to fix it but just forgot. A small PR will quickly follow this one it's merged.

LiranCohen:
> Do you think passing the message JSON directly for revocation creation is easier as dev-ex?

Yeah I was debating this, but wasn't sure which object the user would have more readily available so i opted with the parsed type.

I'll merge this in and we can improve it a bit in a subsequent PR. 🙏

PR #758, opened 2024-06-13
Add testing util to poll for async event testing

LiranCohen:
Due to the async nature of testing `EventStream` and Subscriptions, and some external pub/sub systems taking a longer time to publish and/or receive a message, we have added a helper method to help wait for a test case to assert.

Additionally we modified the tests to be able to more accurately check negative test cases where testing the absence of something which arrives async could be challenging.

- `pollUntilSuccessOrTimeout` will execute a delegate function passed to it, if it fails it will sleep `retrySleep` ms, and will retry. If the total retry time exceeds `timeout`, it will finally fail.

### DIFF
--- a/tests/event-log/event-stream.spec.ts
+++ b/tests/event-log/event-stream.spec.ts
@@ -3,11 +3,11 @@ import type { KeyValues } from '../../src/types/query-types.js';
 import type { MessageEvent } from '../../src/types/subscriptions.js';
 
 import { TestEventStream } from '../test-event-stream.js';
+import { TestTimingUtils } from '../utils/test-timing-utils.js';
 import { Message, TestDataGenerator, Time } from '../../src/index.js';
 
 import chaiAsPromised from 'chai-as-promised';
 import chai, { expect } from 'chai';
-import { TestTimingUtils } from '../utils/test-timing-utils.js';
 
 chai.use(chaiAsPromised);
 
--- a/tests/event-log/event-stream.spec.ts
+++ b/tests/event-log/event-stream.spec.ts
@@ -22,125 +22,127 @@ chai.use(chaiAsPromised);
 // we add an alternate subscription to await results within to give the EventStream ample time to process the message.
 // Additionally in some of these cases the order in which messages are sent to be processed or checked may matter, and they are noted as such.
 
-describe('EventStream', () => {
-  // saving the original `console.error` function to re-assign after tests complete
-  const originalConsoleErrorFunction = console.error;
-  let eventStream: EventStream;
+export function testEventStream(): void {
+  describe('EventStream', () => {
+    // saving the original `console.error` function to re-assign after tests complete
+    const originalConsoleErrorFunction = console.error;
+    let eventStream: EventStream;
 
-  before(async () => {
-    eventStream = TestEventStream.get();
-    await eventStream.open();
+    before(async () => {
+      eventStream = TestEventStream.get();
+      await eventStream.open();
 
-    // do not print the console error statements from the emitter error
-    console.error = (_):void => { };
-  });
-
-  beforeEach(() => {
-    sinon.restore();
-  });
-
-  after(async () => {
-    sinon.restore();
-    console.error = originalConsoleErrorFunction;
-    // Clean up after each test by closing and clearing the event stream
-    await eventStream.close();
-  });
-
-  it('emits all messages to each subscriptions', async () => {
-    // Scenario: We create 2 separate subscriptions that listen to all messages
-    // and we emit 3 messages. We expect both subscriptions to receive all 3 messages.
-
-    const messageCids1: string[] = [];
-    const handler1 = async (_tenant: string, event: MessageEvent, _indexes: KeyValues): Promise<void> => {
-      const { message } = event;
-      const messageCid = await Message.getCid(message);
-      messageCids1.push(messageCid);
-    };
-
-    const messageCids2: string[] = [];
-    const handler2 = async (_tenant: string, event: MessageEvent, _indexes: KeyValues): Promise<void> => {
-      const { message } = event;
-      const messageCid = await Message.getCid(message);
-      messageCids2.push(messageCid);
-    };
-
-    const subscription1 = await eventStream.subscribe('did:alice', 'sub-1', handler1);
-    const subscription2 = await eventStream.subscribe('did:alice', 'sub-2', handler2);
-
-    const message1 = await TestDataGenerator.generateRecordsWrite({});
-    const message1Cid = await Message.getCid(message1.message);
-    eventStream.emit('did:alice', { message: message1.message }, {});
-    const message2 = await TestDataGenerator.generateRecordsWrite({});
-    const message2Cid = await Message.getCid(message2.message);
-    eventStream.emit('did:alice', { message: message2.message }, {});
-    const message3 = await TestDataGenerator.generateRecordsWrite({});
-    const message3Cid = await Message.getCid(message3.message);
-    eventStream.emit('did:alice', { message: message3.message }, {});
-
-    // Use the TimingUtils to poll until the expected results are met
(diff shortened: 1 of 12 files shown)


tests/event-log/event-stream.spec.ts
```diff
@@ -57,45 +77,80 @@ describe('EventStream', () => {
...
+    await subscription2.close();
   });
 
-  it('does not emit messages if event stream is closed', async () => {
-    const messageCids: string[] = [];
-    const handler = async (_tenant: string, event: MessageEvent, _indexes: KeyValues): Promise<void> => {
-      const { message } = event;
-      const messageCid = await Message.getCid(message);
-      messageCids.push(messageCid);
+  it('does not receive messages if event stream is closed', async () => {
```
LiranCohen, line 147: I'm considering removing this test completely. I don't like the hack here, but can't think of a better way to test it here.

We could implement this test in each concrete implementations of the EventStream.

tests/test-suite.ts
```diff
@@ -43,10 +44,12 @@ export class TestSuite {
     messageStore?: MessageStore,
     dataStore?: DataStore,
     eventLog?: EventLog,
+    eventStream?: EventStream,
```
LiranCohen, line 47: I am considering renaming this to `runExternalDependenciesTests` or something to that effect.

I was originally thinking to pull the `eventStream` tests into their own `TestSuite` method, but it seems many of them overlap.

Any thoughts @thehenrytsai

csuwildcat (approved):

tests/event-log/event-stream.spec.ts
thehenrytsai: I am okay for us to remove this if the code paths are covered by other  tests.

Alternatively, you could change the override in `test-suite.ts` from
`eventStream?: EventStream` -> `eventStreamCreator?: () => EventStream`

And create an addition `eventStream` in this test for control.

thehenrytsai (commented):
🐐 
The number the tests you have for subscription is astounding!

tests/utils/test-timing-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+   * @param retrySleep the interval in milliseconds to wait before retrying the delegate function.
+   * @param timeout the maximum time in milliseconds to wait before timing out the delegate function.
+   *
+   * @throws {Error} `Operation timed out` if the timeout is exceeded.
+   */
+  static async pollUntilSuccessOrTimeout(
+    delegate: () => Promise<any>,
+    retrySleep: number = TestTimingUtils.pollRetrySleep,
+    timeout: number = TestTimingUtils.pollTimeout,
+  ): Promise<any> {
```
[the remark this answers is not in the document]
LiranCohen: Great call!

tests/handlers/records-subscribe.spec.ts
```diff
@@ -315,24 +336,52 @@ export function testRecordsSubscribeHandler(): void {
...
+            expect(bobMessages.length).to.equal(1);
+            expect(bobMessages).to.have.members([ chatRecordForBobCid ]);
+          });
 
-          expect(messageCids.length).to.equal(1, 'before delete');
-          expect(messageCids[0]).to.equal(await Message.getCid(chatRecordForBob.message));
+          await TestTimingUtils.pollUntilSuccessOrTimeout(async () => {
+            expect(carolMessages.length).to.equal(2);
+            expect(carolMessages).to.have.members([ chatRecordForCarol1Cid, chatRecordForCarol2Cid ]);
+          });
```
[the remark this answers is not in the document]
LiranCohen: Maybe re-thinking how this scenario is describe would make it easier to understand, possibly using a different protocol definition. It's a copy of the scenario test from `query`.

The intent is to show that users subscribed without any role still receive messages that are destined for them. But in this particular case they aren't issued the `participant` role so they wouldn't get the `thread`

tests/handlers/records-subscribe.spec.ts
```diff
@@ -425,120 +490,43 @@ export function testRecordsSubscribeHandler(): void {
...
           }
 
-          // there should not be any messages in the subscription without a friend role.
-          expect(noRoleRecords.length).to.equal(0);
-
-          // should have all chat messages
-          expect(recordIds).to.have.members(chatRecordIds);
-        });
-
-        it('allows protocol authorized subscriptions', async () => {
```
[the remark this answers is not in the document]
LiranCohen: All record authorization is now forced to be `protocol` scoped. So the other tests already cover it.

tests/features/author-delegated-grant.spec.ts
```diff
@@ -514,19 +516,65 @@ export function testAuthorDelegatedGrant(): void {
...
+      });
+      const chatRecord2Reply = await dwn.processMessage(bob.did, chatRecord2.message, { dataStream: chatRecord2.dataStream });
+      expect(chatRecord2Reply.status.code).to.equal(202);
+
+      await TestTimingUtils.pollUntilSuccessOrTimeout(async () => {
+        expect(subscriptionChatRecords.size).to.equal(2);
+        expect([ ...subscriptionChatRecords ]).to.have.members([ chatRecord1.message.recordId, chatRecord2.message.recordId ]);
+      });
+
+      //TODO: https://github.com/TBD54566975/dwn-sdk-js/issues/759
```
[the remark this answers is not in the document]
LiranCohen: I took out the duplicate test behavior/comment and just left it under `records-subscribe.spec.ts` for now.

I think it *might* be a limitation of the design, but I'd like for us to know it's an issue and track a way to solve it.

When building up filters for `protocol authorized subscribe` some of the filters include `published: false` and this is why the `RecordsDelete` isn't coming through (published is a mutable property) ... However I do think that if a person is subscribing to a protocol path using a role they should be able to receive the deletes necessary to that subscription, so I'll leave the commented test case and issue.

tests/event-log/event-stream.spec.ts
LiranCohen: I think removing it is the best bet, it's tested for `EventEmitterStream` to make sure there are no more event listeners.

tests/test-suite.ts
LiranCohen: Went with `runInjectableDependentTests`

tests/event-log/event-stream.spec.ts
```diff
@@ -20,151 +20,129 @@ chai.use(chaiAsPromised);
...
+      const message1Cid = await Message.getCid(message1.message);
+      eventStream.emit('did:alice', { message: message1.message }, {});
+      const message2 = await TestDataGenerator.generateRecordsWrite({});
+      const message2Cid = await Message.getCid(message2.message);
+      eventStream.emit('did:alice', { message: message2.message }, {});
+      const message3 = await TestDataGenerator.generateRecordsWrite({});
+      const message3Cid = await Message.getCid(message3.message);
+      eventStream.emit('did:alice', { message: message3.message }, {});
+
+      // Use the TimingUtils to poll until the expected results are met
```
thehenrytsai, line 81: Left over `TimingUtils` references.

LiranCohen: Good catch, fixed.

tests/handlers/records-subscribe.spec.ts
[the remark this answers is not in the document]
LiranCohen: re introduced this test.

thehenrytsai (approved):
🚀 

Consider renaming the test to "root/context role authorized ..." for clarity in a future PR.

thehenrytsai (approved):
🐐

alanhkarp:
I needed to test async events on another project.  I initially polled to make sure the event had finished, but the tests were taking a long time.  Instead, I created a promise in the test code that I resolved in the event handler, and the tests ran 2-3 times faster.  I don't know if this approach is appropriate for what you're doing, but I thought I'd mention it.

PR #760, opened 2024-06-15
Improved typing of PermissionGrant.parse() to avoid misuse

thehenrytsai:
1. Improved typing of PermissionGrant.parse() to avoid misuse
1. Detected and fixed a couple of unimportant misuses in the tests after the above change (case in point).


LiranCohen (approved):
🔥 🙇

csuwildcat (approved):

PR #764, opened 2024-06-20
Deprecate `EventsGet` in favor of `EventsQuery` without filters

LiranCohen:
`EventsGet` was used for sync before `EventsQuery` was made available, which allows filtering events/messages vs simply getting all of them.

This PR removes the `EventsGet` interface method and removes the restriction in `EventsQuery` to allow querying without any filters.


thehenrytsai (approved):
Love it when code gets removed!

PR #766, opened 2024-06-20
`EventsQuery` permissions grant

LiranCohen:


tests/utils/test-data-generator.ts
```diff
@@ -333,6 +358,41 @@ export class TestDataGenerator {
     };
   };
 
+  public static async generateGrantCreate(input?: GenerateGrantCreateInput): Promise<GenerateGrantCreateOutput> {
```
LiranCohen, line 361: Added this as a convenience, might add the rest in a separate PR and clean up usage of it manually. Makes for reading the tests much cleaner.

json-schemas/permissions/scopes.json
```diff
@@ -3,6 +3,17 @@
   "$id": "https://identity.foundation/dwn/json-schemas/permissions/scopes.json",
   "type": "object",
   "$defs": {
+    "events-query-scope": {
+      "type": "object",
+      "properties": {
```
thehenrytsai, line 8: Sanity: missing `protocol`?

src/core/events-grant-authorization.ts
```diff
@@ -0,0 +1,44 @@
...
+    await GrantAuthorization.performBaseValidation({
+      incomingMessage: eventsQueryMessage,
+      expectedGrantor,
+      expectedGrantee,
+      permissionGrant,
+      messageStore
+    });
+
+    // if the grant is scoped to a specific protocol, ensure that all of the query filters must include that protocol
+    if ('protocol' in permissionGrant.scope && permissionGrant.scope.protocol !== undefined) {
```
thehenrytsai, line 32: Optional: Is it possible/worth it to use `hasProtocolScope()`, just a thought, ignore if painful or not worth it.

tests/utils/test-data-generator.ts
thehenrytsai: > Makes for reading the tests much cleaner.

No push back for this convenience method 'coz it is probably a matter of preference and it does make the test look cleaner. Though I want to discuss a bit more with you as my usage on `generateXyz` message methods in `TestDataGenerator` has changed over time: I started out using it very heavily during initial phase of DWN development as it allows defaults in many places, but over time I use it less and it's been many months (possibly more than a year) since I last used it to generate DWN messages. The reason is probably because test scenarios are now way more nuanced/specific and defaults just don't cut it most of the time. And I'd prefer to go straight to the source create method than using a wrapper if it just saves me 1 or two defaults (in this case the expiry?), so I can be sure what gets passed in. I found it harder to debug using the generator method when tests fail because I have to reason through the default logic in the generator method, and some of those logic got complicated themselves over time... 🤣 

Again no push back, just wanted to have that discussion to give you context on why I started stay away from the generate methods.

LiranCohen: Yeah that makes sense! I think my favorite part about the `generateXyz` is the data stream, it gets ugly sometimes with long variable names 🤣 

Would you prefer I remove this and just create the grants using the normal records create where I'm using them?

json-schemas/permissions/scopes.json
LiranCohen: Good catch, fixed this and slightly improved the others that were lacking as well.

src/core/events-grant-authorization.ts
LiranCohen: Yup! Totally missed the opportunity for that, no circular deps so it's good!

tests/utils/test-data-generator.ts
thehenrytsai: Nah, we good as we on the same page regarding pros and cons.

thehenrytsai (approved):
🐐 🚀

thehenrytsai (approved):
🐐 🚀

PR #776, opened 2024-07-01
Refactor `MessagesRead` response

LiranCohen:
- Added `entry` to `UnionMessageReply` that represents response from `MessagesRead`.
- Modified `MessagesReadReplyEntry` to include `data` as a sibling property to `message`


thehenrytsai (approved):
🚀

PR #782, opened 2024-07-16
Replaced madge with dependency-cruiser to fix dependency vulnerabilities

thehenrytsai:


LiranCohen (approved):
🙇 🐐 

Awesome!!! I like the name `dependency cruiser`

PR #790, opened 2024-08-07
Return 204 when there is no data

### LINKED ISSUE #695: `RecordsWrite` with no data should return a 204 status code.
Currently when you write a `RecordsWrite` without any data, we assume that this is an `initial write` and the record will not be returned in a query.

When debugging issues, it could be difficult to discern between something that was accepted as a queryable write and something that was accepted as an initial state.

In discussing with @csuwildcat we decided a `204 - No Content` response to writing a record with no data would be most suitable for signaling this.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

andresuribe87:
Fixes #695 

Additionally, changes the `scripts.test` value so that source maps are correct and point to the actual implementation files instead of the transpiled JS files.

### DIFF
--- a/src/handlers/records-write.ts
+++ b/src/handlers/records-write.ts
@@ -163,7 +163,7 @@ export class RecordsWriteHandler implements MethodHandler {
     }
 
     const messageReply = {
-      status: { code: 202, detail: 'Accepted' }
+      status: { code: newMessageIsInitialWrite ? 204 : 202, detail: newMessageIsInitialWrite ? 'No Content' : 'Accepted' }
     };
 
     // delete all existing messages of the same record that are not newest, except for the initial write
--- a/src/handlers/records-write.ts
+++ b/src/handlers/records-write.ts
@@ -162,10 +162,10 @@ export class RecordsWriteHandler implements MethodHandler {
       throw error;
     }
 
-    const messageReply =  {
+    const messageReply = {
       status: (newMessageIsInitialWrite && dataStream !== undefined) ?
-          { code: 204, detail: 'No Content' } :
-          { code: 202, detail: 'Accepted' }
+        { code: 204, detail: 'No Content' } :
+        { code: 202, detail: 'Accepted' }
     };
 
     // delete all existing messages of the same record that are not newest, except for the initial write
--- a/src/handlers/records-write.ts
+++ b/src/handlers/records-write.ts
@@ -163,7 +163,7 @@ export class RecordsWriteHandler implements MethodHandler {
     }
 
     const messageReply = {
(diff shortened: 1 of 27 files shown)


tests/handlers/records-write.spec.ts
```diff
@@ -3020,7 +3020,7 @@ export function testRecordsWriteHandler(): void {
             recipient    : alice.did
           });
           const imageReply = await dwn.processMessage(alice.did, imageRecordsWrite.message, { dataStream: imageRecordsWrite.dataStream });
-          expect(imageReply.status.code).to.equal(202); // allows write but is not readable or queryable
+          expect(imageReply.status.code).to.equal(204); // allows write but is not readable or queryable
```
andresuribe87, line 3023: This seems suspect, but I saw that https://github.com/TBD54566975/dwn-sdk-js/blob/c6d8526e4aa83336b08f646d906c14c435b13311/tests/utils/test-data-generator.ts#L416 is not kicking in

andresuribe87:
@LiranCohen ready for review. Would appreciate some 👀 !

src/handlers/records-write.ts
```diff
@@ -163,7 +163,9 @@ export class RecordsWriteHandler implements MethodHandler {
     }
 
     const messageReply = {
-      status: { code: 202, detail: 'Accepted' }
+      status: (newMessageIsInitialWrite && dataStream === undefined) ?
```
LiranCohen, line 166: Nit: Would be good to add some context in a comment here. I know we mention a bit about writing without data above, but it might be good to add a small comment explaining why we return `No Content`.

LiranCohen (approved):
Awesome stuff!! Just a small nit if you can think of a good blurb to go above it explaining why we return a different code for that reason.

src/handlers/records-write.ts
andresuribe87: Added a comment.

andresuribe87:
@LiranCohen is it cool to merge bypassing branch protections?

tests/handlers/records-write.spec.ts
thehenrytsai: Fair observation, let me update the test!

thehenrytsai:
NOTE: The CVE was discovered in, and is being fixed by this PR waiting to be merged:

https://github.com/indutny/elliptic/pull/317

Therefore I am bypassing the branch protections for merging this.

tests/handlers/records-write.spec.ts
thehenrytsai: Updated.

PR #791, opened 2024-08-11
construction of grants and requests are not async

LiranCohen:
Currently `PermissionGrant` and `PermissionRequest` have an async `parse()` method to instantiate them with a private constructor.

However there is **currently** no async action needed to parse the grant/request. Any objection to removing the `parse()` method in favor of just using the constructor?


LiranCohen:
Keeping the pattern of `async parse()` over constructors.

PR #796, opened 2024-08-15
Disallowed write after delete

thehenrytsai:
Another PR will follow for allowing pruning to occur following a delete without pruning.


tests/handlers/records-write.spec.ts
```diff
@@ -4115,6 +4051,37 @@ export function testRecordsWriteHandler(): void {
         expect(recordsWriteReply.status.detail).to.contain(DwnErrorCode.RecordsWriteMissingEncodedDataInPrevious);
       });
 
+      it('should return 400 for if attempting a write after a delete', async () => {
```
LiranCohen, line 4054: nit:
```suggestion
      it('should return 400 if attempting a write after a delete', async () => {
```

LiranCohen (dismissed):
🐐

csuwildcat (approved):

LiranCohen (approved):

PR #804, opened 2024-09-11
remove ts-sinon dependency, upgrade sinon

LiranCohen:
`ts-sinon` is no longer maintained, and doesn't seem to be very useful for us.


csuwildcat (approved):

thehenrytsai (approved):
Really unfortunate that we now need to reference real object now for stubbing...

LiranCohen:
> Really unfortunate that we now need to reference real object now for stubbing...

Yeah :/ I wish the sinon squad would have taken over support for that wrapper. A few people have forked it, but none that seem worth taking in as a dep.

PR #812, opened 2024-10-02
A non-tenant `author` of a Record should be able to `RecordsRead` their record.

LiranCohen:
Currently a `RecordsQuery` will return any records which you are the `author` of or `recipient` of, regardless of the protocol rules.

However `RecordsRead` would only return a record which you are the `recipient` of, and fail protocol rules for messages that you are an `author` of.

This PR more closely aligns the two by allowing a non-tenant to Read any record which they have authored regardless of the protocol rules.

### DIFF
--- a/tests/handlers/records-read.spec.ts
+++ b/tests/handlers/records-read.spec.ts
@@ -243,7 +243,7 @@ export function testRecordsReadHandler(): void {
         const writeReply = await dwn.processMessage(alice.did, message, { dataStream });
         expect(writeReply.status.code).to.equal(202);
 
-        // Bob reads the data that Alice sent him
+        // Bob reads the record he sent to Alice from Alice's DWN
         const recordsRead = await RecordsRead.create({
           filter: {
             recordId: message.recordId,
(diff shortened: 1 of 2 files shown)


tests/handlers/records-read.spec.ts
```diff
@@ -205,6 +205,72 @@ export function testRecordsReadHandler(): void {
...
+        // Bob writes a foo record to Alice's DWN
+        const { message, dataStream, dataBytes } = await TestDataGenerator.generateRecordsWrite({
+          author       : bob,
+          protocol     : protocolDefinition.protocol,
+          protocolPath : 'foo',
+        });
+        const writeReply = await dwn.processMessage(alice.did, message, { dataStream });
+        expect(writeReply.status.code).to.equal(202);
+
+        // Bob reads the data that Alice sent him
```
thehenrytsai, line 246: Sanity:

```suggestion
        // Bob reads the record he sent to Alice from Alice's DWN
```

thehenrytsai (approved):

PR #813, opened 2024-10-02
Enforce requiring all read actions (`read`, `query`, `subscribe`)  to exist in a protocol role rule.

LiranCohen:
Currently there are some inconsistencies being able to fetch record data. If a user can only query for something, they are able to get the data if it is under the encodedData size limit, however if it is not they are unable to issue a `RecordsRead` for that record.

This change forces the user to include all 3 "read" actions (read, subscribe, query) when using any one of them in a protocol role-based action rule.


csuwildcat (approved):

thehenrytsai (approved):
🐐 🐐 🐐

PR #814, opened 2024-10-03
Updated RecordsRead to return RecordsDelete and initial RecordsWrite when record is deleted

thehenrytsai:
Prior to this change, a 404 status is returned without any data when a RecordsRead is performed against a deleted record.

But 404 status alone prevents a the RecordsDelete from being imported into the caller's DWN for the purpose updating the record's state.

This PR updated RecordsRead to return RecordsDelete and initial RecordsWrite when record is deleted.

### DIFF
--- a/src/handlers/records-read.ts
+++ b/src/handlers/records-read.ts
@@ -45,7 +45,6 @@ export class RecordsReadHandler implements MethodHandler {
     }
 
     // get the latest active messages matching the supplied filter
-    // only RecordsWrite messages will be returned due to 'isLatestBaseState' being set to true.
     const query: Filter = {
       // NOTE: we don't filter by `method` so that we get both RecordsWrite and RecordsDelete messages
       interface         : DwnInterfaceName.Records,
--- a/src/handlers/records-read.ts
+++ b/src/handlers/records-read.ts
@@ -3,7 +3,7 @@ import type { DidResolver } from '@web5/dids';
 import type { Filter } from '../types/query-types.js';
 import type { MessageStore } from '../types//message-store.js';
 import type { MethodHandler } from '../types/method-handler.js';
-import type { RecordsQueryReplyEntry, RecordsReadMessage, RecordsReadReply } from '../types/records-types.js';
+import type { RecordsDeleteMessage, RecordsQueryReplyEntry, RecordsReadMessage, RecordsReadReply } from '../types/records-types.js';
 
 import { authenticate } from '../core/auth.js';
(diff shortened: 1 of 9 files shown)


src/types/records-types.ts
```diff
@@ -203,10 +203,17 @@ export type RecordsReadMessage = {
...
+  /**
+   * The RecordsDelete if the record is deleted.
+   */
+  delete?: RecordsDeleteMessage;
+
+  /**
+   * The initial write of the record if the returned RecordsWrite message itself is not the initial write or if a RecordsDelete is returned.
+   */
+  initialWrite?: RecordsWriteMessage;
+
```
thehenrytsai, line 215: This serves as a starting point for discussion and needs to be finalized. While this approach is backwards compatible, it is but far from ideal. I personally would lean towards just revamping this type entirely and just require the downstream dependency (mainly `web5-js`) update its code. A few things to NOTE:

1. I considered overloading/polluting existing property `record` further, but `data` within it is currently required.
1. `initialWrite` is now duplicated, I am considering to take it out as a top level property (as shown in PR).
1. the naming `record` is less than ideal: it makes the peer property `delete` look out of place. Perhaps we should name them `recordsWrite` and `recordsDelete` or just `write` and `delete` respectively? Or we if combine the two message types (pt. 1), then maybe we call it `state` or `message`?!

csuwildcat (approved):

src/handlers/records-read.ts
```diff
@@ -47,6 +47,7 @@ export class RecordsReadHandler implements MethodHandler {
     // get the latest active messages matching the supplied filter
     // only RecordsWrite messages will be returned due to 'isLatestBaseState' being set to true.
```
LiranCohen, line 48:
```suggestion
```

src/types/records-types.ts
LiranCohen: Yeah I agree that the current form isn't ideal.

Curious to see what combining the two message types could look like, that might be more consistent with how it's normally consumed (just a record and a it's current state).

LiranCohen (approved):
🥇 🐐  This is definitely needed!

src/types/records-types.ts
thehenrytsai: Decided to submit a separate PR because it actually touches a bunch of files, so don't want this change to be drowned out by the refactor noise.

LiranCohen (approved):
🎸 🐐

PR #817, opened 2024-10-04
Refactored reply structure to a RecordsRead

thehenrytsai:
Currently the [reply structure of a `RecordsRead`](https://github.com/TBD54566975/dwn-sdk-js/blob/main/src/types/records-types.ts#L205) is quite messy: (duplicated properties, data stream nested in a message, inconsistent naming etc).

This is an attempt at cleaning it up with a flatter structure.

### DIFF
--- a/src/handlers/records-read.ts
+++ b/src/handlers/records-read.ts
@@ -72,8 +72,8 @@ export class RecordsReadHandler implements MethodHandler {
       const recordsDeleteMessage = matchedMessage as RecordsDeleteMessage;
       const initialWrite = await RecordsWrite.fetchInitialRecordsWriteMessage(this.messageStore, tenant, recordsDeleteMessage.descriptor.recordId);
       return {
-        status: { code: 404, detail: 'Not Found' },
-        recordsDelete: recordsDeleteMessage,
+        status        : { code: 404, detail: 'Not Found' },
+        recordsDelete : recordsDeleteMessage,
         initialWrite
       };
     }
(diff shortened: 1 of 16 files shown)


csuwildcat (dismissed):

src/handlers/records-read.ts
```diff
@@ -65,16 +65,20 @@ export class RecordsReadHandler implements MethodHandler {
 
     const matchedMessage = existingMessages[0];
 
+    // if the matched message is a RecordsDelete, we mark the record as not-found and return both the RecordsDelete the initial write
+    // NOTE: there is an opportunity in the future here to restrict this behavior further:
+    //       ie. perform similar authorization checks as when records exists before returning data about deleted record
```
LiranCohen, line 70:
```suggestion
    // if the matched message is a RecordsDelete, we mark the record as not-found and return both the RecordsDelete and the initial RecordsWrite
    // NOTE: there is an opportunity in the future here to restrict this behavior further:
    //       ie. perform similar authorization checks as when records exists before returning data about deleted record
```

Great call out!
Nit: Can we include an issue and TODO? This is definitely one i'd like to see prioritized at some point.

LiranCohen (approved):
🐐 🐐 🐐 🚀 🚀 🚀 

🙏  This is already much cleaner!

Just a small nit.

src/handlers/records-read.ts
thehenrytsai: Good idea, created issue and added a TODO referencing it.

LiranCohen (approved):

Issue #819, opened 2024-10-07
Consider performing auth logic before returning RecordsDelete and InitialWrite in the RecordsRead reply for a deleted record


thehenrytsai:
Currently we return RecordsDelete and InitialWrite in the RecordsRead reply for a deleted record regardless of who the caller is.

Consider performing auth logic before returning RecordsDelete and InitialWrite in the RecordsRead reply for a deleted record

PR #820, opened 2024-10-07
Auth when reading deleted records

andresuribe87:
This addresses #819.

### DIFF
--- a/src/handlers/records-read.ts
+++ b/src/handlers/records-read.ts
@@ -68,6 +68,22 @@ export class RecordsReadHandler implements MethodHandler {
     if (matchedMessage.descriptor.method === DwnMethodName.Delete) {
       const recordsDeleteMessage = matchedMessage as RecordsDeleteMessage;
       const initialWrite = await RecordsWrite.fetchInitialRecordsWriteMessage(this.messageStore, tenant, recordsDeleteMessage.descriptor.recordId);
+
+      if (initialWrite === undefined) {
+        return messageReplyFromError(new DwnError(
+          DwnErrorCode.RecordsReadInitialWriteNotFound,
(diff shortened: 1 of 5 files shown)


src/handlers/records-read.ts
```diff
@@ -68,6 +68,22 @@ export class RecordsReadHandler implements MethodHandler {
...
+        ), 400);
+      }
+
+      // Perform authorization before returning the delete and initial write messages
+      const parsedInitialWrite = await RecordsWrite.parse(initialWrite);
+      try {
+        await RecordsReadHandler.authorizeRecordsRead(tenant, recordsRead, parsedInitialWrite, this.messageStore);
+      } catch (error) {
+        return messageReplyFromError(error, 401);
+      }
```
andresuribe87, line 85: If this approach makes sense, I'll add tests next.

LiranCohen: @andresuribe87  🔥 Yup this is exactly what we were looking for! 🐐 🙏

andresuribe87: Added tests @LiranCohen 

Please take another look!

thehenrytsai: Hey @andresuribe87, there is a question on how we handle authorship, will sync up with @LiranCohen again and get back to you!

thehenrytsai: I think this is good to approve after talking with @LiranCohen, I'd like to rename some function arguments because the semantic of the write being passed for verification is no longer just the "matched RecordsWrite", but we can do that in a opportunistic clean up..

thehenrytsai (approved):
🏅 🙏  🚀

PR #821, opened 2024-10-07
Corrected UnionMessageReply type for downstream consumption

thehenrytsai:
Corrected UnionMessageReply type due to changes to #817 for downstream consumption such as the `dwn-server`.


LiranCohen (approved):
🙇 🙏 🐐

PR #827, opened 2024-10-23
Add DidWeb as a default resolover method

LiranCohen:
This PR adds the `DidWeb` method as one of the default resolvers when using a DWN.

After this PR the default resolvers will be: `DidJwk`, `DidKey`, `DidDht` and `DidWeb`


thehenrytsai (approved):
🙏 🐐 🚀

